PR #5, opened 2021-11-26
WIP: Signer

Arcticae:


src/net/signer.py
```diff
@@ -0,0 +1,86 @@
...
+        self.key_pair = key_pair
+
+    async def add_transaction(
+        self,
+        tx: InvokeFunction,
+        token: Optional[str] = None,
+    ) -> Dict[str, int]:
+        if tx.tx_type == TransactionType.DEPLOY:
+            return await super().add_transaction(tx, token)
+
```
Solpatium, line 52: We should throw a ValueError if a tx has a signature.

PR #9, opened 2021-11-30
Feat/cairo to python

Solpatium:


src/utils/data_transformer/data_transformer.py
```diff
@@ -0,0 +1,252 @@
...
+            result.extend(values)
+
+        return result
+
+    def to_python(self, cairo_type, name, values) -> (dict, CairoData):
+        definition = self._definition(cairo_type)
+
+        result = {}
+        for member_name, member in definition.members.items():
+            transformed, values = self.resolve_type(member.cairo_type).to_python(
```
Arcticae, line 105: kind of confusing since the parameter to the function is values as well, and the return data is values, can we make a distinction here?

src/utils/data_transformer/data_transformer.py
```diff
@@ -0,0 +1,252 @@
...
+            result[member_name] = transformed
+
+        return result, values
+
+
+class TupleTransformer(TypeTransformer[TypeTuple, tuple]):
+    def from_python(self, cairo_type, name, value):
+        values = [*value]
+        if len(values) != len(cairo_type.members):
+            raise ValueError(
```
Arcticae, line 117: i think we should raise TypeErrors for those type transformers

src/utils/data_transformer/data_transformer_test.py
```diff
@@ -0,0 +1,170 @@
...
+    "value, cairo_value",
+    [
+        ((1, 2, 3), [1, 2, 3]),
+        ((1,), [1]),
+        ((213, 2), [213, 2]),
+        ((1, 1, 1, 1, 1, 1), [1, 1, 1, 1, 1, 1]),
+    ],
+)
+def test_tuple(value, cairo_value):
+    print(value)
```
Arcticae, line 56: needs to be deleted

src/contract.py
```diff
@@ -105,16 +112,10 @@ def _make_invoke_function(self, *args, signature=None, **kwargs):
         return InvokeFunction(
             contract_address=self.contract_data.address,
             entry_point_selector=self.selector,
-            calldata=self._make_calldata(*args, **kwargs),
+            calldata=self._transformer(*args, **kwargs),
```
Arcticae, line 115: i kind of dislike the way of calling this transformer, can we just call `from_python` ?

Arcticae (changes requested):
Good job! I have a few questions and nitpicks but it looks great overall :)

src/utils/data_transformer/data_transformer.py
```diff
@@ -0,0 +1,252 @@
...
+        )
+
+    return values[:n], values[n:]
+
+
+UsedCairoType = TypeVar("UsedCairoType", bound=CairoType)
+PythonType = TypeVar("PythonType")
+
+
+@dataclass
```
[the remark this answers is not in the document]
Solpatium: Yup

src/utils/data_transformer/data_transformer.py
Solpatium: That's on purpose, we change values here

src/utils/data_transformer/data_transformer.py
Solpatium: [Typ](https://docs.python.org/3/library/exceptions.html#TypeError)
>Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

Types match here, so it is a case for value error.

src/contract.py
```diff
@@ -69,20 +69,27 @@ def __init__(
         self.inputs = abi["inputs"]
         self.contract_data = contract_data
         self._client = client
+        self._transformer = DataTransformer(
```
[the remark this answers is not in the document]
Solpatium: payload_transformer?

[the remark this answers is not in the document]
Arcticae: yeah works for me

Arcticae (approved):
LGTM

PR #11, opened 2021-12-01
Add starknet contract compilation and deployment

Arcticae:
TODO:
 

- [x] Make the test better, verifying the structure


src/utils/files.py
```diff
@@ -0,0 +1,5 @@
+import os
+
+
+def file_from_directory(directory: str, filename: str) -> str:
```
Solpatium, line 4: Do we need this function? It is just a wrapper for os.path.join.

src/net/client.py
```diff
@@ -147,3 +151,17 @@ async def add_transaction(
         self, tx: Transaction, token: Optional[str] = None
     ) -> Dict[str, int]:
         return await self._gateway.add_transaction(tx, token)
+
+    async def compile_and_deploy_contract(
```
Solpatium, line 155: I suggest we name it deploy_contract and accept both input_files and compiled_contract. In case compiled_contract is not provided we can compile input_files.

src/utils/compiler/starknet_compile.py
```diff
@@ -0,0 +1,57 @@
...
+    get_module_reader,
+)
+from starkware.cairo.lang.compiler.constants import MAIN_SCOPE, LIBS_DIR_ENVVAR
+from starkware.cairo.lang.compiler.preprocessor.preprocess_codes import preprocess_codes
+from starkware.starknet.compiler.compile import assemble_starknet_contract
+from starkware.starknet.compiler.starknet_pass_manager import starknet_pass_manager
+
+
+@dataclass
+class CairoSourceFile:
```
Solpatium, line 17: Users would need to import it to use it with our SDK, let's stick to regular data structures. I guess we can replace it with `TypedDict`.

Solpatium (changes requested):

src/net/client.py
Solpatium: Also we could accept contract as a string and compile it as contract.cairo. I imagine it would we useful for simple examples.

src/utils/compiler/starknet_compile_test.py
```diff
@@ -0,0 +1,14 @@
...
+
+directory = os.path.dirname(__file__)
+map_filename = os.path.join(directory, "map.cairo")
+test_file_content = open(map_filename).read()
+
+
+def test_starknet_compilation():
+    output_file_str = starknet_compile({map_filename: test_file_content})
+    output_json = json.loads(output_file_str)
+    assert output_json.get("abi") is not None
```
Solpatium, line 14: It will probably be an empty dict, so it won't be None

PR #12, opened 2021-12-02
AccountClient - account deployment

Arcticae:


src/net/account/account_client.py
```diff
@@ -1,13 +1,29 @@
...
+from src.utils.types import (
+    AddressRepresentation,
+    parse_address,
+    NetType,
+    net_address_from_type,
+)
+
+directory = os.path.dirname(__file__)
+compiled_account_contract_fname = os.path.join(directory, "Account-compiled.json")
+compiled_account_contract = open(compiled_account_contract_fname).read()
```
Solpatium, line 26: This is a module's side effect during module initialisation and it will block execution in asynchronous context. I suggest we place this json directly in a python file, we can add script for generating it in the future.

src/net/account/Account.cairo
```diff
@@ -0,0 +1,178 @@
+%lang starknet
```
Solpatium, line 1: I think we can replace this with a script that fetches contract from the source and generates a compiled json file.

src/net/account/account_client.py
```diff
@@ -89,3 +109,24 @@ async def add_transaction(
...
         )
+
+    @staticmethod
+    async def create_account(net: NetType, pk: Optional[int] = None) -> "Client":
+        if not pk:
+            pk = get_random_private_key()
+
+        key_pair = KeyPair.from_private_key(pk)
+
+        net_address = net_address_from_type(net)
```
Solpatium, line 120: What about providing a custom address instead of net selection?

src/net/account/account_client.py
```diff
@@ -89,3 +109,24 @@ async def add_transaction(
                 signature=[r, s],
             )
         )
+
+    @staticmethod
+    async def create_account(net: NetType, pk: Optional[int] = None) -> "Client":
```
Solpatium, line 114: Clients should also accept net parameter.

Solpatium, line 114: This creates `AccountClient`, not just `Client`. It is important because if user creates account without providing private key then at the end user might want to get generated private key from `AccountClient`.

Solpatium (changes requested):

PR #17, opened 2021-12-06
CPP signing library integration

Arcticae:


starknet/utils/crypto/facade.py
```diff
@@ -0,0 +1,71 @@
...
+    return sign(msg_hash, priv_key)
+
+
+def hash_message(
+    account: int,
+    to: int,
+    selector: int,
+    calldata: List[int],
+    nonce: int,
+):
```
Solpatium, line 60: Can we add return types here?

PR #18, opened 2021-12-07
Add basic sphinx setup

Solpatium:


Arcticae:
I think we need to include some readme section on docs, building them and maintaining (optionally). Would be nice to have a poe task to build and open them as well

Arcticae (approved):

PR #19, opened 2021-12-07
Add basic docs about the contract module

Solpatium:


docs/guide.rst
```diff
@@ -0,0 +1,191 @@
...
+
+Using existing contracts
+------------------------
+
+Although it is possible to use :ref:`Client` to interact with contracts, it requires translating python values into Cairo
+values. Contract offers that and some other utilities.
+
+Let's say we have a contract with this interface:
+
+.. code-block:: python
```
Arcticae, line 12: we can do just plain JSON here

docs/guide.rst
```diff
@@ -0,0 +1,191 @@
...
+    # Mixing positional with keyword arguments
+    invocation = await contract.functions.transferFrom.invoke(sender, recipient, amount=10000)
+
+    # Wait for tx
+    await invocation.wait_for_acceptance
+
+    (balance,) = await contract.functions.balanceOf.call(recipient)
+
+    # You can also call .to_dict on value returned from a call to get keyed values. It is useful with many returned
+    # values.
```
Arcticae, line 86: it can be moved to the prev line, no judging

Arcticae (approved):

docs/guide.rst
Solpatium: I wanted to use python value here, this way `abi` below refers to actual value

PR #34, opened 2021-12-13
CPP Signing improvements

Arcticae:


starknet/utils/crypto/facade.py
```diff
@@ -67,7 +67,7 @@ def use_cpp_variant() -> bool:
 def message_signature(msg_hash, priv_key) -> ECSignature:
     if use_cpp_variant():
         return cpp_sign(msg_hash, priv_key)
-    return sign(msg_hash, priv_key)
+    return sign(msg_hash, priv_key, 32)
```
Solpatium, line 70: I think we should accept it as a param with defalut = 32

Solpatium (approved):

PR #35, opened 2021-12-13
Add missing tests

Solpatium:


starknet/net/client.py
```diff
@@ -16,10 +16,6 @@
 
 @add_sync_version
 class Client:
-    @staticmethod
```
Solpatium, line 19: We use "test" in `__init__` for that now.

Arcticae (approved):
LGTM

PR #40, opened 2021-12-13
Add UInt256 support

Arcticae:


starknet/utils/data_transformer/data_transformer.py
```diff
@@ -99,6 +102,20 @@ def from_python(self, cairo_type, name, value):
...
+        (struct_name, *_) = definition.full_name.path
+
+        is_uint_256 = (
+            struct_name == "Uint256"
+            and len(definition.members.items()) == 2
+            and definition.members.get("low")
+            and definition.members.get("high")
+            and isinstance(definition.members["low"].cairo_type, TypeFelt)
+            and isinstance(definition.members["high"].cairo_type, TypeFelt)
+        )
```
Solpatium, line 114: I think we can move it into a separate function.

starknet/utils/data_transformer/data_transformer.py
```diff
@@ -99,6 +102,20 @@ def from_python(self, cairo_type, name, value):
...
+
+        if is_uint_256:
+            low, high = values
+            return UInt256((high << 128) + low), values
```
Solpatium, line 118: You need to return values without 2 elements that you've read. If you had more values `low, high = values` would raise.

Solpatium: Please add uint256 test to mixed params test, this way we'll be sure it works well with other types.

Solpatium (changes requested):

PR #42, opened 2021-12-14
Docs/add sync methods

Solpatium:


Arcticae (changes requested):
One case for async-> sync conversion is not covered by tests, can we add one case to cover it?

Arcticae (approved):

PR #63, opened 2021-12-21
Feat/bundle crypto cpp

Solpatium:


starknet_py/utils/crypto/cpp_bindings_test.py
```diff
@@ -1,53 +1,70 @@
...
-        return
-
     args = 1, 2, 3, [4, 5], 6
+    unload_cpp_lib()
 
-    monkeypatch.setenv("CRYPTO_C_EXPORTS_PATH", test_crypto_lib_path)
-    assert use_cpp_variant()
-    hash_1 = hash_message(*args)
+    times = []
+    for _ in range(2):
```
Solpatium, line 20: I guess we should remove time test here, right?

PR #64, opened 2022-01-03
Add munch-style shorthand for contract functions

0xdecaff:

Will update docs, write more tests, and generally upon this if there's any interest in this PR :)

Feature (minor quality of life / ux improvement)

Functions are accessed through dictionary-style key lookups:
```python
contract = await Contract.from_address(address, client)
await contract.functions["put"].invoke(key, value)
(response,) = await contract.functions["get"].call(key)
```

Functions can also be accessed through a shorthand (based on the [munch](https://github.com/Infinidat/munch) pattern):
```python
contract = await Contract.from_address(address, client)
await contract.fn.put.invoke(key, value)
(response,) = await contract.fn.get.call(key)
```

No breaking changes.

## **Other information**

Just getting into starknet development and this felt like low hanging fruit.

Opted to introduce a new property `.fn` instead of re-using /  upgrading the `.functions` property to avoid breaking changes. By trying to support `contract.functions.function_name.call()`, bugs would be introduced anywhere that a contract function shares the same name as a builtin `dict` method (e.g. "get", "keys", "items", etc...).

Open to suggestions / improvements! Or if you don't think such a feature would be right for the lib, all good :)

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -235,23 +235,18 @@ class ContractFunction:
         return get_selector_from_name(self.name)
 
 
-FunctionsRepository = Dict[str, ContractFunction]
-
-
-class FunctionsRepositoryShorthand(dict):
+class FunctionsRepository(dict):
     """
-    A simplified Munch implementation, supporting attribute-style access,
-    a la JavaScript.
-
-    https://github.com/Infinidat/munch
+    A dictionary of contract functions, allowing attribute-style access for
+    function names that do not overlap with dict builtins.
     """
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.__dict__ = self
-
-    def __getattribute__(self, __name: str) -> ContractFunction:
-        return cast(ContractFunction, super().__getattribute__(__name))
+    def __getattr__(self, name: str) -> ContractFunction:
+        if name not in self:
+            raise AttributeError(
+                f"'{type(self).__name__}' object has no attribute '{name}'"
+            )
+        return self[name]
 
 
 @add_sync_methods
@@ -277,13 +272,6 @@ class Contract:
         :return: All functions exposed from a contract.
         """
         return self._functions
-
-    @property
-    def fn(self) -> FunctionsRepositoryShorthand:
-        """
-        :return: All functions exposed from a contract, allowing attribue-style access.
-        """
-        return FunctionsRepositoryShorthand(**self._functions)
 
     @property
     def address(self) -> int:
@@ -441,7 +429,7 @@ class Contract:
     def _make_functions(
         cls, contract_data: ContractData, client: Client
     ) -> FunctionsRepository:
-        repository = {}
+        repository = FunctionsRepository()
 
         for abi_entry in contract_data.abi:
             if abi_entry["type"] != "function":
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -238,6 +238,22 @@ class ContractFunction:
 FunctionsRepository = Dict[str, ContractFunction]
 
 
+class FunctionsRepositoryShorthand(dict):
+    """
+    A simplified Munch implementation, supporting attribute-style access,
+    a la JavaScript.
+
+    https://github.com/Infinidat/munch
+    """
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


starknet_py/contract.py
```diff
@@ -228,6 +237,22 @@ def selector(self):
...
+
+    https://github.com/Infinidat/munch
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__dict__ = self
+
+    def __getattribute__(self, __name: str) -> ContractFunction:
+        return cast(ContractFunction, super().__getattribute__(__name))
```
0xdecaff, line 253: thoughts on guarding against the potential for shorthands to implicitly pass through to underlying `dict` methods?

example:
- contract does not have a "get" function
- `contract.fn.get` returns the `contract.functions.get`builtin dictionary function instead of raising a key error on `contract.functions["get"]`
- causing `contract.fn.get.call()` to raise this misleading error: `AttributeError: 'builtin_function_or_method' object has no attribute 'call'`

possible solutions:
1. keep a set of the contract function names and assert membership on access here
2. assert that the type returned from `super().__getattribute__(__name)` is instance of `ContractFunction`
3. other suggestions?

Solpatium: Thank you for spending your time and creating a PR, we really appreciate this! 👌 

We have implemented a similar solution, but decided that it might be problematic for reasons you mention. The goal for this library api is to be easy to use and hard to misuse. We were worried that someone might use `getattr` for instance when trying to get function with name in a variable and it could work for long time before someone adds a `get`/`update` etc function to a contract. 
We were not sure if saving some key presses is worth this risk, so we decided to use the simplest implementation that is just a simple dict with no additional magic.

Now that we have a bigger community we can revisit the topic. If we decide to implement something like that I would go with a solution that uses `__getattr__` instead of `__getattribute__`. This way user can be sure that dict methods will always be present. If contract has functions with names colliding with dict methods user can still use brackets to get wanted functions. This approach can also be added without any breaking changes.

There is also a solution that allows developers to use both dict methods and functions from contract, but it is way too tricky in my opinion 🧙.

0xdecaff: apologies on the delay! been logged out for a bit. appreciate your patience and being open to outside contribs :)

> We were not sure if saving some key presses is worth this risk, so we decided to use the simplest implementation that is just a simple dict with no additional magic.

Totally agree. Primary motivation here being that it's implemented in the [default contract testing library](https://github.com/starkware-libs/cairo-lang/blob/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/testing/contract.py#L76) so devs may be used to the pattern already. Certainly understand if this is undesirable for now! No offense taken at a PR close :)

> If we decide to implement something like that I would go with a solution that uses `__getattr__` instead of `__getattribute__`

Ahh interesting, wasn't fully aware of the nuance between the two. I see it now, changed :+1: 

> There is also a solution that allows developers to use both dict methods and functions from contract, but it is way too tricky in my opinion mage.

I've looked into some of those strategies, certainly cool, but agree that it's not worth the added complexity here

---

I've rebased & changed the implementation to use `__getattr__` while also consolidating the feature onto the `contract.functions` property

If there is support for this feature, is there anything else you forsee needing to be done? (aside from some unit coverage)

Should a similar strategy be used for `PreparedFunctionCall` arguments so there is feature parity between the two similar patterns?

Solpatium: >Totally agree. Primary motivation here being that it's implemented in the default contract testing library so devs may be used to the pattern already. Certainly understand if this is undesirable for now! No offense taken at a PR close :)

This is a very good point. What if we did it the same way here? So we would leave functions as they are now and add `__getattr__` to Contract for shorthand functions? 
I think the biggest problem would be that if we wanted to add some new functionalities to Contract that could cause a collision with existing usages. This problem is less likely to happen with a separate dict since dict's api probably won't change much between python versions 🤔 

@Arcticae also suggested to implement something like `contract.call.do_sth()` and `contract.invoke.do_sth()`.

>Should a similar strategy be used for PreparedFunctionCall arguments so there is feature parity between the two similar patterns?

I don't see a value in doing so, but I am open to suggestions. Calling/invoking functions is very common, getting arguments from prepared call is not.

PR #69, opened 2022-01-12
Refactor/utils cleanup

Solpatium:


starknet_py/net/client.py
```diff
@@ -1,24 +1,28 @@
...
     JsonObject,
 )
 from starkware.starknet.services.api.gateway.gateway_client import GatewayClient
-from services.external_api.base_client import RetryConfig, BadRequest as BadRequestError
+from starkware.starkware_utils.error_handling import StarkErrorCode
 
 from starknet_py.constants import TxStatus, ACCEPTED_STATUSES
 from starknet_py.utils.sync import add_sync_methods
-from starknet_py.utils.types import net_address_from_net, InvokeFunction, Transaction
+from .models import InvokeFunction, Transaction, Deploy
```
Arcticae, line 17: i'd rather we used absolute imports everywhere

starknet_py/net/client.py
```diff
@@ -213,3 +217,24 @@ async def add_transaction(
...
+    async def deploy(
+        self,
+        compiled_contract: Optional[str],
+        constructor_calldata: List[int],
+        salt: Optional[int] = None,
+    ) -> dict:
+        res = await self.add_transaction(
+            tx=Deploy(
+                contract_address_salt=ContractAddressSalt.get_random_value()
+                if salt is None
```
Arcticae, line 230: maybe `salt or ContractAddressSalt.get_random_value()` would be more readable

starknet_py/net/client.py
```diff
@@ -213,3 +217,24 @@ async def add_transaction(
...
+                else salt,
+                contract_definition=ContractDefinition.loads(compiled_contract),
+                constructor_calldata=constructor_calldata,
+            )
+        )
+
+        if res["code"] != StarkErrorCode.TRANSACTION_RECEIVED.name:
+            raise Exception("Transaction not received")
```
Arcticae, line 238: Can we cover it with tests?

starknet_py/net/networks.py
```diff
@@ -0,0 +1,20 @@
+from typing import Union
+
+MAINNET = "mainnet"
+TESTNET = "testnet"
+
+try:
+    from typing import Literal  # pylint: disable=no-name-in-module
+
+    PredefinedNetwork = Literal["mainnet", "testnet"]
+except:  # pylint: disable=bare-except
```
Arcticae, line 10: shouldn't we catch ImportError here only?

starknet_py/net/networks.py
```diff
@@ -0,0 +1,20 @@
...
+    PredefinedNetwork = str
+
+Network = Union[PredefinedNetwork, str]
+
+
+def net_address_from_net(net: Network) -> str:
```
Arcticae, line 16: I think it should actually return Optional[str] since network can be a string

Arcticae (changes requested):

starknet_py/net/client.py
Solpatium: salt can be `0`

PR #75, opened 2022-01-20
L1 <> L2 Messaging

Arcticae:
- [x] We should probably consider having block number parameter passed over to the user if he wishes to check latest/pending blocks 
- [x] Document all of the new features
- [x] Write unit tests


starknet_py/net/l1/contracts.py
```diff
@@ -0,0 +1,24 @@
...
+from web3.contract import Contract as Web3Contract
+from web3.providers import BaseProvider as L1Provider
+
+from starknet_py.net.l1.networks import L1Network
+from starknet_py.net.l1.starknet_core_abi import STARKNET_CORE_L1_ABI
+
+
+class StarknetL1Contract:
+    @staticmethod
+    def on_l1_net(l1_net: L1Network, l1_net_provider: L1Provider) -> Web3Contract:
```
Solpatium, line 11: We already have a chain parameter (for instance when creating a Contract), can we reuse it here?

Solpatium, line 11: Names are a little bit long, what do you think about `net` and `provider`?

starknet_py/net/l1/messages.py
```diff
@@ -0,0 +1,95 @@
...
+
+from web3.providers import BaseProvider as L1Provider
+from starknet_py.net.l1.contracts import StarknetL1Contract
+from starknet_py.net.l1.networks import L1Network
+
+
+def encode_packed(*args: List[int]) -> bytes:
+    return reduce(
+        lambda acc, x: acc + x,
+        map(lambda x: x.to_bytes(32, byteorder="big", signed=False), args),
```
Solpatium, line 14: `[x.to_bytes(32, byteorder="big", signed=False)  for x in args]` is shorter and simpler.

starknet_py/net/l1/messages.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @classmethod
+    def from_hash(cls, msg_hash: int) -> "L1Message":
+        return cls(hash=msg_hash)
+
+    @classmethod
+    def from_content(cls, msg_content: L1MessageContent) -> "L1Message":
+        return cls.from_hash(msg_content.hash)
+
+    def get_status(self, l1_net: L1Network, l1_net_provider: L1Provider):
```
Solpatium, line 48: What can be returned here?

starknet_py/net/l1/messages.py
```diff
@@ -0,0 +1,95 @@
...
+        return (
+            StarknetL1Contract.on_l1_net(l1_net, l1_net_provider)
+            .functions.l2ToL1Messages(self.hash)
+            .call()
```
Solpatium, line 52: Is this synchronous?

docs/guide.rst
```diff
@@ -300,4 +300,113 @@ You can read more about how cairo treats shortstrings in `the documentation <htt
...
+                l1_sender=123, # Integer representation of L1 hex address
+                l2_recipient="0x123123123", # Either a hex L2 address, or it's integer representation
+                nonce=1, # Can be retrieved from L1 transaction's receipt (the one containing the sent message)
+                function_name="dummy_name", # L2 recipient function name
+                payload=[32, 32, 32, 32], # L2 Function calldata, list of ints
+            )
+        )
+        status = await l1_to_l2_msg.get_status(  # For sync version, use 'get_status_sync'
+            chain_id=StarknetChainId.TESTNET,
+            endpoint_uri="https://my-rpc-endpoint.com/", # Only HTTP RPC endpoints are supported for now
```
Solpatium, line 341: Can we provide a default?

docs/guide.rst
```diff
@@ -300,4 +300,113 @@ You can read more about how cairo treats shortstrings in `the documentation <htt
...
+    l1_to_l2_msg = L2Message.from_content(
+            L2MessageContent(
+                l1_sender=123, # Integer representation of L1 hex address
+                l2_recipient="0x123123123", # Either a hex L2 address, or it's integer representation
+                nonce=1, # Can be retrieved from L1 transaction's receipt (the one containing the sent message)
+                function_name="dummy_name", # L2 recipient function name
+                payload=[32, 32, 32, 32], # L2 Function calldata, list of ints
+            )
+        )
+        status = await l1_to_l2_msg.get_status(  # For sync version, use 'get_status_sync'
```
Solpatium, line 339: I guess we could rename it to `count_queued` in both l1->l2 and l2->l1, right?

starknet_py/net/l1/contracts.py
```diff
@@ -0,0 +1,84 @@
...
+class StarknetL1Contract:
+    def __init__(
+        self,
+        net: StarknetChainId,
+        endpoint_uri: str,
+    ):
+        self.w3 = get_w3_provider(endpoint_uri)
+        self.contract_address = get_l1_starknet_contract_address(net)
+
+    async def l2ToL1Messages(
```
Solpatium, line 36: Why not snake_case since we wrap it anyway?

Solpatium (changes requested):

docs/guide.rst
Arcticae: I don't think this would be wise, since we can't guarantee really that the default endpoint works

starknet_py/net/l1/messages.py
```diff
@@ -0,0 +1,169 @@
...
+    def from_hash(cls, msg_hash: bytes) -> "L1Message":
+        """
+
+        :param msg_hash: Message's hash
+        :return: Instance of `L1Message`
+        """
+        return cls(hash=msg_hash)
+
+    @classmethod
+    def from_content(cls, msg_content: L1MessageContent) -> "L1Message":
```
bbrandtom, line 69: I think that for this to be useful, there should also be a way to extract `L1MessageContent` from the transaction receipt.
Otherwise it is very hard for me as a user to know the exact L1MessageContent that I've sent.

bbrandtom (changes requested):

docs/guide.rst
```diff
@@ -300,4 +300,117 @@ You can read more about how cairo treats shortstrings in `the documentation <htt
...
+
+        # 2. From message hash
+        l1_to_l2_msg = L1ToL2Message.from_hash(
+            (123).to_bytes(32, "big") # Provide 32 bytes as an input here, instead of message's content
+        )
+
+        # 3. From transaction receipt
+        w3 = web3.Web3(web3.providers.HTTPProvider("https://my-rpc-endpoint.com/"))
+        tx_receipt = w3.eth.wait_for_transaction_receipt("0x123123123")
+        l2_to_l1_msg = L2ToL1Message.from_tx_receipt(tx_receipt)
```
bbrandtom, line 355: Please change l2_to_l1_msg to l1_to_l2_msg

same for L2ToL1Message

Same for exmaple 4.

docs/guide.rst
```diff
@@ -300,4 +300,117 @@ You can read more about how cairo treats shortstrings in `the documentation <htt
...
+                payload=[32, 32, 32, 32], # L2 Function calldata, list of ints
+            )
+        )
+
+        # 2. From message hash
+        l1_to_l2_msg = L1ToL2Message.from_hash(
+            (123).to_bytes(32, "big") # Provide 32 bytes as an input here, instead of message's content
+        )
+
+        # 3. From transaction receipt
```
bbrandtom, line 352: Might be worth explaining what transaction should be used here.

bbrandtom (changes requested):
One error in the guide

bbrandtom (approved):

starknet_py/contract.py
```diff
@@ -238,9 +238,13 @@ async def invoke(self, *args, **kwargs) -> InvocationResult:
         """
         return await self.prepare(*args, **kwargs).invoke()
 
-    @property
-    def selector(self):
-        return get_selector_from_name(self.name)
+    @staticmethod
```
Solpatium, line 241: This is just a wrapper for `get_selector_from_name`. What do we gain by this change?

starknet_py/net/l1/l1_messaging_test.py
```diff
@@ -0,0 +1,175 @@
...
+                "type": "0x2",
+            }
+        )
+
+    w3_mock_receipt.eth.getTransactionReceipt = get_tx_receipt
+
+    mock_l2_client = Mock()
+    # L2 Mock
+
+    @asyncio.coroutine
```
Solpatium, line 138: Can't we write `async def` instead?

starknet_py/net/models/address.py
```diff
@@ -53,3 +54,4 @@ def compute_address(
 
 
 BlockIdentifier = as_our_module(BlockIdentifier)
+EthBlockIdentifier = as_our_module(EthBlockIdentifier)
```
Solpatium, line 57: I think using web3 types is ok as it is a widely known library and it only used when communicating with L1.

starknet_py/contract.py
Arcticae: We're requiring the user to pass the selector to the L1 message dataclass, and since it's hidden inside our dependency, user would have to import that from there. We can expose it this way to make it easier

PR #76, opened 2022-01-21
Pending block support

Arcticae:
- [x] Check documentation


starknet_py/contract.py
```diff
@@ -10,6 +10,7 @@
 from starkware.starknet.services.api.contract_definition import ContractDefinition
 from starkware.starknet.services.api.feeder_gateway.feeder_gateway_client import (
     CastableToHash,
+    BlockIdentifier,
```
Solpatium, line 13: This will be visible in documentation. Can we use`as_our_module` or do something else?

Solpatium (approved):

PR #89, opened 2022-02-22
Sync deploy fix

### LINKED ISSUE #82: [BUG] `deploy_sync` seems to be blocking
Hey thanks for developing `starknet.py`, awesome lib we are starting to use at https://yagi.fi! Friendly (possible) bug report.

I'm trying to deploy an account contract using `deploy_sync` but it hangs rather than returning the pending tx (which is what `invoke_sync` does).

## Python version
3.8

## Expected Behavior
Should return pending tx.

## Current Behavior
Hangs.

## Possible Solution
Match `invoke_sync` behavior with `deploy_sync`.

## Steps to Reproduce
```python
source_code = Path("contracts/Account.cairo").read_text("utf-8")
result = Contract.deploy_sync(
    client=Client("testnet"),
    compilation_source=source_code,
    constructor_args={
        "_public_key": public_key
    }
)
```

## Context (Environment)
Not able to deploy an account contract using `starknet.py`.

## Detailed Description
-

## Possible Implementation
-

MaksymilianDemitraszek:

## **What kind of change does this PR introduce?** 
Fixes #82, improves interface for contracts deployment

#82 

Yes, users need to transform contract deployment flow to
```
deployment_result = await Contract.deploy()
await deployment_result.wait_for_acceptance()
contract = deployment_result.contract
```

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -21,7 +21,7 @@ Setup
     poetry install
 
     # Make sure everything was installed properly
-     poe test
+    poe test
 
 Git hooks
 ---------
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -21,7 +21,7 @@ Setup
     poetry install
 
     # Make sure everything was installed properly
-    poe test
+     poe test
 
 Git hooks
 ---------
@@ -39,6 +39,9 @@ Documentation
 `Sphinx <https://www.sphinx-doc.org/en/master/>`_ is used for generating documentation.
 
 .. code-block:: bash
+
+    # Install additional dependencies for docs
+    poetry install -E docs
 
     # Generate HTML documentation
     poe docs_create
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -1,7 +1,7 @@
 import dataclasses
 from dataclasses import dataclass
 from functools import lru_cache
-from typing import List, Optional, Tuple, Union, Dict, Collection, NamedTuple
+from typing import List, Optional, Union, Dict, Collection, NamedTuple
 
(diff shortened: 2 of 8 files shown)


starknet_py/contract.py
```diff
@@ -82,6 +81,24 @@ async def wait_for_acceptance(
         )
 
 
+@add_sync_methods
+@dataclass(frozen=True)
+class InvokeResult(SentTransaction):
+    contract: ContractData = None
+
+    def __post_init__(self):
+        assert self.contract is not None
```
Solpatium, line 90: Is this required if you delete
```
 = None
```
above? 🤔

docs/development.rst
```diff
@@ -21,7 +21,7 @@ Setup
     poetry install
 
     # Make sure everything was installed properly
-    poe test
+     poe test
```
Solpatium, line 24: 👁️

starknet_py/contract.py
```diff
@@ -51,24 +51,23 @@ def from_abi(address: int, abi: ABI) -> "ContractData":
...
+    Dataclass exposing the interface of transaction related to a performed action
     """
 
     hash: CastableToHash
-    contract: ContractData
     _client: "Client"
     status: Optional[str] = None
     block_number: Optional[int] = None
 
     async def wait_for_acceptance(
```
Solpatium, line 64: It has to return same type as self when subclassed. We probably need a `TypeVar` here.
```
...self: T, ...): T
```

Solpatium (changes requested):
Good job! 🔥

starknet_py/contract.py
Solpatium: Unless it works like this already of course :)

Issue #91, opened 2022-03-01
Upgrade web3.py dependency


unparalleled-js:
**Is your feature request related to a problem? Please describe.**
I am trying to install this package in an environment using `web3.py==5.27.0` but this package is pinned to `web3.py==5.3.0`.

**Describe the solution you'd like**
Upgrade the `web3.py` dependency to include `5.27.0` in its supported range as well addressing any migrations related to this upgrade.

**Describe alternatives you've considered**
Not using this library directly or forking it? I don't know, the starknet APIs currently seem tricky. This is an excellent resources nonetheless!

PR #92, opened 2022-03-01
chore: upgrade web.py and adjust supported python 3.7 range

unparalleled-js:

update web3.py

fixes: #91 

Using latest web3.py

The user may need to upgrade any logic related to older web3.py (though I am not sure what that would be, likely, it will be dealing with dependency related issues).

## **Other information**
I am not sure how to test locally because I think the `development` branch is assuming unreleased changes to `cairo-lang`; correct me if I am wrong, please. I have not figured out how to get a working development environment for `cairo-lang` (and I really don't need to, for the most part).


unparalleled-js:
@Arcticae Do you have any thoughts on this one?
I am nervous with web3.py in general, as it is a shared dependency often.

Python projects using starknet and etheruem together are likely going to need web3.py and getting these dependencies to agree on a version might be challenging.

unparalleled-js:
I did some digging and found out `web.py` needed to adjust their python dependency support: https://github.com/ethereum/web3.py/issues/2378

So, technically the latest `web.py` requires python `>=3.7.2`.

I think `Starknet-Py` should also, as most people use the latest minor in their major series, like python `3.7.12` if they are using python `3.7`, and there are real flaws with these version as well, which is why `ipfshttpclient` does not support them: https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/master/pyproject.toml#L15-L23

Thus, I was able to create lock file by bumping said python version!

unparalleled-js:
@Arcticae I believe I have found the magical range.

Testing I have done:

I ran the following commands in this order and there were no errors:

```bash
poetry lock
poetry install --extras docs
poetry install --extras eth-tester
poetry install --no-dev
poetry install
```

Arcticae:
@unparalleled-js Thank you for your contribution!

PR #101, opened 2022-03-07
fix: correct the type of Multicall.calls

unparalleled-js:

Fix / Docs

The type of `Multicall.calls` was documented as a a `Call` but used everywhere as a list of `Call` objects.

The type of `Multicall.calls` was documented as a a `Call`.

It changed the type to a `List[Call]`.

No breaking changes.

### DIFF
--- a/starknet_py/utils/crypto/facade.py
+++ b/starknet_py/utils/crypto/facade.py
@@ -44,7 +44,7 @@ class Call:
 @dataclass(frozen=True)
 class MultiCall:
     account: int
-    calls: List[Call]
+    calls: Iterable[Call]
     nonce: int
     max_fee: int = 0
     version: int = 0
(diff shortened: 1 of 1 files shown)


starknet_py/utils/crypto/facade.py
```diff
@@ -44,7 +44,7 @@ class Call:
 @dataclass(frozen=True)
 class MultiCall:
     account: int
-    calls: Call
+    calls: List[Call]
```
MaksymilianDemitraszek, line 47:
```suggestion
    calls: Iterable[Call]
```
We probably can generalise it to

MaksymilianDemitraszek (changes requested):
Good catch, we definietely need to plug some static type checking :D

PR #102, opened 2022-03-09
Add Compiler module

cptartur:

Create Compiler module for compiling contracts into `ContractDefinition`s, to decouple compile functionality from the rest of the Contract module.

Removes `Contract._make_definition` and replaces it with `Compiler.create_contract_definition`


starknet_py/compile/compiler.py
```diff
@@ -0,0 +1,60 @@
...
+    "CairoSource", Union[CairoSourceCode, List[CairoFilename]]
+)
+
+
+class Compiler:
+    """
+    Class for compiling Cairo contracts
+    """
+
+    def __init__(self):
```
Arcticae, line 19: This seems obsolete

starknet_py/compile/compiler.py
```diff
@@ -0,0 +1,60 @@
...
+            )
+
+        return ContractDefinition.loads(compiled_contract)
+
+    @staticmethod
+    def compile_contract(
+        compilation_source: StarknetCompilationSource,
+        search_paths: Optional[List[str]] = None,
+    ) -> str:
+        """
```
Arcticae, line 53: I think it's worth adding to the docs if you provided the docstrings :)

starknet_py/compile/compiler_test.py
```diff
@@ -0,0 +1,25 @@
+import json
+from starkware.starknet.services.api.contract_definition import ContractDefinition
+
+from starknet_py.compile.compiler import Compiler
+from starknet_py.compile.map import MAP_CONTRACT
```
Arcticae, line 5: I think it'd be best to avoid adding compiled cairo code to the repo, and we can re-use map contract from `starknet_py/utils/compiler` module

Arcticae (changes requested):
Nice job! I'd do 2 more things to make this more complete:
- Replace usages of `starknet_py/utils/compiler` with the new compiler 
- Consolidate, or merge `starknet_py/utils/compiler/starknet_compile.py` implementation with this new module (tests as well). Having 2 separate ones seems confusing

cptartur:
> Nice job! I'd do 2 more things to make this more complete:
> 
> * Replace usages of `starknet_py/utils/compiler` with the new compiler
> * Consolidate, or merge `starknet_py/utils/compiler/starknet_compile.py` implementation with this new module (tests as well). Having 2 separate ones seems confusing

Right, consolidating these two modules seem logical. I'll work on that

starknet_py/compile/compiler.py
```diff
@@ -0,0 +1,127 @@
...
+        :param search_paths: a ``list`` of paths used by starknet_compile to resolve dependencies within contracts
+        :return: a ContractDefinition
+        """
+        if not contract_source and not compiled_contract:
+            raise ValueError(
+                "One of compiled_contract or compilation_source is required."
+            )
+
+        if not compiled_contract:
+            compiled_contract = Compiler.compile_contract(
```
unparalleled-js, line 49: I am curious why we are using static methods for everything? Couldn't these just be normal methods on the class so you can do `self.compile_contract` here? Else, couldn't all these methods just occur at the module level?

starknet_py/compile/compiler.py
```diff
@@ -0,0 +1,127 @@
...
+        :return: string of compiled contract
+        """
+        return Compiler._compile(compilation_source, search_paths=search_paths)
+
+    @staticmethod
+    def _load_cairo_source_code(filename: CairoFilename) -> str:
+        source_file = Path(filename)
+
+        if not source_file.is_file():
+            raise TypeError(f"{filename} does not exist")
```
unparalleled-js, line 74: Both `TypeError`s that get raised in this method seem more like `ValueError` to me.
`TypeError` is only meant to be used when something literally has the wrong type, like it expects a list instead of a str.

unparalleled-js (commented):
I live moving this out of utils, that feels good.
My main comment is that I think we should use `ValueError` instead of `TypeError`.

starknet_py/compile/compiler.py
Arcticae: Yep we should probably instantiate compiler with `compilation_source, search_paths` and then use methods accordingly

Arcticae (commented):
Looks good! After addressing remaining discussions, can be merged :)

starknet_py/compile/compiler.py
cptartur: Right, this will need to be changed

starknet_py/compile/compiler.py
cptartur: We can probably change these methods to non static, but I think that `compilation_source` and `search_paths` should still be methods' parameters. Otherwise you'd have to initialize a new compiler instance every time you wanted to compile some other contract

cptartur:
We'll go back to this PR after support for starknet 0.8.0 is done

Arcticae (changes requested):

starknet_py/compile/compiler.py
```diff
@@ -0,0 +1,133 @@
...
+        if not contract_source and not compiled_contract:
+            raise ValueError(
+                "One of compiled_contract or compilation_source is required."
+            )
+
+        self.compiled_contract = compiled_contract
+        self.contract_source = contract_source
+        self.search_paths = search_paths
+
+    def create_contract_definition(
```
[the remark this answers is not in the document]
cptartur: This method is currently only used in `contract.py` so maybe it could be a static method (or event just a method) on Contract?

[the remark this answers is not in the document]
Arcticae: Whichever you are more comfortable with. I think it's weird to pass a compiled contract to the compiler, and since it creates contract definition it might live on contract, but wouldn't it bloat the code too much?

starknet_py/compile/compiler.py
```diff
@@ -0,0 +1,133 @@
...
+
+        return ContractDefinition.loads(self.compiled_contract)
+
+    def compile_contract(self) -> str:
+        """
+        Compiles a contract and returns it as string
+
+        :return: string of compiled contract
+        """
+        return Compiler._compile(self.contract_source, search_paths=self.search_paths)
```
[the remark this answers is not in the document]
Arcticae: and we wouldn't have to pass params then. if we want to compile something as a one-off we'd do `Compiler(param1, param2).compile()`

Arcticae (changes requested):

Arcticae (approved):

PR #107, opened 2022-03-11
feat: expose public key in AccountClient

unparalleled-js:

Feature: Expose the public key in the `AccountClient`.

You have to access a protected property to find the public key of an `AccountClient`.

You can do `account_client.public_key` (similar to `account_client.private_key`)

No

## **Other information**
I don't think this all that useful for most use-cases, but from an "I need to implement this abstract base class perspective", this helps me out.

Am I missing something though? I may not be fully understanding the reason why it is not currently exposed.

### DIFF
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -56,6 +56,10 @@ class AccountClient(Client):
     def private_key(self) -> int:
         return self._key_pair.private_key
 
+    @property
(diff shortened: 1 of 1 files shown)


unparalleled-js:
Something may be weird here since you can technically change the public key in an account later on, in which case both of these props would be out-dated... hmm..

Arcticae (approved):

Issue #113, opened 2022-04-05
Show rejection reason in exception


unparalleled-js:
**Is your feature request related to a problem? Please describe.**

I am unable to make useful assertions on rejection reasons because the exception that gets raised in a generic exception that does not reveal any of the information.

**Describe the solution you'd like**

In `client.wait_for_tx()`, raise a custom error `TransactionRejected(result["message"])` that includes that failure reason from the `result`.

**Describe alternatives you've considered**

Implementing a custom `wait_for_tx()`.

**Additional context**

I was failing to sign a transaction correctly but the error message was not telling me that. I had to set a breakpoint in the lib to find out the details of the failure.

unparalleled-js:
I'm actually seeing some strange properties on the `TransactionInfo` object...

cptartur:
Hey, thanks for the suggestion about custom exceptions.

`TransactionInfo` object are actually created internally in cairo-lang library, they're not part of starknet.py. If you think it's a problem with starknet.py I'd need more details on what's failing.

unparalleled-js:
> `TransactionInfo` object are actually created internally in cairo-lang library, they're not part of starknet.py. If you think it's a problem with starknet.py I'd need more details on what's failing.

Ah, I don't actually think it is a problem with starknet.py but the error is so strange, I would be remiss if I did not at least hint at it.
Basically, I can see some random OS paths in my error message that are definitely not me. I am running `starknet-devnet` as my provider.

cptartur:
Of course. We're actually using `starknet-devnet` for our end-to-end tests and for this limited use it seems to be working most of the time. But I don't dismiss you findings, there may be some hidden problems - I'd recommend investigating more and opening an issue with devnet developers. 

If you manage to find it is actually related to starknet.py, please let us now!

unparalleled-js:
Yay, the PR looks great. I had one comment through (I know it is already merged, but just in case! https://github.com/software-mansion/starknet.py/pull/115/files#r846394692)

PR #115, opened 2022-04-06
Add custom exceptions for transaction failures

cptartur:

Add custom exceptions for transaction failures to be used in `Client.wait_for_tx()`. 

See #113 

Generic `Exception` with custom message is being raised.

Custom exceptions are used.

Custom exceptions still extend base `Exception` types so no changes in error handling on user side should be necessary.

### DIFF
--- a/starknet_py/net/client.py
+++ b/starknet_py/net/client.py
@@ -211,9 +211,7 @@ class Client:
                 if not wait_for_accept and "block_number" in result:
                     return result.block_number, status
             elif status == TxStatus.REJECTED:
-                raise TransactionRejectedError(
-                    str(result.transaction_failure_reason)
-                )
+                raise TransactionRejectedError(str(result.transaction_failure_reason))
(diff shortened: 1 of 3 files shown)


cptartur:
I guess some test cases should be written for these new exceptions? We were not necessarily testing them before though.

cptartur:
I've added some tests for custom exceptions. Codecov wants me to test `__str__` methods for these exceptions, but I don't think it makes much sense.

Arcticae (approved):

starknet_py/net/client.py
```diff
@@ -26,6 +27,11 @@
     chain_from_network,
 )
 from starknet_py.net.networks import Network, net_address_from_net
+from starknet_py.transaction_exceptions import (
+    TransactionFailedException,
```
unparalleled-js, line 31: It is more conventional to end custom error classes with `Error` instead of `Exception`

https://namingconvention.org/python/exception-naming.html

PR #118, opened 2022-04-06
Support cairo-lang 0.8.1

cptartur:

Support cairo-lang 0.8.1

No

## **Other information**

Making this a draft since `starknet-devnet` does not support cairo-lang 0.8.1 as of creating this PR, so we cannot realistically test for errors other than manual testing (limited).

Automatic tests will most likely fail

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -30,7 +30,7 @@ jobs:
         poetry run poe lint
     - name: Unit & e2e test
       run: |
-        poetry run poe test
+        poetry run poe test_unit
     - name: Generate coverage in XML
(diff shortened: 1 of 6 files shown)


cptartur:
This PR would use `starknet-devnet` directly from [github](https://github.com/Shard-Labs/starknet-devnet/tree/cairo-lang-0.8.1) since it's not available on PyPi as of yet. This is not ideal, but it's our best shot at keeping e2e working for now. 

I'd recommend we restore support for python 3.9 when new version of devnet is released and everything can be tested properly.

Arcticae (approved):

PR #135, opened 2022-04-29
Contract.from_address proxy support

### LINKED ISSUE #128: [BUG] Proxy contracts are not able to find ABIs
When I interact with a smart-contract via calling a mutable method on the live Testnet (goerli alpha), I get an Unknown location error.

## Python version
Python 3.7.9

## Expected Behavior
I can call mutable methods on contracts from goerli

## Current Behavior
I get the following error:

```
ERROR: (ContractLogicError) Error at pc=0:32:Got an exception while executing a hint.Cairo traceback (most recent call last):Unknown location (pc=0:497)Unknown location (pc=0:456)Unknown location (pc=0:513)Error in the called contract (0x7394cbe418daa16e42b87ba67372d4ab4a5df0b05c6e554d158458ce245bc10):Error at pc=0:6:Got an exception while executing a hint.Cairo traceback (most recent call last):Unknown location (pc=0:795)Unknown location (pc=0:772)Unknown location (pc=0:1232)Traceback (most recent call last):  File "<hint0>", line 3, in <module>AssertionError: assert_not_zero failed: 0 = 0.').
```
I am using the test token contract deployed to 0x07394cbe418daa16e42b87ba67372d4ab4a5df0b05c6e554d158458ce245bc10

When I deploy an ERC20 to my local `starknet-devnet`, I can call mutable methods and things work just fine.

## Possible Solution

We need to grab the ABI from the implementation contract.

## Steps to Reproduce

```python
from starknet_py.net import AccountClient
from starknet_py.net import KeyPair

# This account comes from ArgentX and is a proxy contract.
# The error I get is that it cannot find the __execute__ ABI.
# The __execute__ method needs to be grabbed from the implementation contract.
account_address = 2976162931320574283751166485501184860501729793478009661243729394508001134570

# Create the account client (note: my pkey is omitted for obvious reasons)
key_pair = KeyPair(
    private_key=******,
    public_key=888158711641686699726990655481927574337047407568
)
account_client = AccountClient(
    account_address,
    key_pair,
    'https://alpha4.starknet.io',
    chain=1536727068981429685321,
)

my_other_account = 3037460148268451690947737348211282823053350676597561718645135357343745417603

# Init the ETH token contract from its address
token_address = "0x07394cbe418daa16e42b87ba67372d4ab4a5df0b05c6e554d158458ce245bc10"
token_contract = Contract.from_address_sync(token_address, account_client)

receipt = token_contract.functions["transfer"].invoke_sync(my_other_account, 100, max_fee=100)
print(receipt)  # Does not get here
```

## Context (Environment)

Trying to interact with the ETH contract (which is a proxy)
Also, trying to use my Argent X account (which is a proxy)
In general, I have trouble with proxy contracts

## Detailed Description

Fetch ABI from proxy first by calling `implementation()`

cptartur:

Added support for proxy contracts in `Contract.from_address`. 

Closes #128 

Proxies are not supported

If `resolve_proxies=True`, `Contract.from_address` will recursively follow proxy contracts to the base implementation. Returned `Contract` will then have a address of the first proxy but abi of implementation allowing for using `Contract.functions` with this methods directly.

Also added a interface for `ProxyCheck` that provides methods for checking whether contract is a proxy or not and accessing a base implementation address, this allows for support for multiple types of proxy contracts as there isn't a uniform standard currently.

Added default proxy checks for Argent and OpenZeppelin proxy contracts.

No

## **Other information**

n/a


cptartur:
Will probably need to be rebased after #132 is merged.

starknet_py/contract.py
```diff
@@ -370,9 +376,17 @@ def functions(self) -> FunctionsRepository:
     def address(self) -> int:
         return self._data.address
 
+    @property
+    def data(self) -> ContractData:
```
Arcticae, line 380: no need to write properties like this, you can do just `self.data = ...` and `self.client = ...` if you want this field to be "public"

starknet_py/contract.py
```diff
@@ -370,9 +376,17 @@ def functions(self) -> FunctionsRepository:
...
+        return self._data
+
     @staticmethod
     async def from_address(
-        address: AddressRepresentation, client: Client
+        address: AddressRepresentation,
+        client: Client,
+        resolve_proxies: bool = False,
```
Arcticae, line 387: This adds too much overhead - i'd rather have a typed dict named `proxy_config` with some reasonable defaults

starknet_py/contract.py
```diff
@@ -381,10 +395,71 @@ async def from_address(
...
+            client=client,
+            proxy_checks=proxy_checks if resolve_proxies else [],
+            max_steps=max_steps,
+            step=1,
+        )
+
+        return Contract(address=address, abi=contract.data.abi, client=client)
+
+    @staticmethod
+    async def _from_address(
```
Arcticae, line 420: it's confusing to have 2 different methods - one public and one private named the same

starknet_py/proxy_check.py
```diff
@@ -0,0 +1,42 @@
...
+
+    async def implementation_address(self, contract: "Contract") -> int:
+        res = await contract.functions["get_implementation"].call()
+        return res[0]
+
+
+class OpenZeppelinProxyCheck(ProxyCheck):
+    def __init__(self):
+        self.storage_key = (
+            1662968741416162122709795668327058413808168082301235317239158738933362048792
```
Arcticae, line 32: how is this computed?

starknet_py/contract.py
```diff
@@ -381,10 +395,71 @@ async def from_address(
...
+        step: int,
+        max_steps: int,
+        processed_addresses: Optional[set] = None,
+        proxy_checks: Optional[List[ProxyCheck]] = None,
+    ) -> "Contract":
+        if processed_addresses is None:
+            processed_addresses = set()
+
+        if address in processed_addresses:
+            raise RecursionError("Proxy cycle detected while resolving proxies.")
```
Arcticae, line 433:
```suggestion
            raise RecursionError("Proxy cycle detected while resolving proxies")
```

starknet_py/contract.py
```diff
@@ -381,10 +395,71 @@ async def from_address(
...
+
+        if step > max_steps:
+            raise RecursionError("Too many redirects while resolving proxies")
```
Arcticae, line 436: i wouldn't call those redirects per se, let's just say we exceeded the number of steps while resolving proxies

Arcticae (changes requested):

starknet_py/proxy_check.py
cptartur: https://starknet.io/docs/hello_starknet/cli.html#get-storage-at

Should I add this calculation to the constructor?

Arcticae: yeaah probably. is this memory access something we want to ditch in the future? if that's correct please make a task to correct this and mark it here

Arcticae: as FIXME

starknet_py/contract.py
```diff
@@ -381,10 +416,74 @@ async def from_address(
...
+            proxy_checks=proxy_checks,
+            max_steps=max_steps,
+            step=1,
+        )
+
+        return Contract(address=address, abi=contract.data.abi, client=client)
+
+    @staticmethod
+    async def _create_contract_from_address(
+        # pylint: disable=too-many-arguments
```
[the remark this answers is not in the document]
cptartur: The problem with too many arguments comes from the fact that this method is recursive and `steps` and `processed_address` have to be passed every iteration. I'm not sure if it would make much sense to extract this method.

starknet_py/contract.py
```diff
@@ -368,11 +379,35 @@ def functions(self) -> FunctionsRepository:
...
+
+        If list of :class:`starknet_py.proxy_check.ProxyCheck` are provided, will use provided instances.
+        """
 
     @staticmethod
     async def from_address(
-        address: AddressRepresentation, client: Client
+        address: AddressRepresentation,
+        client: Client,
+        proxy_config: Optional[ProxyConfig] = None,
```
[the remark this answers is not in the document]
cptartur: This, functionally, would not be any different from how it is now though, right? With what I've written `proxy_checks = True` uses default config, `False` doesn't resolve proxies and if any config is given, that's what is used (and it's still possible to provide partial config).

starknet_py/contract.py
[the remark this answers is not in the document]
Arcticae: Contract being a part of the API, having those methods, also static ones doesn't help with the clarity of code. If we extracted this to a factory - you'd notice that you can again use instance fields again which is more OOP-like.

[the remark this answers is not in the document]
cptartur: Alright, I can write a factory for creating contracts from address.

starknet_py/proxy_check.py
cptartur: So while this is a imperfect solution, it's uncertain if/when OZ adds a simpler way of checking if contract is a proxy. 
If this PR is merged I'll make a issue to indicate this need to be fixed at some point. I'd rather not add FIXME as it will probably cause our checks to fail.

starknet_py/contract.py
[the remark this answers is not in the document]
cptartur: `proxy_config` is now of type `Union[bool, ProxyConfig]`. Optional is not needed since the default value is `False` and not `None`

starknet_py/proxy_check.py
```diff
@@ -40,3 +40,7 @@ async def implementation_address(self, contract: "Contract") -> int:
             contract_address=contract.address, key=self.storage_key
         )
         return parse_address(res)
+
+
+def generate_oz_storage_key():
```
Arcticae, line 45: It could be a const TBH, no need to re-generate it a couple of times

PR #143, opened 2022-05-11
Add parameter for changing between testnet and devnet in e2e tests

### LINKED ISSUE #139: Running tests on starknet testnet

The problem comes from PR #131 where we have functionality that should be tested on testnet, but as correctly pointed out by @Arcticae these tests probably shouldn't be run on every commit as they introduce significant overhead.

Some change in how these tests are run should be introduced.

Here are some ideas:
* Introduce manually triggered task that runs tests on testnet (most flexible, but requires most maintnance)
* Run testnet tests only on commits to development and master (automatic but may result in detecting problems after merge)
* Run testnet tests just before merging PRs (most optimal, but I don't know if it's event possible)

cptartur:

Added a `--net` parameter to e2e tests that allows changing between running tests on "devnet" and "testnet".

Will require further verification as running tests on testnet is super slow.

Should close #139 

Test can only be run on devnet

Test can be run on devnet and testnet

No

### DIFF
--- a/starknet_py/tests/e2e/conftest.py
+++ b/starknet_py/tests/e2e/conftest.py
@@ -6,6 +6,15 @@ from contextlib import closing
 import pytest
 
 
+def pytest_addoption(parser):
+    parser.addoption(
+        "--net",
+        action="store",
+        default="devnet",
+        help="Network to run tests on: possible 'testnet', 'devnet'",
+    )
+
+
 def get_available_port() -> int:
     with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
         sock.bind(("", 0))
@@ -13,8 +22,7 @@ def get_available_port() -> int:
         return sock.getsockname()[1]
 
 
-@pytest.fixture(scope="module", autouse=True)
-def run_devnet():
+def start_devnet():
     devnet_port = get_available_port()
 
     command = [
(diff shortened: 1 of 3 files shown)


cptartur:
We should probably also add a option (perhaps a pytest mark) for some test to be run **only** on devnet/testnet

starknet_py/tests/e2e/conftest.py
```diff
@@ -29,5 +37,35 @@ def run_devnet():
...
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--net") == "testnet":
+        skip_devnet = pytest.mark.skip()
+        for item in items:
+            if "run_on_testnet" not in item.keywords:
+                item.add_marker(skip_devnet)
+    else:
+        skip_testnet = pytest.mark.skip()
+        for item in items:
+            if "run_on_testnet" in item.keywords:
```
Solpatium, line 70: Should this if be the same as above? Also, repeated code.

starknet_py/tests/e2e/conftest.py
```diff
@@ -29,5 +37,35 @@ def run_devnet():
...
+                item.add_marker(skip_testnet)
```
Solpatium, line 71: Simplify this

starknet_py/tests/e2e/conftest.py
```diff
@@ -29,5 +37,28 @@ def run_devnet():
...
+@pytest.fixture(scope="module", autouse=True)
+def run_devnet():
+    devnet_port, proc = start_devnet()
     yield f"http://localhost:{devnet_port}"
     proc.kill()
+
+
+@pytest.fixture
+def run_net(request):
+    return request.config.getoption("--net")
```
Solpatium, line 52: This might return "all". What it the purpose of `run_net`?

Solpatium (approved):

PR #148, opened 2022-05-12
Full node support and refactor

cptartur:

This PR refactors `Client` to be less dependent on gateway client from cairo-lang library and adds support for full node rpc:
 * Added BaseClient ABC base for all clients
 * Added dataclasses for objects returned by client and schema validation for respective types
 * Added own gateway client
 * Added rpc client
 * Modified parts of library that depended on old client

Client based on cairo-lang's gateway client is used

No support for rpc calls

There are multiple interface changes that will be listed later and will need to be included in a migration guide

## **Other information**

This is a work in progress


starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,328 @@
...
+    StarknetBlockSchema,
+    TransactionReceiptSchema,
+    SentTransactionSchema,
+)
+from starknet_py.net.models import StarknetChainId, chain_from_network
+from starknet_py.net.networks import Network, net_address_from_net
+
+
+class GatewayClient(BaseClient):
+    def __init__(self, net: Network, chain: StarknetChainId = None):
```
cptartur, line 47: Original Client included parameter `retry_config` as required by cairo-lang gateway client. Should we introduce a similiar parameter?

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,328 @@
...
+            method_name="get_block", params=block_identifier
+        )
+        return StarknetBlockSchema().load(res, unknown=EXCLUDE)
+
+    async def get_state_update(
+        self,
+        block_hash: Optional[Union[int, str]] = None,
+        block_number: Optional[int] = None,
+    ) -> BlockState:
+        # TODO implement
```
cptartur, line 94: Should we support get_state_update if it wasn't supported in previous client?

cptartur:
What definitely will be changed: exceptions thrown by GatewayClient and FullNodeClient should be uniform, defined in something like `BaseExceptions`. 

Also I'm open to any renaming suggestions, as many classes/files don't have particularity great names. 

Last thing, if there are any suggestions for thing that you know will have to be changed/reworked, I'm open for any comments.

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,328 @@
...
+            },
+            **block_identifier,
+        }
+
+        res = await self._feeder_gateway_client.call(
+            method_name="get_code", params=params
+        )
+
+        if len(res["bytecode"]) == 0:
+            raise NoContractFoundError(
```
cptartur, line 161: Is defining errors like this in separate file and then having them in docstrings for`BaseClient` a good idea>

starknet_py/net/base_client.py
```diff
@@ -0,0 +1,187 @@
...
+    SentTransaction,
+    InvokeFunction,
+    StarknetTransaction,
+    ContractDefinition,
+    TransactionStatus,
+)
+
+
+@dataclass
+class BlockHashIdentifier:
```
Solpatium, line 20: Not the best name :(

starknet_py/net/base_client.py
```diff
@@ -0,0 +1,187 @@
...
+    block_hash: int
+    index: int
+
+
+@dataclass
+class BlockNumberIdentifier:
+    block_number: int
+    index: int
```
Solpatium, line 28: Isn't block number the same as index?

starknet_py/net/client_errors.py
```diff
@@ -0,0 +1,33 @@
...
+        self,
+        block_hash: Optional[Union[int, str]] = None,
+        block_number: Optional[int] = None,
+    ):
+        if block_hash is not None and block_number is not None:
+            raise ValueError(
+                "Block hash and number cannot be provided at the same time"
+            )
+        if block_hash is None and block_number is None:
+            raise ValueError("One of block hash or number must be provided")
```
Solpatium, line 25: Do we need to check that? In the first case we lose information about the block hash/number by throwing a different error.

starknet_py/net/gateway_client.py
Solpatium: I think it is pointless, retries should be handled on user's side.

starknet_py/net/gateway_client.py
Solpatium: Why not?

starknet_py/net/gateway_client.py
Solpatium: I don't see docstrings here. But yes, documenting them and exposing them is a good idea :)

starknet_py/net/gateway_client.py
cptartur: Sure, guess there's not reason for not supporting it now

starknet_py/net/gateway_client.py
cptartur: I've pushed an first version of custom errors in `client_errors.py`. There's one thing to consider though, RPC and gateway throw completely different errors so I think it'd be most resonable to wrap _most of these_ in uniform `ClientError` and let the user handle them.

Instead we'd have to write at least a dozen of custom exceptions and then handle them.

starknet_py/net/base_client.py
cptartur: What name would you suggest instead?

starknet_py/net/client_errors.py
cptartur: I haven't consider that, I think I should move this logic out of the constructor then

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,328 @@
...
+        )
+
+        res = await self._feeder_gateway_client.post(
+            method_name="call_contract",
+            params=block_identifier,
+            payload=InvokeFunction.Schema().dump(invoke_tx),
+        )
+
+        # TODO should we have better type validation here?
+        return [int(v, 16) for v in res["result"]]
```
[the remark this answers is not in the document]
Solpatium: I don't think so, this code **does** validate the output:
1. Will throw if result is not a key of res.
2. `int` throws if value provided is not in the right format.

starknet_py/net/gateway_client.py
Solpatium: 🤔  shouldn't we map them?

starknet_py/net/base_client.py
Solpatium: So this is in fact something related to tx, right?

starknet_py/net/base_client.py
Solpatium: Why do we need it?

starknet_py/net/base_client.py
cptartur: This is used as a parametr for some client methods since RPC allows querying transactions not only by block hash, but also by pairs like this one.

Solpatium: We should create a type
```
Hash = Union[string, int]
TxIdentifier = Union[Hash, BlockHashIdentifier, BlockNumberIdentifier]
```
and reuse it.

starknet_py/net/base_client.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+
+    @abstractmethod
+    async def get_state_update(
+        self,
+        block_hash: Optional[Union[int, str]] = None,
+        block_number: Optional[int] = None,
+    ) -> BlockState:
+        """
+        # TODO expand specs of this class, create proper dataclass
```
Solpatium, line 53: 👁️

starknet_py/net/base_client_schemas.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class BlockHashIdentifierSchema(Schema):
+    block_hash = fields.Integer()
+    index = fields.Integer()
+
+
+class BlockNumberIdentifierSchema(Schema):
+    block_number = fields.Integer()
+    index = fields.Integer()
```
Solpatium, line 12: Why do we need schemas for **input**? I think that if we specify expected type and someone provides invalid value it is their problem.

starknet_py/net/client_errors.py
Solpatium: I think we just can take both. In this case `block_hash` will have a priority since `identifier = block_hash or block_number`. This is not a problem since this is something that **we** create, library users won't ever create this error..

starknet_py/net/common_schemas/common_schemas.py
```diff
@@ -0,0 +1,104 @@
...
+        return value.name if value is not None else ""
+
+    def _deserialize(
+        self,
+        value: Any,
+        attr: Union[str, None],
+        data: Union[Mapping[str, Any], None],
+        **kwargs,
+    ) -> TransactionStatus:
+        # should we be treating NOT_RECEIVED as UNKNOWN?
```
Solpatium, line 52: I think so

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,267 @@
...
+            raise ValueError("Block_hash or block_number must be provided.")
+
+        res = None
+        if block_hash is not None:
+            res = await self.rpc_client.call(
+                method_name="getBlockByHash",
+                params={
+                    "block_hash": str(hex(block_hash))
+                    if isinstance(block_hash, int)
+                    else block_hash,
```
Solpatium, line 73: Together with Hash type there should be a helper to do this type of logic.

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,267 @@
...
+    ) -> StarknetBlock:
+        if block_hash is not None and block_number is not None:
+            raise ValueError(
+                "Block_hash and block_number parameters are mutually exclusive."
+            )
+
+        if block_hash is None and block_number is None:
+            raise ValueError("Block_hash or block_number must be provided.")
+
+        res = None
```
Solpatium, line 66: replace this with 
```
if block_hash:

else if block_number:

else:
 raise
```

starknet_py/net/full_node_client.py
Solpatium: like `to_hex`

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,267 @@
...
+        pass
+
+    async def get_storage_at(
+        self,
+        contract_address: Union[int, str],
+        key: int,
+        block_hash: Optional[Union[int, str]] = None,
+        block_number: Optional[int] = None,
+    ) -> int:
+        if block_hash is None:
```
Solpatium, line 99: block number is ignored in this function

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,267 @@
...
+            block_hash_identifier = BlockHashIdentifierSchema().load(tx_identifier)
+            res = await self._get_transaction_by_block_hash(block_hash_identifier)
+        except ValidationError as ex:
+            error_message = error_message or str(ex)
+
+        try:
+            block_number_identifier = BlockNumberIdentifierSchema().load(tx_identifier)
+            res = await self._get_transaction_by_block_number(block_number_identifier)
+        except ValidationError as ex:
+            error_message = error_message or str(ex)
```
Solpatium, line 139: This doesn't look good.

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,267 @@
...
+
+    async def get_code(
+        self,
+        contract_address: Union[int, str],
+        block_hash: Optional[Union[int, str]] = None,
+        block_number: Optional[int] = None,
+    ) -> ContractCode:
+        res = await self.rpc_client.call(
+            method_name="getCode",
+            params={"contract_address": str(hex(contract_address))}
```
Solpatium, line 202: `TypeError: 'str' object cannot be interpreted as an integer`

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,267 @@
...
+        return ContractCodeSchema().load(res, unknown=EXCLUDE)
+
+    async def call_contract(
+        self,
+        invoke_tx: InvokeFunction,
+        block_hash: Optional[Union[int, str]] = None,
+        block_number: Optional[int] = None,
+    ) -> List[int]:
+        if block_hash is None:
+            raise ValueError("Block_hash must be provided when using FullNodeClient")
```
Solpatium, line 215: what about block_number?

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,267 @@
...
+            "method": f"starknet_{method_name}",
+            "params": params,
+            "id": 0,
+        }
+
+        async with aiohttp.ClientSession() as session:
+            async with session.post(self.url, json=payload) as request:
+                result = await request.json()
+                if "result" not in result:
+                    await self.handle_full_node_exceptions(result)
```
Solpatium, line 258: `handle_full_node_exceptions` is not async.

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,315 @@
...
+        )
+
+        if len(res["bytecode"]) == 0:
+            raise ContractNotFoundError(
+                f"No contract found with following identifier {block_identifier}"
+            )
+
+        return ContractCodeSchema().load(res, unknown=EXCLUDE)
+
+    async def wait_for_tx(
```
Solpatium, line 168: This should be a part of `BaseClient`.

starknet_py/net/client_errors.py
cptartur: I guess we can remove this type of checks. I thought it may be good to be explicit that two of these parameters at the same time are not supported.

starknet_py/net/client_models.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    Dataclass representing details of sent transaction
+    """
+
+    hash: int
+    status: TransactionStatus
+    events: List[Event]
+    l2_to_l1_messages: List[L2toL1Message]
+    l1_to_l2_consumed_message: Optional[L1toL2Message] = None
+    block_number: Optional[int] = None
```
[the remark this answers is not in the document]
cptartur: This is how it's done in cairo-lang, I guess we can do that. There will be a problem with full node client, as realistically it will only be able to return base `Transaction` still.

starknet_py/net/full_node_client.py
cptartur: Or should we throw if it's provided?

starknet_py/net/full_node_client.py
cptartur: I agree it's not the most elegant way to write but here's a problem: user passes a block_hash_identifier or block_number_identifier, which is a typeddict. So for python it's just a dict so to check what private method to call, we either have to use these schemas or manually check the keys in provided dict.

Or we can change this parameters entirely (how?)

starknet_py/net/full_node_client.py
cptartur: RPC doesn't support block number in this method

starknet_py/net/common_schemas/common_schemas.py
```diff
@@ -0,0 +1,104 @@
...
+
+class Felt(fields.Field):
+    # TODO test that serialization and deserialization is correct
+    """
+    Field that serializes int to felt (hex encoded string)
+    """
+
+    def _serialize(self, value: Any, attr: str, obj: Any, **kwargs):
+        if value is None:
+            return None
```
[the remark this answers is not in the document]
cptartur: What do you mean by that?

starknet_py/net/base_client.py
cptartur: Dataclasses have been converted to TypeDicts now, they're still used as parameters for get_transaction

cptartur: I'm open on suggestions for how to replace it though

starknet_py/net/base_client.py
cptartur: We have a `Hash` type now, but what about the first question in this thread?

Solpatium (changes requested):

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,264 @@
...
+            return {"blockHash": convert_to_felt(block_hash)}
+
+        if block_number is not None:
+            return {"blockNumber": block_number}
+
+        return {}
+
+
+# TODO rename
+class StarknetClient:
```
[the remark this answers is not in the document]
cptartur: I though it'd make more sense to separate actual rpc/rest api clients from our higher level `Client`s implementations. It could be renamed but I don't think it'd be good to make it a part of `Client`s itself.

starknet_py/net/full_node_client.py
cptartur: That makes sense. I think I focused too much on making the interface as inclusive as possible and these kind of problems could be resolved with just simplifying the interface. I'll try this approach for this and maybe some other methods.

starknet_py/net/base_client.py
cptartur: It has been removed

starknet_py/net/base_client.py
cptartur: Not applicable anymore. Dataclass removed entirely.

starknet_py/net/client_errors.py
cptartur: Second ValueError has now been removed. I've retained the first one though, because I think we should actually validate if any of the optional parameters was provided. What do you think?

starknet_py/net/base_client_schemas.py
cptartur: Removed

starknet_py/net/common_schemas/common_schemas.py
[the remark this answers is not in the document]
cptartur: I'm not sure why I've added this check. Removed return None entirely

starknet_py/net/full_node_client.py
cptartur: Added en exception, but I'm unsure if handling it this way is the best for user experience. We could theoretically separate this into two methods and have only one in the `BaseSigner`. Which do you think is a better approach in that case?

starknet_py/net/full_node_client.py
cptartur: Same as above

requirements.txt
```diff
@@ -0,0 +1,116 @@
+aiohttp==3.8.1; python_version >= "3.6" and python_version < "4"
```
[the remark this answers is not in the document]
cptartur: Requirements.txt is now autogenerated in pre-commit hook in starknet.py, so that's just that

starknet_py/net/client_utils.py
```diff
@@ -0,0 +1,11 @@
+from starknet_py.net.client_models import Hash
+
+
+def convert_to_felt(value: Hash) -> str:
```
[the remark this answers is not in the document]
cptartur: It's used in both clients

starknet_py/net/common_schemas/common_schemas.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if value not in values:
+            raise ValidationError("Invalid TransactionStatus enum key")
+
+        return TransactionStatus(value)
+
+
+class BlockStatusField(fields.Field):
+    def _serialize(self, value: Any, attr: str, obj: Any, **kwargs):
+        # TODO should we serialize to string?
```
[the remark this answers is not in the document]
cptartur: I think this question is no longer valid: We're using string values for enums now so both enum .value and .name will be the same. Previously we were representing enum values as numbers hence my concern if we should serialize to something else.

starknet_py/net/common_schemas/common_schemas.py
```diff
@@ -0,0 +1,103 @@
...
+        attr: Union[str, None],
+        data: Union[Mapping[str, Any], None],
+        **kwargs,
+    ) -> TransactionType:
+        # TODO maybe simplify
+        enum_values = {v.name: k for k, v in enumerate(TransactionType)}
+
+        if value not in enum_values:
+            # TODO should we use different default type?
+            return TransactionType.INVOKE
```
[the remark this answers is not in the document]
cptartur: Probably not needed, default values should be handled on schema side with `load_default` or something similiar.

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,235 @@
+import typing
```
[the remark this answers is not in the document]
cptartur: Definitely, also naming of some files could probably be improved. I want to wait with these kind of refactors until other things are done though so it doesn't break ongoing discussions.

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,235 @@
...
+            )
+            return StarknetBlockSchema().load(res, unknown=EXCLUDE)
+        elif block_number is not None:
+            res = await self.rpc_client.call(
+                method_name="getBlockByNumber",
+                params={"block_number": block_number, "requested_scope": "FULL_TXNS"},
+            )
+            return StarknetBlockSchema().load(res, unknown=EXCLUDE)
+        else:
+            raise ValueError("Block_hash or block_number must be provided.")
```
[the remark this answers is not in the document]
cptartur: This has been split into separate methods.

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,235 @@
...
+            params={
+                "contract_address": convert_to_felt(invoke_tx.contract_address),
+                "entry_point_selector": convert_to_felt(invoke_tx.entry_point_selector),
+                "calldata": [convert_to_felt(i) for i in invoke_tx.calldata],
+                "block_hash": convert_to_felt(block_hash),
+            },
+        )
+        return [int(i, 16) for i in res["result"]]
+
+    async def add_transaction(self, tx: StarknetTransaction) -> SentTransaction:
```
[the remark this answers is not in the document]
cptartur: I'll consider adding it in this PR: we have now way of testing it though, since it's not supported in devnet atm.

starknet_py/net/client_models.py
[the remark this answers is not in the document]
cptartur: Don't you think it will eventually create a complicate structure of dataclasses? 

Because realistically we should have receipt for accepted transaction, pending, rejected etc.

starknet_py/net/full_node_client.py
Solpatium: Hmm can you inherit a method in python and add additional parameters that would be recognised by type checker?

starknet_py/net/gateway_client.py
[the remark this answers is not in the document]
Solpatium: What's the potential use case for this indirection?

starknet_py/net/client_models.py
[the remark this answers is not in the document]
Solpatium: Ok, let's leave it like this for now, but ideally we should properly model the result.

starknet_py/net/client_errors.py
Solpatium: Can we create some helpers for behaviours like this? In here it could be `requireOneOf(NAME_TO_VALUE)`.

starknet_py/net/full_node_client.py
cptartur: From what I see it is possible to add addition parameters to the inherited method. I'm not sure if that's the ideal solution though?

starknet_py/net/gateway_client.py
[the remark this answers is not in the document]
cptartur: Renamed to <>HttpClient

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,261 @@
...
+
+        async with aiohttp.ClientSession() as session:
+            async with session.get(address, params=params or {}) as request:
+                if request.status != 200:
+                    raise ClientError(
+                        code=str(request.status), message=await request.text()
+                    )
+                return await request.json()
+
+    async def post(
```
Solpatium, line 248: I think you can use `.request` https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession.trace_config to remove repeated code.

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,261 @@
...
+
+        async with aiohttp.ClientSession() as session:
+            async with session.post(
+                address, params=params or {}, json=payload
+            ) as request:
+                if request.status != 200:
+                    raise ClientError(
+                        code=str(request.status), message=await request.text()
+                    )
+                return await request.json(content_type=None)
```
Solpatium, line 261: Why do you specify `content_type=None` here?

Solpatium (changes requested):

starknet_py/contract.py
```diff
@@ -240,11 +244,11 @@ async def invoke(
         tx = self._make_invoke_function(signature=signature)
         response = await self._client.add_transaction(tx=tx)
 
-        if response["code"] != StarkErrorCode.TRANSACTION_RECEIVED.name:
+        if response.code != StarkErrorCode.TRANSACTION_RECEIVED.name:
             raise Exception("Failed to send transaction. Response: {response}.")
 
         invoke_result = InvokeResult(
-            hash=response["transaction_hash"],  # noinspection PyTypeChecker
+            hash=response.hash,  # noinspection PyTypeChecker
```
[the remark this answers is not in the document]
cptartur: I think so. RPC write api specs says both invoke and declare transactions return at least transaction_hash. And that's also the case with gateway if transaction status is at least `RECEIVED`.

starknet_py/net/common_schemas/common_schemas.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> TransactionStatus:
+        values = [v.value for v in TransactionStatus]
+
+        if value == "NOT_RECEIVED":
+            return TransactionStatus.UNKNOWN
+
+        if value not in values:
+            raise ValidationError("Invalid TransactionStatus enum key")
+
+        return TransactionStatus(value)
```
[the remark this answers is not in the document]
cptartur: It throws a ValueError, e.g. `ValueError: 'WRONG_NAME' is not a valid TransactionStatus`. Marshmallow uses ValidationError for it's schema errors so i think it better to be uniform with that though.

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,228 @@
...
+                },
+            )
+            return StarknetBlockSchema().load(res, unknown=EXCLUDE)
+        if block_number is not None:
+            res = await self._client.call(
+                method_name="getBlockByNumber",
+                params={"block_number": block_number, "requested_scope": "FULL_TXNS"},
+            )
+            return StarknetBlockSchema().load(res, unknown=EXCLUDE)
+        raise ValueError("Block_hash or block_number must be provided.")
```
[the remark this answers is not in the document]
cptartur: They do get ignored in current full node implementation. JSON-RPC general specs doesn't specify what to do in case of extra parameters.

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,228 @@
...
+
+    async def get_state_update(
+        self,
+        block_hash: Optional[Union[int, str]] = None,
+        block_number: Optional[int] = None,
+    ) -> BlockStateUpdate:
+        if block_number is not None:
+            raise ValueError(
+                "Block_number is not supported in this method when using FullNodeClient"
```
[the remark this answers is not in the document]
cptartur: How about `call_contract` and `get_storage_at` then? It's the same case with them.

starknet_py/net/gateway_client.py
cptartur: Gateway was sometimes incorrectly setting Content-Type to `text/html` in responses and aiohttp `.json()` fails by default if `Content-Type` is not set to `application/json`.

This seems to no longer be the case so I've removed it.

starknet_py/net/base_client.py
```diff
@@ -0,0 +1,187 @@
...
+class BlockNumberIdentifier(TypedDict):
+    block_number: int
+    index: int
+
+
+class BaseClient(ABC):
+    @abstractmethod
+    async def get_block(
+        self,
+        block_hash: Optional[Union[int, str]] = None,
```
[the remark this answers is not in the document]
cptartur: Fixed

starknet_py/net/full_node_client.py
```diff
@@ -0,0 +1,228 @@
...
+class RpcHttpClient:
+    def __init__(self, url):
+        self.url = url
+
+    async def call(self, method_name: str, params: dict) -> dict:
+        payload = {
+            "jsonrpc": "2.0",
+            "method": f"starknet_{method_name}",
+            "params": params,
+            "id": 0,
```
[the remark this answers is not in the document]
cptartur: I'm not aware of any popular projects like this.

starknet_py/net/full_node_client.py
cptartur: Fixed by simplifying interface and only adding additional parameters to GatewayClient.

starknet_py/net/full_node_client.py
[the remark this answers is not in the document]
Solpatium: Ok, we can use our implementation

starknet_py/net/common_schemas/common_schemas.py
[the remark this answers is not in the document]
Solpatium: I forgot this is a schema.

starknet_py/net/full_node_client.py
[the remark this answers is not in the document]
cptartur: BaseClient has been updated

starknet_py/net/common_schemas/common_schemas.py
```diff
@@ -26,10 +26,12 @@ def _deserialize(
         data: Union[Mapping[str, Any], None],
         **kwargs,
     ):
+        if not isinstance(value, str) or not value.startswith("0x"):
+            raise ValidationError("Invalid felt")
```
Solpatium, line 30: Please add value provided, like `f"Invalid felt: [{value}]"`

starknet_py/net/full_node_client.py
```diff
@@ -120,9 +105,16 @@ async def get_transaction(
         )
         return TransactionSchema().load(res, unknown=EXCLUDE)
 
-    async def get_transaciton_by_block_hash_and_index(
+    async def get_transaction_by_block_hash(
```
Solpatium, line 108: I know we decided to split get_transaction_* to two methods. Do you remember the reason?

Solpatium (changes requested):
Almost there 🔥

starknet_py/net/gateway_client.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class GatewayHttpClient:
+    def __init__(self, url):
+        self.url = url
+
+    async def call(self, method_name: str, params: Optional[dict] = None) -> dict:
+        address = f"{self.url}/{method_name}"
+
+        async with aiohttp.ClientSession() as session:
```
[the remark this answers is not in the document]
cptartur: Done

starknet_py/net/full_node_client.py
cptartur: Actually get_transaction is more complicated. We have get_transaction (by hash) that's in both gateway and rpc client, and rpc client also has get_transaction_by_block_hash(_and_index) and get_transaction_by_block_number(_and_index) that are non existent on gateway client.

starknet_py/net/common_schemas/common_schemas.py
cptartur: Added

starknet_py/net/full_node_client.py
```diff
@@ -204,12 +214,23 @@ async def call(self, method_name: str, params: dict) -> dict:
...
-        async with aiohttp.ClientSession() as session:
-            async with session.post(self.url, json=payload) as request:
-                result = await request.json()
-                if "result" not in result:
-                    self.handle_error(result)
-                return result["result"]
+        async with self.http_session() as session:
+            return await self._make_request(session=session, payload=payload)
+
+    def http_session(self) -> ClientSession:
```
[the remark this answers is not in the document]
Solpatium: Or just create a helper function that returns the session context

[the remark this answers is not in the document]
cptartur: Replaced with separate http client

Solpatium (commented):
What's with `test_signing_single_transaction.py`?

PR #150, opened 2022-05-13
get_balance for account_client

war-in:

The account's balance of token can be easily checked now.

The account can't check its own balance.

account_client.get_balance(token_address)
Gets balance of specified (or default) token

No


starknet_py/net/account/account_client.py
```diff
@@ -73,6 +75,34 @@ async def _get_nonce(self) -> int:
...
+                              If not specified it will be mainnet or testnet payment token address.
+        :return: Account's balance of token.
+        """
+        if token_address is None:
+            if self.net == MAINNET:
+                token_address = MAINNET_ETH_CONTRACT
+            elif self.net == TESTNET:
+                token_address = TESTNET_ETH_CONTRACT
+            else:
+                raise BadRequest(200, "Specify token_address for custom url.")
```
cptartur, line 92: I think it would be better to use standard python error here, or write a custom one instead of using internal errors from cairo lang library.

starknet_py/net/account/account_client.py
```diff
@@ -73,6 +75,34 @@ async def _get_nonce(self) -> int:
...
 
+    async def get_balance(self, token_address: AddressRepresentation = None) -> int:
+        """
+        Checks account's balance of specified token._
+
+        :param token_address: Address of the ERC20 contract.
+                              If not specified it will be mainnet or testnet payment token address.
+        :return: Account's balance of token.
+        """
+        if token_address is None:
```
cptartur, line 86: Maybe we should consider making token_address a property of AccountClient?

I'm not sure if there's use case where someone would want to use multiple token addresses per account but in that case we could also allow for passing a different `token_address` as a parameter to `get_balance` that would override address set on AccountClient.

starknet_py/constants.py
```diff
@@ -9,3 +9,6 @@
 
 ROOT_DIR = os.path.realpath(os.path.join(os.path.dirname(__file__), "."))
 DEFAULT_CPP_LIB_PATH = os.path.join(ROOT_DIR, "utils", "crypto")
+
+MAINNET_ETH_CONTRACT = "0xc662c410C0ECf747543f5bA90660f6ABeBD9C8c4"
```
cptartur, line 13: Maybe it would be good to document where these addresses come from? As a comment or in docs.

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -51,3 +58,60 @@ async def test_error_when_tx_signed(run_devnet):
...
+
+    with pytest.raises(BadRequest) as b_req:
+        await acc_client.get_balance()
+
+    assert "Specify token_address for custom url." in str(b_req.value)
+
+
+@pytest.mark.asyncio
+async def test_balance_when_token_specified(run_devnet):
+    acc_client = await AccountClient.create_account(
```
cptartur, line 77: DevnetClientFactory should be used here

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -51,3 +58,60 @@ async def test_error_when_tx_signed(run_devnet):
...
+
+    assert balance == 200
+
+
+@pytest.mark.asyncio
+async def test_default_token_address():
+
+    acc_client_testnet = AccountClient("0x123", KeyPair(123, 456), TESTNET)
+    acc_client_mainnet = AccountClient("0x321", KeyPair(456, 123), MAINNET)
+
```
cptartur, line 97: Would it be possible to `@pytest.mark.parametrize` this test with respective clients and expected eth contract address? Now we are testing two accounts at the same time and it makes the test a little bit confusing

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -51,3 +58,60 @@ async def test_error_when_tx_signed(run_devnet):
...
+    deployment_result = await deployment_result.wait_for_acceptance()
+    erc20_contract = deployment_result.deployed_contract
+
+    balance = await acc_client.get_balance(erc20_contract.address)
+
+    assert balance == 200
+
+
+@pytest.mark.asyncio
+async def test_default_token_address():
```
cptartur, line 93:
```suggestion
async def test_get_balance_default_token_address():
```

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -51,3 +58,60 @@ async def test_error_when_tx_signed(run_devnet):
     assert "Adding signatures to a signer tx currently isn't supported" in str(
         t_err.value
     )
+
+
+@pytest.mark.asyncio
+async def test_error_when_token_not_specified(run_devnet):
```
cptartur, line 64:
```suggestion
async def test_get_balance_throws_when_token_not_specified(run_devnet):
```

starknet_py/net/account/account_client.py
```diff
@@ -73,6 +75,34 @@ async def _get_nonce(self) -> int:
         )
         return nonce
 
+    async def get_balance(self, token_address: AddressRepresentation = None) -> int:
+        """
+        Checks account's balance of specified token._
```
cptartur, line 80:
```suggestion
        Checks account's balance of specified token.
```
Or is it on purpose?

starknet_py/net/account/account_client.py
```diff
@@ -73,6 +75,34 @@ async def _get_nonce(self) -> int:
...
+
+        :param token_address: Address of the ERC20 contract.
+                              If not specified it will be mainnet or testnet payment token address.
+        :return: Account's balance of token.
```
cptartur, line 84:
```suggestion
        :return: Token balance
```

cptartur (commented):
Good work, a couple of suggestions

starknet_py/net/account/account_client.py
```diff
@@ -73,6 +74,37 @@ async def _get_nonce(self) -> int:
...
 
+    def _get_default_token_address(self) -> str:
+        if self.net not in [TESTNET, MAINNET]:
+            raise ValueError(
+                "Token_address must be specified when using a custom net address"
+            )
+
+        return W_ETH_CONTRACT
+
+    async def get_balance(self, token_address: AddressRepresentation = None) -> int:
```
cptartur, line 85:
```suggestion
    async def get_balance(self, token_address: Optional[AddressRepresentation] = None) -> int:
```

starknet_py/net/account/account_client.py
```diff
@@ -73,6 +74,37 @@ async def _get_nonce(self) -> int:
...
+        """
+        Checks account's balance of specified token.
+
+        :param token_address: Address of the ERC20 contract.
+                              If not specified it will be wETH token address.
```
cptartur, line 90: Should it be `wETH` or `ETH`?

cptartur (commented):
Looks good to me, after addressing remaining comments, can be merged.

cptartur (approved):

PR #178, opened 2022-06-01
Feat/declare deploy account client

war-in:

Declare and deploy for account client

### DIFF
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -1,22 +1,26 @@
 from dataclasses import dataclass
-from typing import Dict, Optional
+from typing import Dict, Optional, List
 
 from starkware.crypto.signature.signature import (
     private_to_stark_key,
     get_random_private_key,
 )
+from starkware.starknet.definitions.fields import ContractAddressSalt
 from starkware.starknet.public.abi import get_selector_from_name
 from starkware.starknet.public.abi_structs import identifier_manager_from_abi
 from starkware.starknet.core.os.transaction_hash.transaction_hash import (
     calculate_transaction_hash_common,
     TransactionHashPrefix,
 )
+from starkware.starkware_utils.error_handling import StarkErrorCode
 
+from starknet_py.compile.compiler import Compiler, create_contract_class, StarknetCompilationSource
 from starknet_py.constants import FEE_CONTRACT_ADDRESS
+from starknet_py.net.models.transaction import Declare
 from starknet_py.utils.data_transformer.data_transformer import DataTransformer
 from starknet_py.net import Client
 from starknet_py.net.account.compiled_account_contract import COMPILED_ACCOUNT_CONTRACT
-from starknet_py.net.models import InvokeFunction, StarknetChainId, TransactionType
+from starknet_py.net.models import InvokeFunction, StarknetChainId, TransactionType, Deploy
 from starknet_py.net.networks import Network, MAINNET, TESTNET
 from starknet_py.utils.sync import add_sync_methods
 from starknet_py.utils.crypto.facade import message_signature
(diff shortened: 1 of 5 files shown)


starknet_py/net/account/account_client.py
```diff
@@ -179,6 +192,64 @@ async def add_transaction(
...
+                if salt is None
+                else salt,
+                contract_definition=class_hash,  # TODO: for now it is wrong
+                constructor_calldata=constructor_calldata,
+                version=version,
+            )
+        )
+
+        if res["code"] != StarkErrorCode.TRANSACTION_RECEIVED.name:
+            raise Exception("Transaction not received")
```
cptartur, line 249: Same as above

starknet_py/net/account/account_client.py
```diff
@@ -179,6 +192,64 @@ async def add_transaction(
...
 
+    async def declare(
+        self,
+        compilation_source: Optional[StarknetCompilationSource] = None,
+        compiled_contract: Optional[str] = None,
+        max_fee: Optional[int] = 0,
+        version: Optional[int] = 0,
+        search_paths: Optional[List[str]] = None,
+    ) -> dict:
+        if not compiled_contract and not compilation_source:
```
cptartur, line 203: Add a docstring please

starknet_py/net/account/account_client.py
```diff
@@ -179,6 +192,64 @@ async def add_transaction(
...
+        salt: Optional[int] = None,
+        version: Optional[int] = 0,
+    ) -> AddressRepresentation:
+        res = await self.add_transaction(
+            tx=Deploy(
+                contract_address_salt=ContractAddressSalt.get_random_value()
+                if salt is None
+                else salt,
+                contract_definition=class_hash,  # TODO: for now it is wrong
+                constructor_calldata=constructor_calldata,
```
cptartur, line 243: One thing that came to my mind now: To still support translating calldata between python and starknet, we need to have an ABI for the given class_hash, and use the same data transformation logic as we used in deploy before. To implement this we will need to first support #179.

cc @Solpatium

cptartur:
Since 0.9 deploy will be handled through a syscall with a class hash of a declared contract. To support this OZ/Argent will need to add support for this to their account client, and we will have to update account contract used in starknet.py. 

Also, old way of deploying contract will not be deprecated in 0.9 so we may consider supporting both versions.

starknet_py/net/account/account_client.py
```diff
@@ -179,6 +192,64 @@ async def add_transaction(
...
+                sender_address=self.address,
+                max_fee=max_fee,
+                signature=[],
+                nonce=0,
+                version=version,
+            )
+        )
+
+        if res["code"] != StarkErrorCode.TRANSACTION_RECEIVED.name:
+            raise Exception("Transaction not received")
```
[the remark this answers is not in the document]
war-in: It was done this way in the Client.deploy. Should I change it there too?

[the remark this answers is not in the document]
cptartur: I think so

starknet_py/net/account/account_client.py
```diff
@@ -179,6 +192,50 @@ async def add_transaction(
 
         return await super().add_transaction(await self._prepare_invoke_function(tx))
 
+    async def declare(
+        self,
+        compilation_source: Optional[StarknetCompilationSource] = None,
+        compiled_contract: Optional[str] = None,
+        version: int = 0,
+        search_paths: Optional[List[str]] = None,
```
cptartur, line 200: I think we should rename it to `cairo_path` so it's uniform with Compiler constructor.

cptartur (commented):
Good work. One think I'd consider is we should add some kind of deprecation warning to `Client.deploy` - this way of deploying contract will most likely be removed in future versions of starknet.

You may consider using [this](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) or [this](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning).

Arcticae:
LGTM codewise

cptartur (approved):

PR #182, opened 2022-06-02
Use BaseSigner in AccountClient, add default KeypairSigner

cptartur:


starknet_py/net/account/account_client.py
```diff
@@ -43,22 +31,15 @@ class AccountClient(Client):
     def __init__(
         self,
         address: AddressRepresentation,
-        key_pair: KeyPair,
         net: Network,
+        signer: BaseSigner,
```
Solpatium, line 35: I think we can accept either a signer o keypair that will be used with the default signer.

starknet_py/net/signer/keypair_signer.py
```diff
@@ -0,0 +1,61 @@
...
+class KeyPair:
+    private_key: int
+    public_key: int
+
+    @staticmethod
+    def from_private_key(key: int) -> "KeyPair":
+        return KeyPair(private_key=key, public_key=private_to_stark_key(key))
+
+
+class KeypairSigner(BaseSigner):
```
Solpatium, line 26: There's a lot of algorithms that involve a key pair. What if we named this DefaultSigner?

Solpatium (changes requested):

starknet_py/net/signer/keypair_signer.py
cptartur: I've considered that but I thought `DefaultSigner` isn't descriptive enough - can change it to that though if you think it would be better that way.

starknet_py/net/account/account_client.py
Solpatium: I think we may need to get rid of `create_account` in format that we use currently.

starknet_py/net/signer/keypair_signer.py
Solpatium: Then name it after signature method used in starknet

starknet_py/net/account/account_client.py
cptartur: How would we replace it? I think it's good to have a easy way of deploying an account through starknet.py.

starknet_py/net/signer/keypair_signer.py
cptartur: How about something like `EllipticCurveSigner`?

cptartur: Or `StarkCurveSigner`?

starknet_py/net/account/account_client.py
Solpatium: I guess it can stay for now, but it can't accept signer - it can only deploy the default account. Still, I would accept signer | keypair in account's constructor.

cptartur: Okay, we can do `signer | keypair`

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @staticmethod
+    def from_private_key(key: int) -> "KeyPair":
+        return KeyPair(private_key=key, public_key=private_to_stark_key(key))
+
+
+class StarkCurveSigner(BaseSigner):
+    def __init__(
+        self,
+        address: AddressRepresentation,
```
Arcticae, line 35: Why is that part of the class and not the `sign_transaction` parameter?

starknet_py/net/account/account_client.py
```diff
@@ -203,25 +183,56 @@ async def create_account(
...
+        compiled_contract=COMPILED_ACCOUNT_CONTRACT,
+    )
+    await client.wait_for_tx(
+        tx_hash=result["transaction_hash"],
+    )
+    address = result["address"]
+    return address
+
+
+def add_signature_to_transaction(
```
Arcticae, line 228: cannot this be done by `.replace` method? (since InvokeFunction is a dataclass)

starknet_py/net/account/account_client.py
```diff
@@ -203,25 +183,56 @@ async def create_account(
...
+    client = Client(net=net, chain=chain)
+    result = await client.deploy(
+        constructor_calldata=[public_key],
+        compiled_contract=COMPILED_ACCOUNT_CONTRACT,
+    )
+    await client.wait_for_tx(
+        tx_hash=result["transaction_hash"],
+    )
+    address = result["address"]
+    return address
```
Arcticae, line 225:
```suggestion
    return result["address"]
```

starknet_py/net/account/account_client.py
```diff
@@ -203,25 +183,56 @@ async def create_account(
...
         :param private_key: Private Key used for the account
+        :param signer: Signer used to create account and sign transaction
         :return: Instance of AccountClient which interacts with created account on given network
         """
-        if not private_key:
-            private_key = get_random_private_key()
-
-        key_pair = KeyPair.from_private_key(private_key)
+        if signer is None:
+            if private_key is None:
```
Arcticae, line 190: what if signer is `None` and private key is not `None`?

Arcticae (changes requested):

starknet_py/net/signer/stark_curve_signer.py
cptartur: Signer is created once (or passed as a constructor) to AccountClient, so i don't think it makes much sense to have it as `sign_transaction` parameter.

Arcticae: this address is account address right?

Arcticae: if that's the case, we should name the parameter more clearly

Arcticae (approved):

PR #185, opened 2022-06-03
Remove errors in whole codebase

cptartur:

Apparently there were many errors in our codebase that pylint wasn't detecting (especially minor and some major typerrors, since we aren't using any static typechecker). I was able to find and fix/ignore most of then using Pycharm code analysis tool.

This PR introduces multiple changes, most not related to logic though. Since I've changed things in pretty much every file in starknet.py, I ask for very throughout review so we don't introduce any serious bugs or different errors. 

There are some parts of this PR that require particular attention, I will mark them with comments.

Errors exists 

Most of errors are gone

Hopefully not, but there were some interface changes related to incorrect typing

## **Other information**

n/a


starknet_py/compile/compiler.py
```diff
@@ -27,7 +31,7 @@ class Compiler:
 
     def __init__(
         self,
-        contract_source: List[StarknetCompilationSource],
+        contract_source: StarknetCompilationSource,
```
cptartur, line 34: First `StarknetCompilationSource = NewType(
    "CairoSource", Union[CairoSourceCode, List[CairoFilename]]
)` 
Also current compiler doesn't seem to work correctly with list of `CairoSources`. When i changed `test_create_definition` in `compiler_test.py` to pass `[test_file_content]` instead of just `test_file_content` I was getting errors. We either need to change compiler somehow so it accepts multiple CairoSourceCodes (don't know if it's possible), or change typing like I did here.

starknet_py/compile/compiler.py
```diff
@@ -27,7 +31,7 @@ class Compiler:
...
         cairo_path: Optional[List[str]] = None,
     ):
```
cptartur, line 36: Also since we're accepting cairo path as list of strings, I've added `str()` cast to places where we used `Path`. Perhaps we should use `Path` here instead?

starknet_py/net/l1/messages.py
```diff
@@ -80,20 +84,20 @@ def from_content(cls, msg_content: MessageToEthContent) -> "MessageToEth":
...
+                    starknet_sender=message.from_address,
+                    eth_recipient=int(message.to_address, 16),
+                    payload=[int(felt_str) for felt_str in message.payload],
                 )
             )
-            for message in tx_receipt["l2_to_l1_messages"]
+            for message in tx_receipt.l2_to_l1_messages
         ]
 
     @classmethod
```
cptartur, line 103: Lines 133 and 249 have incorrect typing: both `l1_to_l2_message` and `l2_to_l1_message` return int according to typing that is then cast `int_from_hexbytes` that doesn't make sense. Should we remove casting or change typing here? 
cc @Arcticae

docs/conf.py
```diff
@@ -13,8 +13,6 @@
 #
 import os
 import sys
-import sphinx_rtd_theme
-import enum_tools
```
Solpatium, line 17: Why do you delete it from here? Do you still get a nice description of enums?

starknet_py/compile/compiler.py
Solpatium: I think I noticed it somewhere long time ago, @Arcticae do you confirm a list doesn't work? If so we should just remove it from the type.

starknet_py/compile/compiler.py
Solpatium: You mean we should require Path? It is not ergonomic

starknet_py/net/account/account_client.py
```diff
@@ -162,7 +162,7 @@ async def add_transaction(
         self,
         tx: InvokeFunction,
         token: Optional[str] = None,
-    ) -> Dict[str, int]:
+    ) -> Dict[str, str]:
```
Solpatium, line 165: What can be returned in here?

starknet_py/net/client.py
```diff
@@ -228,7 +234,7 @@ async def wait_for_tx(
     # Mutating methods
     async def add_transaction(
         self, tx: Transaction, token: Optional[str] = None
-    ) -> Dict[str, int]:
+    ) -> Dict[str, str]:
```
Solpatium, line 237: What's returned here?

starknet_py/proxy_check.py
```diff
@@ -4,6 +4,7 @@
 from starknet_py.net.models import parse_address
 
 
+# noinspection PyUnresolvedReferences
```
Solpatium, line 7: Why is this needed here?

Solpatium (changes requested):

starknet_py/compile/compiler.py
cptartur: That's what i meant. I'm fine with still accepting string and adding necessary casts though (also it works without explicit casting so it doesn't matter for end users)

docs/conf.py
cptartur: They were showing as unused import. I'm not sure if it changes anything in the docs, perhaps you can compare [docs built for this PR](https://starknetpy--185.org.readthedocs.build/en/185/) with [current](https://starknetpy.readthedocs.io/en/latest/index.html)

starknet_py/proxy_check.py
cptartur: Otherwise `"Client"` was showing as unresolved reference for some reason, and we cannot import it here due to circular reference

starknet_py/net/client.py
cptartur: There's no schema exposed in cairo-lang library, gateway's add_transaction that is used in this method just returns `Dict[str, str]`. We could build our own schema but it's part of the ongoing refactor anyway.

starknet_py/net/account/account_client.py
cptartur: Same as below

starknet_py/compile/compiler.py
Solpatium: If Path is required by cairo-lang then can't we just accept strings and map them to Path, removing casts?

docs/conf.py
Solpatium: Interesting, looks like enums are still rendered properly.

starknet_py/net/client.py
Solpatium: But do they actually return strings in current version (not 0.9)?

starknet_py/proxy_check.py
Solpatium: I love circular references in python <3

starknet_py/net/client.py
Solpatium: I am sure there was at least one endpoint with invalid typing.

starknet_py/compile/compiler.py
cptartur: It's the other way around, we're internally passing Paths multiple times, when actually there not used anywhere in cairo-lang or our compiler module

starknet_py/net/client.py
cptartur: I took typing straight from cairo-lang 0.8.2.1. It seems like they introduced this change in 0.8.1
[see 0.8.1](https://github.com/starkware-libs/cairo-lang/blob/v0.8.1/src/starkware/starknet/services/api/gateway/gateway_client.py)
[see 0.8.0](https://github.com/starkware-libs/cairo-lang/blob/v0.8.0/src/starkware/starknet/services/api/gateway/gateway_client.py)

starknet_py/compile/compiler.py
Arcticae: Hmm i think the intent was here to provide the interface which can do 2 things:

- Provide a list of input files 
- Provide a snippet of compilable cairo, for demonstration purposes and quick compiling. 

I don't think list of loaded source codes is viable in the implementation - i'll take a look

Arcticae: Yep that's right
```

def load_source_code(
    src: List[StarknetCompilationSource],
) -> List[Tuple[str, str]]:
    if isinstance(src, str):
        return [(src, str(hash(src)))]
    return [(load_cairo_source_code(filename), filename) for filename in src]
```

cptartur: So it's correct that I changed it to just `StarknetCompilationSource` without the `List`?

starknet_py/compile/compiler.py
Solpatium: Hmmm do we lose anything if we switched to strings?

starknet_py/net/client.py
Solpatium: Ok, just please see if they return strings in response. You can resolve this thread id they do :)

starknet_py/compile/compiler_test.py
```diff
@@ -51,7 +51,7 @@ def test_throws_on_compile_with_wrong_extension():
 def test_compile_with_search_path():
     output_file_str = Compiler(
         contract_source=[base_contract_path.resolve().absolute()],
-        cairo_path=[mock_contracts_base_path],
+        cairo_path=[str(mock_contracts_base_path)],
```
cptartur, line 54: Here

starknet_py/compile/compiler.py
cptartur: Probably the convenience that comes with using `Path` instead of raw strings. [See here for example of casts I added.](https://github.com/software-mansion/starknet.py/pull/185/files#r889915693)

starknet_py/net/client.py
cptartur: I've verified on testnet:

If transaction is successful it will be dict with `code` and `transaction_hash`, both strings.

If transaction fails it returns dict with string `code` and `message` that is also a string (but contains serialized dict, e.g.: `"{'calldata': {0: ['Expected int string, got: \"xyz\".']}}"`).

But I'd say typing is correct since we're not doing any parsing of the output.

starknet_py/net/l1/messages.py
Arcticae: we should probably change typing. keeping the interface intact seems the way to go

starknet_py/net/l1/contracts.py
```diff
@@ -64,8 +67,8 @@ def l2_to_l1_messages(
         )
 
     def l1_to_l2_messages(
-        self, msg_hash: int, block_number: EthBlockIdentifier = None
-    ) -> int:
+        self, msg_hash: bytes, block_number: EthBlockIdentifier = None
+    ) -> HexBytes:
```
Arcticae, line 71: why not int here? we would just like to return the count of messages

cptartur: Updated this to bytes, count of messages is returned in messages.py, where this bytes response is cast to int.

Arcticae (approved):

starknet_py/compile/compiler.py
Solpatium: Where do we use utilities from Path?

Solpatium: ok 👍

Solpatium (approved):

PR #188, opened 2022-06-08
Add is_account_contract parameter to compiler

cptartur:

Adds is_account_contract parameter to compiler, fixes compiler for cairo-lang 0.9.0

Compiler is not working at all

as above

While new parameter is optional, compiler is now defaulting to `is_account_contract=False`. Trying to compile account contracts with explicitly setting this to `True` will fail, breaking code from previous versions.

## **Other information**

n/a


starknet_py/compile/compiler.py
```diff
@@ -28,24 +28,32 @@ class Compiler:
...
         """
         self.contract_source = contract_source
+        self.is_account_contract = is_account_contract
         self.search_paths = cairo_path
 
     def compile_contract(self) -> str:
         """
         Compiles a contract and returns it as string
 
+        :raises PreprocessorError: when is_account_contract parameter is incorrectly set
```
Arcticae, line 49: Shouldn't raise, since we're providing a default? Or will it throw when it's set to true, and contract has a \_\_default\_\_ method?

starknet_py/compile/compiler.py
```diff
@@ -28,24 +28,32 @@ class Compiler:
...
         :return: string of compiled contract
         """
-        return starknet_compile(self.contract_source, search_paths=self.search_paths)
+        return starknet_compile(
+            self.contract_source,
```
Arcticae, line 53: let's provide all args as keyword args

starknet_py/compile/compiler.py
cptartur: It will throw in 2 cases:
* when trying to compile contract that has `__execute__` method and `is_account_contract=False`
* in the opposite case

See [this line](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/compiler/compile.py#L159)

PR #203, opened 2022-06-14
Tests working with max_fee > 0

war-in:

Test update

Tests are using pre-deployed devnet accounts 

No


cptartur (commented):
I'd consider moving AccountClient to tests directory so users don't attempt using it thinking it's a part of the library. 

Good work otherwise, especially restoring multicall - we could probably port these changes back to the standard AccountClient once we start work on proper multicall support.

starknet_py/tests/e2e/utils.py
```diff
@@ -1,18 +1,22 @@
...
 DEVNET_PORT = os.environ.get("DEVNET_PORT")
 if not DEVNET_PORT:
     raise RuntimeError("DEVNET_PORT environment variable not provided!")
 
 DEVNET_ADDRESS = f"http://localhost:{DEVNET_PORT}"
 
+ACCOUNT_CLIENT_ADDR = (
+    "0x2109334107efc348a86e72fc3c313061c599e359d46d7b2cd48d22415585b24"
+)
+ACCOUNT_CLIENT_PIV_KEY = "0xcd613e30d8f16adf91b7584a2265b1f5"
```
cptartur, line 18:
```suggestion
ACCOUNT_CLIENT_PRIVATE_KEY = "0xcd613e30d8f16adf91b7584a2265b1f5"
```

starknet_py/tests/e2e/utils.py
```diff
@@ -1,18 +1,22 @@
...
 from starknet_py.net.models.chains import StarknetChainId
 
-
 DEVNET_PORT = os.environ.get("DEVNET_PORT")
 if not DEVNET_PORT:
     raise RuntimeError("DEVNET_PORT environment variable not provided!")
 
 DEVNET_ADDRESS = f"http://localhost:{DEVNET_PORT}"
 
+ACCOUNT_CLIENT_ADDR = (
```
cptartur, line 15:
```suggestion
ACCOUNT_CLIENT_ADDRESS = (
```

cptartur (commented):
I'm not sure if I like this approach of creating a separate AccountClient, even temporary. Two reasons: 

1. We'd have to manually ensure parity of functionalities (except for signing) between AccountClient and AccountClientForTests.
2. By using a entirely different client in utils.py, changing our tests to also run on testnet/integration becomes even more difficult task

I think we may need to consider waiting with merging this PR until we've managed to improve AccountClient's Signer in the way, that allows us to use standard AccountClient on devnet, without rewriting it entirely.

PR #210, opened 2022-06-20
migration guide deleted and guide updated

### LINKED ISSUE #200: Move relevant docs from 0.8 migration guide to relevant categories, deprecate migration guide

war-in:

Resolves #200


docs/migration_guide.rst
```diff
@@ -1,95 +0,0 @@
...
-===========================
-
-Starknet 0.8.0 brought some changes to both its operation and internal APIs.
-Due to that, your existing programs written using starknet.py will require some
-changes to be able to work.
-
-New account contract
---------------------
-
-Starknet.py now uses a `account contract from cairo-lang library <https://github.com/starkware-libs/cairo-lang/blob/4e233516f52477ad158bc81a86ec2760471c1b65/src/starkware/starknet/third_party/open_zeppelin/Account.cairo>`_.
```
cptartur, line 11: I think this information should be kept somewhere so it is know where our account comes from.

cptartur (changes requested):

docs/migration_guide.rst
war-in: It is kept here: 
https://github.com/software-mansion/starknet.py/blob/13a63175a1e48c9133dc51f738fcb8d2ea19bf52/starknet_py/net/account/account_client.py#L194

So it is in the docs too

cptartur: My bad then, I though it was being removed entirely.

PR #213, opened 2022-06-22
Max_fees are higher than 0

war-in:

Adapt tests for using max_fee > 0


starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```diff
@@ -63,23 +68,23 @@ async def test_using_existing_contracts(run_devnet):
 
     # Using only positional arguments
     invocation = await contract.functions["transferFrom"].invoke(
-        sender, recipient, 10000, max_fee=0
+        sender, recipient, 10000, max_fee=MAX_FEE
```
Arcticae, line 71: MAX_FEE is ambiguous from reader's standpoint. the content should be copy-pasteable as-is

starknet_py/tests/e2e/utils.py
```diff
@@ -27,5 +33,16 @@ async def make_devnet_client(self) -> Client:
         client = await AccountClient.create_account(net=self.net, chain=self.chain)
         return client
 
+    def make_devnet_client_from_predefined_account(self) -> AccountClient:
+        key_pair = KeyPair.from_private_key(int(ACCOUNT_CLIENT_PRIVATE_KEY, 0))
```
Arcticae, line 37: Let's replace `make_devnet_client` with that one (contents only) - and then there's no need for this other method

Arcticae (changes requested):

PR #220, opened 2022-06-30
Extract making txs to a separate modules

Arcticae:

### DIFF
--- a/starknet_py/transactions/declare.py
+++ b/starknet_py/transactions/declare.py
@@ -1,0 +1,36 @@
+from typing import Optional, List
+
+from starkware.starknet.services.api.gateway.transaction import DECLARE_SENDER_ADDRESS
+
+from starknet_py.common import create_compiled_contract
+from starknet_py.compile.compiler import StarknetCompilationSource
+from starknet_py.net.models.transaction import Declare
+
+
+def make_declare_tx(
+        compilation_source: Optional[StarknetCompilationSource] = None,
+        compiled_contract: Optional[str] = None,
+        version: int = 0,
+        cairo_path: Optional[List[str]] = None,
+) -> Declare:
+    """
+        Create declaration tx.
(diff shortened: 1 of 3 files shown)


starknet_py/transactions/deploy.py
```diff
@@ -0,0 +1,25 @@
...
+
+from starknet_py.net.client_models import ContractClass
+from starknet_py.net.models import Deploy
+
+
+def make_deploy_tx(
+        compiled_contract: Union[ContractClass, str],
+        constructor_calldata: List[int],
+        salt: Optional[int] = None,
+        version: int = 0,
```
cptartur, line 13: Perhaps we could also add option to pass `compilation source` as in declare?

starknet_py/transactions/declare.py
```diff
@@ -0,0 +1,36 @@
...
+from starkware.starknet.services.api.gateway.transaction import DECLARE_SENDER_ADDRESS
+
+from starknet_py.common import create_compiled_contract
+from starknet_py.compile.compiler import StarknetCompilationSource
+from starknet_py.net.models.transaction import Declare
+
+
+def make_declare_tx(
+        compilation_source: Optional[StarknetCompilationSource] = None,
+        compiled_contract: Optional[str] = None,
```
cptartur, line 12: I'd change it to `Union[ContractClass, str]`

cptartur (changes requested):

starknet_py/transactions/deploy.py
Arcticae: ah i don't quite like it because it adds too many arguments to the function, but if you really want i'll do it

cptartur: These are helper methods so I think it's fine if they're like this

cptartur: These are helper methods so I think it's fine if they're like this

cptartur (approved):

PR #228, opened 2022-07-04
new "Creating AccountClient" section in quickstart

war-in:

It adds a new section to quickstart.

# **What is the current behavior?** (You can also link to an open issue here)

There is no example of creating AccountClient if an account is already deployed.

There is the new section in quickstart about creating account client

No

### DIFF
--- a/starknet_py/tests/e2e/docs/quickstart/test_creating_account_client.py
+++ b/starknet_py/tests/e2e/docs/quickstart/test_creating_account_client.py
@@ -1,0 +1,50 @@
+import pytest
+
+from starknet_py.net.signer.stark_curve_signer import StarkCurveSigner
+
+
+@pytest.mark.asyncio
+async def test_creating_account_client(run_devnet):
+    # pylint: disable=import-outside-toplevel, unused-variable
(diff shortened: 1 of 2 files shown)


starknet_py/tests/e2e/docs/quickstart/test_creating_account_client.py
```diff
@@ -0,0 +1,50 @@
...
+    account_client = AccountClient(
+        address="0x1234", signer=signer, net=TESTNET, chain=StarknetChainId.TESTNET
+    )
+    # add to docs: end
+    testnet = run_devnet
+    # add to docs: start
+
+    # Deploys an account on local network and returns an instance
+    account_client = await AccountClient.create_account(
+        net=testnet, chain=StarknetChainId.TESTNET
```
Arcticae, line 48: same as above, chain is not required when you say net is TESTNET. And also - user doesn't know what `testnet` is here, there is no variable of that name in the docs from user's standpoint. Let's define it, and then switch it behind the scenes

Arcticae (approved):

PR #230, opened 2022-07-05
Old client removal

Arcticae:


starknet_py/net/client.py
```diff
@@ -1,299 +1,227 @@
...
         """
-        Declares contract class.
 
-        :param tx: Declare transaction object
-        :return: Dictionary with 'transaction_hash' and 'class_hash'
+    @abstractmethod
+    async def declare(self, transaction: Declare) -> SentTransaction:
         """
-        return await self.add_transaction(tx)
+        Send a declare trasnaction
```
war-in, line 206: trasnaction -> transaction

starknet_py/transactions/declare.py
```diff
@@ -24,10 +24,9 @@ def make_declare_tx(
     :param cairo_path: a ``list`` of paths used by starknet_compile to resolve dependencies within contracts
     :return: A "Declare" transaction object
     """
-    if isinstance(compiled_contract, str):
-        compiled_contract = create_compiled_contract(
-            compilation_source, compiled_contract, cairo_path
-        )
+    compiled_contract = create_compiled_contract(
+        compilation_source, compiled_contract, cairo_path
+    )
```
war-in, line 29: What would happen if compiled_contract is ContractClass?

Arcticae: it cannot be `ContractClass` - it's an optional string right now. Is something wrong with that?

war-in: image

Are you sure? Because a few lines above it is Union[str, ContractClass].
Or am I missing something?

war-in (approved):

PR #232, opened 2022-07-06
CairoSerializer and FunctionCallSerializer

### LINKED ISSUE #231: Add event support to DataTransformer

cecco_ asked on discord:
"How do I decode event's data from a list of felts? provided that I have the abi of course."

I've checked it and we don't support transforming events now.

DataTransformer shouldn't assume that it transforms a function. It should allow transforming anything from and to python

war-in:

It resolves #231 

There is CairoSerializer and FunctionCallSerializer.
CairoSerializer can serialize any cairo structure 
FunctionCallSerializer replaces current DataTransformer

No


docs/guide.rst
```diff
@@ -178,6 +178,18 @@ Starknet.py transforms python values to Cairo values and the other way around.
      - int
 
 
+Using CairoSerializer
+---------------------
+
+CairoSerializer can be used to transform any data (like a function call or an event). It requires an abi of the contract, types of values and data to be serialized.
```
cptartur, line 184:
```suggestion
CairoSerializer can be used to transform any data (like a function call or an event) between cairo and python format. It requires an abi of the contract, types of values and data to be serialized.
```

What do you think?

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -361,5 +371,26 @@ def _remove_array_lengths(type_by_name: dict) -> dict:
...
+    def from_python(self, *args, **kwargs) -> (List[int], Dict[str, List[int]]):
+        return self.structure_transformer.from_python(
+            self.abi["inputs"], *args, **kwargs
+        )
+
+    def to_python(self, values: CairoData) -> NamedTuple:
+        return self.structure_transformer.to_python(self.abi["outputs"], values)
+
+
+DataTransformer = FunctionCallSerializer
```
cptartur, line 396: I don't like it, I think we should just rename every usage of DataTransformer.

Or we could do some kind of deprecation [like this ](https://stackoverflow.com/questions/9008444/how-to-warn-about-class-name-deprecation), and remove it in the future.

starknet_py/tests/e2e/docs/guide/test_using_cairo_serializer.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    transaction_hash = invoke_result.hash
+    transaction_receipt = await client.get_transaction_receipt(transaction_hash)
+
+    # Takes events from transaction receipt
+    events = transaction_receipt.events
+
+    # Takes an abi of the event which data we want to serialize
+    emitted_event_abi = contract.data.abi[0]
```
cptartur, line 75: Could we have a more explicit example? I think this way it's a little bit unclear what are you actually assigning to emitted_event_abi, but maybe it's just me.

starknet_py/utils/data_transformer/data_transformer.py
war-in: @Solpatium what do you think?

starknet_py/tests/e2e/docs/guide/test_using_cairo_serializer.py
war-in: You're right it's unclear. I can explicitly pass the event's abi here. It would be something like this:

emitted_event_abi = {'data': [{'name': 'key', 'type': 'felt'}, {'name': 'prev_value', 'type': 'felt'}, {'name': 'value', 'type': 'felt'}], 'keys': [], 'name': 'put_called', 'type': 'event'}

Is that ok?

cptartur: I think its better in this case.

starknet_py/utils/data_transformer/data_transformer.py
Solpatium: Yes, please add a deprecation warning.

PR #249, opened 2022-07-18
Fixtures for creating clients and tests easily runnable on testnet

war-in:

DevnetClientFactory replaced by fixtures

Tests are runnable on devnet/testnet/integration (--net parameter)

Just specify --net parameter in pytest to run tests on different nets

No


starknet_py/tests/e2e/conftest.py
```diff
@@ -84,3 +112,55 @@ def fixture_run_prepared_devnet(run_devnet) -> Tuple[str, dict]:
...
+def create_rpc_client(run_devnet):
+    return FullNodeClient(
+        node_url=run_devnet + "/rpc", chain=StarknetChainId.TESTNET, net=run_devnet
+    )
+
+
+@pytest.fixture(name="account_client", scope="function")
+# pylint: disable=redefined-outer-name
+def create_account_client(pytestconfig, gateway_client):
+    net = pytestconfig.getoption("--net")
```
cptartur, line 139: I'd consider returning inside ifs instead of overwriting `client` and returning it at the end of the function

cptartur (changes requested):
We should move `run_prepared_devnet` and `clients` fixtures to the `clients/conftest.py`. It makes more sense to have them there.

starknet_py/tests/e2e/conftest.py
```diff
@@ -137,30 +114,26 @@ def create_rpc_client(run_devnet):
...
-            key_pair=key_pair,
-        )
-
-    return client
+    key_pair = KeyPair.from_private_key(int(DEVNET_ACCOUNT_PRIVATE_KEY, 0))
+    return AccountClient(
+        address=DEVNET_ACCOUNT_ADDRESS,
+        client=gateway_client,
+        key_pair=key_pair,
+    )
```
cptartur, line 139: Now I've realized: could we replace this lines with helper method that takes `private_key` and `address` and returns an account client? This would reduce duplication in if branches.

cptartur (commented):
Just a one suggestion.

cptartur (approved):

PR #250, opened 2022-07-19
Cairo files refactor and tests speed up

war-in:

It moves .cairo files to one directory and adds fixtures to deploy the most used contracts only once per package.

.cairo files are in different folders and are duplicated. Contracts like map.cairo are deployed in many tests (deploys are slow)

No


starknet_py/tests/e2e/client/conftest.py
```diff
@@ -11,14 +11,15 @@
...
+from starknet_py.tests.e2e.conftest import directory_with_contracts
 
 directory = os.path.dirname(__file__)
 
 
 def prepare_devnet(net: str) -> dict:
-    script_path = Path(directory, "prepare_devnet_for_gateway_test.sh")
-    contract_compiled = Path(directory, "balance_compiled.json")
-    contract_abi = Path(directory, "balance_abi.json")
+    script_path = Path(directory + "/prepare_devnet_for_gateway_test.sh")
```
Arcticae, line 20: these days i tend to prefer this style of merging paths:
`    script_path = Path(directory) / "prepare_devnet_for_gateway_test.sh"`

Arcticae (changes requested):

starknet_py/tests/e2e/conftest.py
```diff
@@ -90,7 +93,7 @@ def pytest_collection_modifyitems(config, items):
             item.add_marker(pytest.mark.skip())
 
 
-@pytest.fixture(name="gateway_client", scope="function")
+@pytest.fixture(name="gateway_client", scope="module")
```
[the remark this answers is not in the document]
war-in: Firstly, because account_client fixture uses it and its scope is a module. Secondly, it is ok for us because we don't need a new one everywhere.

starknet_py/tests/e2e/conftest.py
```diff
@@ -134,3 +137,30 @@ def account_client(pytestconfig, gateway_client):
...
     return create_account_client(address, private_key, gateway_client)
+
+
+directory_with_contracts = Path(os.path.dirname(__file__)) / "mock_contracts_dir"
+
+
+@pytest.fixture(name="map_contract", scope="module")
+def deploy_map_contract(account_client) -> Contract:
+    # pylint: disable=no-member
+    map_source_code = (directory_with_contracts / "map.cairo").read_text("utf-8")
```
Arcticae, line 148: we could define those at module-level so it would be reusable

Arcticae (changes requested):

Arcticae (approved):

PR #254, opened 2022-07-21
Token parameter in gateway client

### LINKED ISSUE #253: Add the token parameter to the gateway API methods

war-in:

(closes #253 )


starknet_py/net/gateway_client.py
```diff
@@ -285,12 +295,21 @@ async def get_class_by_hash(self, class_hash: Hash) -> DeclaredContract:
...
     async def _add_transaction(
-        self, tx: StarknetTransaction
+        self,
+        tx: StarknetTransaction,
+        token: Optional[str] = None,
     ) -> SentTransactionResponse:
-        res = await self._gateway_client.post(
-            method_name="add_transaction",
-            payload=StarknetTransaction.Schema().dump(obj=tx),
-        )
```
cptartur, line 293: How about `params = {"token": token} if token is not None else {}` then you don't have to duplicate client.post(...) code.

cptartur (changes requested):

cptartur (approved):
Looks good to me. However I'd like @Arcticae to also review this one before we merge.

Arcticae:
This seems fine codewise but i'd rather have confirmation from SW that we really need this, and this is going to get used some day

starknet_py/net/gateway_client.py
```diff
@@ -259,15 +259,25 @@ async def call_contract(
         return [int(v, 16) for v in res["result"]]
 
     async def send_transaction(
-        self, transaction: InvokeFunction
+        self,
+        transaction: InvokeFunction,
+        token: Optional[str] = None,
     ) -> SentTransactionResponse:
-        return await self._add_transaction(transaction)
+        return await self._add_transaction(transaction, token)
```
ca11ab1e, line 266:
```suggestion
        return await self._add_transaction(transaction, token=token)
```

starknet_py/net/gateway_client.py
```diff
@@ -259,15 +259,25 @@ async def call_contract(
...
 
-    async def deploy(self, transaction: Deploy) -> SentTransactionResponse:
-        return await self._add_transaction(transaction)
+    async def deploy(
+        self,
+        transaction: Deploy,
+        token: Optional[str] = None,
+    ) -> SentTransactionResponse:
+        return await self._add_transaction(transaction, token)
```
ca11ab1e, line 273:
```suggestion
        return await self._add_transaction(transaction, token=token)
```

cptartur:
> This seems fine codewise but i'd rather have confirmation from SW that we really need this, and this is going to get used some day

I think we should focus on releasing 0.9.1 and deal with this PR once SW replies then.

starknet_py/net/gateway_client.py
```diff
@@ -259,15 +259,25 @@ async def call_contract(
...
 
-    async def declare(self, transaction: Declare) -> SentTransactionResponse:
-        return await self._add_transaction(transaction)
+    async def declare(
+        self,
+        transaction: Declare,
+        token: Optional[str] = None,
+    ) -> SentTransactionResponse:
+        return await self._add_transaction(transaction, token)
```
ca11ab1e, line 280:
```suggestion
        return await self._add_transaction(transaction, token=token)
```

Arcticae:
> > This seems fine codewise but i'd rather have confirmation from SW that we really need this, and this is going to get used some day
> 
> I think then we should focus on releasing 0.9.1 and deal with this PR once SW replies.

That is fine by me

ca11ab1e:
> > This seems fine codewise but i'd rather have confirmation from SW that we really need this, and this is going to get used some day
> 
> I think then we should focus on releasing 0.9.1 and deal with this PR once SW replies.

I can reply here: we will use it :)

ca11ab1e:
And thanks for being fast to respond!

ca11ab1e:
> > This seems fine codewise but i'd rather have confirmation from SW that we really need this, and this is going to get used some day
> 
> I think then we should focus on releasing 0.9.1 and deal with this PR once SW replies.

@cptartur please include tha tsmall patch in the same release as you will ship Cairo 0.9.1 :pray: 
Not only it would block SW to move forward, but also the whole Ape StarkNet tooling.

PR #262, opened 2022-07-22
Data transformer namedtuple wrapper

Arcticae:

Replaces namedtuple with a wrapper class, fixes the bug with underscore-prefixed properties

Namedtuple doesn't allow underscore-prefixed properties

Return a new "Return" class which leverages closures to implement a thin proxy for the namedtuple object

It should not, unless someone uses `isinstance(ret_value, NamedTuple)`


starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -30,6 +31,38 @@
...
+        def _asdict():
+            return value
+
+        def __getitem__(self, item):
+            return tpl_value[item]
+
+        def __iter__(self):
+            return tpl_value.__iter__()
+
+    return Result()
```
war-in, line 63: The returned type is not correct. Maybe there is a way to fix this?

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -30,6 +31,38 @@
...
+    fields = result.keys()
+    res_tuple = namedtuple(
+        field_names=fields,
+        typename="Result",
+        rename=True,
+    )
+    # pylint: disable=protected-access
+    mappings = dict(zip(fields, res_tuple._fields))
+    value = {mappings[key]: value for key, value in result.items()}
+    tpl_value = res_tuple(**value)
```
war-in, line 44: Maybe some changes in naming? I'm not sure if it's important here, but it would be great to have more descriptive ones.

war-in (approved):

PR #269, opened 2022-07-27
fix: missing transaction rejection reason

ca11ab1e:

Fix a 0.4 regression about missing rejection reason from failed transactions.

I also reworked several custom exceptions, to homogenize with others, and to ease using exception attributes outside the project.

No way to know what was the root cause of a failed transaction.

Now, the rejection reason is filled properly.

No. I changed internal stuff only.

## **Other information**

It's a critical bug fix for Ape StarkNet, and SithSwap. Hopefully it will gain some attention to move forward quickly :)

### DIFF
--- a/starknet_py/net/client.py
+++ b/starknet_py/net/client.py
@@ -161,7 +161,8 @@ class Client(ABC):
                         if result.block_number is not None:
                             return result.block_number, status
                 elif status == TransactionStatus.REJECTED:
-                    raise TransactionRejectedError(result.rejection_reason)
+                    reason = result.rejection_reason or {}
+                    raise TransactionRejectedError(**reason)
                 elif status == TransactionStatus.UNKNOWN:
                     if not first_run:
                         raise TransactionNotReceivedError()
--- a/starknet_py/net/client_errors.py
+++ b/starknet_py/net/client_errors.py
@@ -8,7 +8,7 @@ class ClientError(Exception):
 
     def __init__(self, message: str, code: Optional[str] = None):
         self.code = code
-        self.message = message.strip()
+        self.message = message
         super().__init__(self.message)
 
     def __str__(self):
--- a/starknet_py/net/client_errors.py
+++ b/starknet_py/net/client_errors.py
@@ -8,8 +8,14 @@ class ClientError(Exception):
 
     def __init__(self, message: str, code: Optional[str] = None):
         self.code = code
-        self.message = f"Client failed{f' with code {code}' if code is not None else ''}: {message}"
+        self.message = message.strip()
         super().__init__(self.message)
+
+    def __str__(self):
+        return (
+            f"Client failed{f' with code {self.code}' if self.code is not None else ''}"
+            f": {self.message}"
+        )
 
 
 class ContractNotFoundError(ClientError):
@@ -27,7 +33,10 @@ class ContractNotFoundError(ClientError):
         identifier = block_hash or block_number
         self.identifier = str(identifier) if isinstance(identifier, int) else identifier
 
-        super().__init__(message=f"No contract found for identifier: {self.identifier}")
+        super().__init__(f"No contract found for identifier: {self.identifier}")
+
+    def __str__(self):
+        return f"No contract found for identifier: {self.identifier}"
 
 
 def require_block_identifier(block_hash, block_number):
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -153,7 +153,7 @@ class TransactionReceipt:
     block_number: Optional[int] = None
     version: int = 0
     actual_fee: int = 0
-    rejection_reason: Optional[str] = None
+    rejection_reason: Optional[Dict[str, Any]] = None
 
     events: List[Event] = None
     l2_to_l1_messages: List[L2toL1Message] = None
--- a/starknet_py/net/gateway_schemas/gateway_schemas.py
+++ b/starknet_py/net/gateway_schemas/gateway_schemas.py
@@ -114,8 +114,12 @@ class TransactionReceiptSchema(Schema):
     block_number = fields.Integer(data_key="block_number", load_default=None)
     version = fields.Integer(data_key="version", allow_none=True)
     actual_fee = Felt(data_key="actual_key", allow_none=True)
-    rejection_reason = fields.String(
-        data_key="transaction_rejection_reason", allow_none=True, load_default=None
+    rejection_reason = fields.Dict(
+        keys=fields.String(),
+        values=fields.Raw(),
+        data_key="transaction_failure_reason",
+        allow_none=True,
+        load_default=None,
     )
     events = fields.List(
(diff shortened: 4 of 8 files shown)


starknet_py/net/client_errors.py
```diff
@@ -8,9 +8,15 @@ class ClientError(Exception):
 
     def __init__(self, message: str, code: Optional[str] = None):
         self.code = code
-        self.message = f"Client failed{f' with code {code}' if code is not None else ''}: {message}"
+        self.message = message.strip()
```
unparalleled-js, line 11: Is the strip() necessary? What if for some reason the user is expecting trailing whitespace

ca11ab1e: It seems to be always JSON encoded with a trailing `\n` when it comes from the server.
What would be the use case to add a trailing whitespace?

unparalleled-js: Maybe something like this, once strings are more of a thing in Cairo:

```cairo
"Expected number but receiver value: "
```

in the case when a user might supply whitespace as an argument and the whitespace gets interpolated in the error message

ca11ab1e: In that case, the full error would be something like:
```json
{
    "code": 42,
    "message": "Expected number but received value: "
}
```
And so the `strip() applies only to the stringified response:

```json
{"code":42,"message":"Expected number but received value: "}\n
```

I may miss some specific usages though. So if it's problematic, let's remove that `strip()` :)

ca11ab1e: Oh wait, you're right. Sorry :sweat: 
I'll remove it.

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -158,13 +162,16 @@ async def test_get_storage_at_incorrect_address(clients):
...
-    with pytest.raises(ClientError):
+    with pytest.raises(ClientError) as exc:
         await full_node_client.get_storage_at(
             contract_address=0x1111,
             key=916907772491729262376534102982219947830828984996257231353398618781993312401,
             block_hash="latest",
         )
 
+    assert exc.value.code == 20
+    assert exc.value.message == "Contract not found"
```
ca11ab1e, line 173:
```suggestion
    with pytest.raises(ClientError) as err:
        await full_node_client.get_storage_at(
            contract_address=0x1111,
            key=916907772491729262376534102982219947830828984996257231353398618781993312401,
            block_hash="latest",
        )

    assert err.value.code == 20
    assert err.value.message == "Contract not found"
```

starknet_py/net/client.py
```diff
@@ -161,7 +161,8 @@ async def wait_for_tx(
                         if result.block_number is not None:
                             return result.block_number, status
                 elif status == TransactionStatus.REJECTED:
-                    raise TransactionRejectedError(result.rejection_reason)
+                    reason = result.rejection_reason or {}
+                    raise TransactionRejectedError(**reason)
```
cptartur, line 165: Same with changes to the `TransactionReceipt`, we'd prefer to keep the `TransactionRejectedError` oriented to RPC which doesn't include code. We could add a subclass of this exception specifically for the gateway_client, that incudes the code along with the message, so it doesn't confuse the RPC users.

starknet_py/net/gateway_schemas/gateway_schemas.py
```diff
@@ -114,8 +114,12 @@ class TransactionReceiptSchema(Schema):
     block_number = fields.Integer(data_key="block_number", load_default=None)
     version = fields.Integer(data_key="version", allow_none=True)
     actual_fee = Felt(data_key="actual_key", allow_none=True)
-    rejection_reason = fields.String(
-        data_key="transaction_rejection_reason", allow_none=True, load_default=None
```
cptartur, line 118: Just acknowledging that the data_key was incorrect in this case and since our tests didn't check error messages we didn't catch that it's not working correctly.

cptartur (commented):
Thanks for finding the issue with `rejection_reason` in TransactionReceipt. 👍 

This issue is however a little bit more complicated - we're trying to orient starknet.py around the RPC api since it will become a default way of interaction with starknet at some point. That's why simply changing the typing of the `TransactionReceipt` is not currently possible, see my comments where I outline a possible solution for this issue, let me know what you think.

I believe these changes could be done on our side if you're fine with us keeping the tests improvements added in this PR and modifying the rest.

ca11ab1e:
> I believe these changes could be done on our side if you're fine with us keeping the tests improvements added in this PR and modifying the rest.

Of course, the sooner it's fixed the better.

Am I getting it right if I say you will work on a proper patch on your side?

starknet_py/net/gateway_schemas/gateway_schemas.py
ca11ab1e: IIR I encountered a validation error because it's effectively a `dict`.

starknet_py/net/client_errors.py
```diff
@@ -8,9 +8,15 @@ class ClientError(Exception):
...
-        self.message = f"Client failed{f' with code {code}' if code is not None else ''}: {message}"
+        self.message = message
         super().__init__(self.message)
 
+    def __str__(self):
+        return (
+            f"Client failed{f' with code {self.code}' if self.code is not None else ''}"
+            f": {self.message}"
+        )
+
```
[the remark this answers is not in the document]
ca11ab1e: The interesting change I would like to keep here is the fact that `.message` only includes the `message` argument, and not a modified version appending "Client failed...". It would ease using the `message` attribute outside the project (else I would need to "parse" the message, and try to get back the original value, which is not really future-proof nor neat).

starknet_py/net/client_models.py
```diff
@@ -153,7 +153,7 @@ class TransactionReceipt:
     block_number: Optional[int] = None
     version: int = 0
     actual_fee: int = 0
-    rejection_reason: Optional[str] = None
+    rejection_reason: Optional[Dict[str, Any]] = None
```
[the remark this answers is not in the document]
ca11ab1e: IIR I encountered a validation error because it's effectively a `dict`.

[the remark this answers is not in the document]
cptartur: It is a dict in the gateway response but just a string in rpc. That's why I'd like to have rejection_reason as a string in `TransactionReceipt` (base class) since it's present in both, and then also have a `code` in `GatewayTransactionReceipt`. From my quick tests it seems like this dict only contains these two keys.

cptartur:
> > I believe these changes could be done on our side if you're fine with us keeping the tests improvements added in this PR and modifying the rest.
> 
> Of course, the sooner it's fixed the better.
> 
> Am I getting it right if I say you will work on a proper patch on your side? If so, no worries, just go ahead :)

Yes, we'd like to work on proper patch on our side. We'll however try to include the tests you added in this PR since they're very useful, so please keep the PR open for the time being.

war-in:
Closing this, because #271 was merged. Thanks for contributing!

PR #276, opened 2022-07-29
fix: get_code() and ContractNotFound usage

### LINKED ISSUE #263: [BUG] block hash or number is always required for get_code call
Before, I did not have to specify a block hash or block number when calling `get_code()`
However, as of the latest release, if i fail to do this, I get the error:

```
No contract found for identifier: No contract found with following identifier {}
```

It is complaining about an empty dictionary, however when looking at the custom exception, it seems to expect a hash or number (might be slightly unrelated but still worth noting that the exception logic might need to handle Block identifier dictionaries now).

## Python version

3.9

## Expected Behavior

* Uses latest number by default without need to specify
* Exception adjustments - perhaps telling me it is missing rather than {} is not a valid block ID

## Current Behavior

Always errors

## Possible Solution

Grab latest ID if nothing passed in as argument

## Steps to Reproduce

1. connect client to devnet
2. deploy a contract
3. try to get the code and abi of the contract using RPC (don't provider a block ID arg)
4. notice the error
5.

ca11ab1e:

- Fix `ContractNotFound` usage.
- Fix #263.

### DIFF
--- a/starknet_py/net/gateway_client.py
+++ b/starknet_py/net/gateway_client.py
@@ -232,7 +232,9 @@ class GatewayClient(Client):
         )
 
         if len(res["bytecode"]) == 0:
-            raise ContractNotFoundError(block_hash=block_hash, block_number=block_number)
+            raise ContractNotFoundError(
+                block_hash=block_hash, block_number=block_number
+            )
 
         return ContractCodeSchema().load(res, unknown=EXCLUDE)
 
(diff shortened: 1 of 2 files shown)


war-in:
Thanks for fixing it! We noticed that the default block_number parameter would cause an error while providing block_hash, because these two parameters are mutually exclusive. I'll merge your PR and develop a more complex solution for it. Thanks!

PR #277, opened 2022-08-01
Block number is "pending" for default

war-in:

It sets block_number parameter as "pending" everywhere

**What is the current behaviour?** (You can also link to an open issue here)

block_number is set to "pending" as default only in two methods (get_code and estimate_fee).
If block_hash is provided to these methods there is an error, because block_hash and block_number are passed (first as an argument and second as default)

block_number is default everywhere

No

### DIFF
--- a/starknet_py/net/gateway_client.py
+++ b/starknet_py/net/gateway_client.py
@@ -163,8 +163,8 @@ class GatewayClient(Client):
         Get the information about the result of executing the requested block
 
         :param block_hash: Block's hash
-        :param block_number: Block's number
-        :return: BlockStateUpdate oject representing changes in the requested block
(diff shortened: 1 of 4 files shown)


starknet_py/net/gateway_client.py
```diff
@@ -233,7 +233,7 @@ async def get_code(
 
         if len(res["bytecode"]) == 0:
             raise ContractNotFoundError(
-                block_hash=block_hash, block_number=block_number
+                block_hash=block_hash, block_number=block_identifier["blockNumber"]
```
Solpatium, line 236: This will throw if block_hash was used.

Solpatium (approved):

PR #279, opened 2022-08-01
Include gas_price in block schema

unparalleled-js:

Includes property from response gas_price in block schema

gas_price is missing from block schema

gas_price is available in the block


war-in:
Thanks for contributing! Tests are failing because RPC doesn't support gas_price in the block (even in the newest specification). The solution could be to create a new dataclass "GatewayBlock" (in the client_models.py) which extends StarknetBlock and adds gas_price field. We don't want to modify the StarknetBlock dataclass, because it's compatible with RPC specification.

PR #280, opened 2022-08-02
Remove chain id from Client interface

war-in:

It removes unnecessary chain parameter from the Client interface

## **What is the current behaviour?** (You can also link to an open issue here)

Chain id is used only in the AccountClient to sign transactions, but it is in the Client interface

Chain id is present only in the AccontClient

Yes, the Client interface has changed

### DIFF
--- a/docs/contract.rst
+++ b/docs/contract.rst
@@ -12,7 +12,7 @@ Contract
 
 .. autoclass:: PreparedFunctionCall
     :exclude-members: __init__, __new__
-    :members: call, call_raw, invoke, call_sync, call_raw_sync, invoke_sync, hash, estimate_fee, estimate_fee_sync
+    :members: call, call_raw, invoke, call_sync, call_raw_sync, invoke_sync, estimate_fee, estimate_fee_sync
 
 .. autoclass:: InvokeResult
     :exclude-members: __init__, __new__
--- a/starknet_py/contract_test.py
+++ b/starknet_py/contract_test.py
@@ -1,7 +1,6 @@
 import pytest
 
-from starknet_py.contract import Contract, PreparedFunctionCall, ContractData
-from starknet_py.net.gateway_client import GatewayClient
+from starknet_py.contract import Contract
 from starknet_py.tests.e2e.conftest import directory_with_contracts
 
 SOURCE = """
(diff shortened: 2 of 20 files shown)


starknet_py/net/account/account_client.py
```diff
@@ -55,34 +55,39 @@ def __init__(
...
+            chain = chain_from_network(net=client.net, chain=chain)
+            signer = StarkCurveSigner(
+                account_address=self.address, key_pair=key_pair, chain_id=chain
+            )
+        self.signer = signer
 
     @property
     def chain(self) -> StarknetChainId:
-        return self.client.chain
+        return self.signer.chain_id
```
cptartur, line 90: I wonder if it makes sense to expose this property here at all if it's going through signer anyway.

starknet_py/net/account/account_client.py
```diff
@@ -55,34 +55,39 @@ def __init__(
...
-        if chain is None and client is None:
-            raise ValueError("One of chain or client must be provided")
         self.address = parse_address(address)
         self.client = client
-        self.signer = signer or StarkCurveSigner(
-            account_address=self.address, key_pair=key_pair, chain_id=self.client.chain
-        )
+
+        if signer is None:
+            chain = chain_from_network(net=client.net, chain=chain)
```
cptartur, line 82: What if someone has provided a chain? This way we're overwriting it. Shouldn't it be something along the lines of `chain = chain or ...`

starknet_py/net/account/account_client.py
```diff
@@ -370,14 +376,15 @@ async def create_account(
         :param private_key: Private Key used for the account
         :param signer: Signer used to create account and sign transaction
         :return: Instance of AccountClient which interacts with created account on given network
+        :param chain: ChainId of the chain used by the client
```
cptartur, line 379: We should be doing some verification here if either signer or chain was provided, same as in constructor.

docs/contract.rst
```diff
@@ -12,7 +12,7 @@ Contract
 
 .. autoclass:: PreparedFunctionCall
     :exclude-members: __init__, __new__
-    :members: call, call_raw, invoke, call_sync, call_raw_sync, invoke_sync, hash, estimate_fee, estimate_fee_sync
```
cptartur, line 15: This removal needs to be mentioned in the release notes of the next version.

cptartur (changes requested):

starknet_py/net/account/account_client.py
war-in: But the chain is overridden only if net=("mainnet" or "testnet"). Isn't it ok? What should we do if someone provides the wrong chain while using these two networks?

cptartur (approved):

starknet_py/contract_test.py
```diff
@@ -119,24 +118,6 @@ def test_compute_address_throws_on_no_source():
     )
 
 
-def test_transaction_hash():
```
Arcticae, line 122: So there's no possibility to compute tx hash beforehand right now?

Arcticae (approved):

starknet_py/contract_test.py
war-in: There is an option. Invoke function has the "hash" property so it is possible

war-in:
It will be merged someday, but first, we must deprecate the chain.

PR #284, opened 2022-08-03
Feat/update rpc v0.1.0

war-in:

It updates RPC to the newest version

Yes, in terms of RPC.


starknet_py/net/rpc_schemas/rpc_schemas.py
```diff
@@ -140,8 +139,10 @@ def make_dataclass(self, data, **kwargs) -> DeclareTransaction:
 
 class DeployTransactionSchema(TransactionSchema):
     contract_address = Felt(data_key="contract_address", required=True)
-    constructor_calldata = fields.List(Felt(), data_key="calldata", required=True)
-    # class_hash = Felt(data_key="class_hash", load_default=None)
+    constructor_calldata = fields.List(
+        Felt(), data_key="constructor_calldata", required=True
+    )
+    class_hash = Felt(data_key="class_hash", load_default=None)
```
cptartur, line 145: I wonder if accepting `None` as class_hash is a good idea. This seems like the most important field in the `DeployTransaction`.

cptartur: Perhaps we should get rid of `load_default` here entirely.

starknet_py/net/rpc_schemas/rpc_schemas.py
```diff
@@ -225,26 +228,33 @@ def preprocess(self, data, **kwargs):
...
+        deployed_contracts = data["state_diff"]["deployed_contracts"]
+        deployed_contracts = [
+            DeployedContractSchema().load(contract) for contract in deployed_contracts
+        ]
+
+        storage_diffs = []
+        for diff in data["state_diff"]["storage_diffs"]:
+            storage_diff = StorageDiffSchema().load(
+                {"address": diff["address"], "key": diff["key"], "value": diff["value"]}
+            )
```
cptartur, line 233: Why do we need to construct this dict, couldn't we pass data directly to the schema?

starknet_py/net/rpc_schemas/rpc_schemas.py
```diff
@@ -225,26 +228,33 @@ def preprocess(self, data, **kwargs):
 
     @post_load
     def make_dataclass(self, data, **kwargs) -> BlockStateUpdate:
-        storage_diffs, contract_diffs = BlockStateUpdateSchema._extract_diffs(
-            data["state_diff"]
-        )
+        deployed_contracts = data["state_diff"]["deployed_contracts"]
```
cptartur, line 224: I wonder if it wouldn't make more sense to declare a separate schema for StateDiff and then have `state_diff = fields.Nested(StateDiffSchema(), ...)`. This way we'd only have to flatten this data in `make_dataclass`, without the need for all of this processing that is error prone.

starknet_py/net/rpc_schemas/rpc_schemas.py
```diff
@@ -304,3 +314,15 @@ class DeployTransactionResponseSchema(SentTransactionSchema):
     @post_load
     def make_dataclass(self, data, **kwargs):
         return DeployTransactionResponse(**data)
+
+
+class PendingTransactionsSchema(Schema):
```
cptartur, line 312: Are we testing these? Perhaps if devnet doesn't support pending transactions, we should mock one based off the RPC response and add tests for these.

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -192,4 +192,6 @@ async def test_rejection_reason_in_transaction_receipt(account_clients, map_cont
         res = await map_contract.functions["put"].invoke(key=10, value=20, max_fee=1)
         transaction_receipt = await account_client.get_transaction_receipt(res.hash)
 
+        print(transaction_receipt)
+
```
cptartur, line 196: 👁️

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -259,10 +261,10 @@ async def test_state_update_gateway_client(
         new_root=block_with_deploy_root,
         old_root=genesis_block_root,
         storage_diffs=[],
-        contract_diffs=[
-            ContractDiff(
+        deployed_contracts=[
+            DeployedContract(
                 address=contract_address,
-                contract_hash=0x711941B11A8236B8CCA42B664E19342AC7300ABB1DC44957763CB65877C2708,
+                class_hash=0x711941B11A8236B8CCA42B664E19342AC7300ABB1DC44957763CB65877C2708,
```
cptartur, line 267: Would it be possible to replace this hardcoded value with a fixture?

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -23,51 +25,25 @@ async def test_node_get_transaction_by_block_hash_and_index(
...
-        # class_hash=class_hash,
-    )
-
-
-@pytest.mark.asyncio
-async def test_get_block_throws_on_no_block_hash_and_no_number(rpc_client):
-    with pytest.raises(ValueError) as exinfo:
-        await rpc_client.get_block()
-
-    assert "Either block_hash or block_number is required" in str(exinfo.value)
```
cptartur, line 51: Why was this test removed?

cptartur (commented):
Left some changes.

Unrelated, but I wonder if we should be adding docstrings to some of our fixtures. Our tests are becoming very dependent on these.

starknet_py/tests/e2e/client/full_node_test.py
war-in: Because get_block will not throw. We have default block_id="pending"

starknet_py/contract.py
```diff
@@ -224,10 +225,16 @@ async def invoke(
...
+    async def estimate_fee(
+        self,
+        block_hash: Optional[Union[Hash, Tag]] = None,
+        block_number: Optional[Union[int, Tag]] = None,
+    ):
         """
         Estimate fee for prepared function call
 
+        :param block_hash: Estimate fee at specific block hash
+        :param block_number: Estimate fee at given block number (or "pending" for pending block), default is "pending"
```
cptartur, line 237: Is it though? It doesn't seem like it from the function declaration?

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -92,14 +93,24 @@ async def test_estimate_fee_called(erc20_contract):
...
-        await erc20_contract.functions["balanceOf"]
-        .prepare("1234", max_fee=0)
-        .estimate_fee()
-    )
+async def test_estimated_fee_greater_than_zero(erc20_contract, account_clients):
+
+    for account_client in account_clients:
+        erc20_contract = Contract(
+            erc20_contract.address, erc20_contract.data.abi, account_client
+        )
```
cptartur, line 101: This could use parametrized fixtures with `indirect=True` 😉

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -58,3 +62,72 @@ async def test_node_get_declare_transaction_by_block_number_and_index(
...
+                    "constructor_calldata": [],
+                }
+            ]
+        )
+        mocked_http_call.return_value = result
+
+        pending_transactions = await rpc_client.get_pending_transactions()
+
+        assert len(pending_transactions) == 1
+        assert pending_transactions[0].hash == 1
```
cptartur, line 131:
```suggestion
        assert pending_transactions[0].hash == 0x1
```
Doesn't change anything in logic but i think it's more explicit since in mock you're using felts not ints.

starknet_py/contract.py
war-in: All methods which use block_hash or block_number and FullNode/Gateway Client has the block_number parameter default to "pending"

cptartur: Just to be safe in case we change anything in Client I think we could also include default parameter in `Contract` anyways.

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -286,13 +287,15 @@ async def test_state_update_full_node_client(
...
             )
         ],
-        declared_contracts=[],
+        declared_contracts=[
+            0x711941B11A8236B8CCA42B664E19342AC7300ABB1DC44957763CB65877C2708
+        ],
```
cptartur, line 298: Couldn't these also be replaced with the fixture you defined?

starknet_py/net/full_node_client.py
```diff
@@ -200,25 +171,60 @@ async def estimate_fee(
...
+
+        res = await self._client.call(
+            method_name="estimateFee",
+            params={
+                "request": {
+                    "transaction_hash": convert_to_felt(tx_hash),
+                    "max_fee": convert_to_felt(tx.max_fee),
+                    "version": hex(tx.version),
+                    "signature": [convert_to_felt(i) for i in tx.signature],
+                    "nonce": convert_to_felt(0),  # TODO: do something with nonce
```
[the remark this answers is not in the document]
war-in: this will be used in the next 0.10.0 cairo

Solpatium (approved):
Looks ok for me, but @cptartur has more info about RPC, so please wait for his approval before merging.

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -334,7 +336,9 @@ async def test_deploy(balance_contract, clients):
 @pytest.mark.asyncio
 async def test_get_class_hash_at(clients, contract_address):
     for client in clients:
-        class_hash = await client.get_class_hash_at(contract_address=contract_address)
+        class_hash = await client.get_class_hash_at(
+            contract_address=contract_address, block_hash="latest"
+        )
         assert (
             class_hash
             == 0x711941B11A8236B8CCA42B664E19342AC7300ABB1DC44957763CB65877C2708
```
cptartur, line 344: I'd use a fixture here if possible.

cptartur (commented):
Except for one comment, looks fine to me 👍

PR #285, opened 2022-08-04
Custom gateway_url and feeder_gateway_url for GatewayClient

war-in:

It adds a way to provide custom gateway_url and feeder_gateway_url to GatewayClient constructor

## **What is the current behaviour?** (You can also link to an open issue here)

There is no way to provide custom urls

No

### DIFF
--- a/starknet_py/net/networks.py
+++ b/starknet_py/net/networks.py
@@ -11,7 +11,7 @@ try:
 except ImportError:
     PredefinedNetwork = str
 
-Network = Union[PredefinedNetwork, str]
+Network = Union[PredefinedNetwork, str, dict]
 
 
 def net_address_from_net(net: Network) -> str:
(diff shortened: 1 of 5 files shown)


starknet_py/net/networks.py
```diff
@@ -11,7 +11,7 @@
 except ImportError:
     PredefinedNetwork = str
 
-Network = Union[PredefinedNetwork, str]
+Network = Union[PredefinedNetwork, str, dict]
```
cptartur, line 14: Maybe we could use TypedDict here so it's more explicit what keys we expect this dict to have?

starknet_py/net/networks.py
```diff
@@ -1,4 +1,10 @@
 from typing import Union
+import sys
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
```
Solpatium, line 7: Please add this dependency to our project directly.

cptartur (approved):

PR #286, opened 2022-08-04
Deprecate chain in the Client interface

war-in:

It deprecates chain

Chain id is used only in the AccountClient to sign transactions, but it is in the Client interface

Chain id is deprecated and chain parameter is  added to AccountClient

No

### DIFF
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -55,14 +55,16 @@ class AccountClient(Client):
         client: Client,
         signer: Optional[BaseSigner] = None,
         key_pair: Optional[KeyPair] = None,
+        chain: Optional[StarknetChainId] = None,
     ):
         """
         :param address: Address of the account contract
         :param client: Instance of GatewayClient which will be used to add transactions
         :param signer: Custom signer to be used by AccountClient.
-                       If none is provieded, default
+                       If none is provided, default
                        :py:class:`starknet_py.net.signer.stark_curve_signer.StarkCurveSigner` is used.
         :param key_pair: Key pair that will be used to create a default `Signer`
+        :param chain: ChainId of the chain used by the client
         """
         # pylint: disable=too-many-arguments
         if signer is None and key_pair is None:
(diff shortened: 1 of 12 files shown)


starknet_py/net/full_node_client.py
```diff
@@ -57,10 +58,16 @@ def __init__(
...
+        :param chain: Chain id of the network used by the rpc client. Chain is deprecated
+                        and will be removed in the next releases
         :param session: Aiohttp session to be used for request. If not provided, client will create a session for
-                        every request. When using a custom session, user is resposible for closing it manually.
+                        every request. When using a custom session, user is responsible for closing it manually.
         """
+        if chain is not None:
+            warnings.warn(
+                "Chain is deprecated and will be deleted in the next releases"
+            )
```
cptartur, line 69: Add a `DeprecationWarning` category to the warning.

starknet_py/net/full_node_client.py
```diff
@@ -72,6 +79,7 @@ def net(self) -> Network:
 
     @property
     def chain(self) -> StarknetChainId:
+        warnings.warn("Chain is deprecated and will be deleted in the next releases")
```
cptartur, line 82: Same here and everywhere else.

starknet_py/net/full_node_client.py
cptartur: Also the internet suggests adding `stacklevel=2` to make warning more readable but I haven't actually verified that so you can test yourself if it looks better with or without this parameter.

starknet_py/net/gateway_client.py
```diff
@@ -53,17 +54,23 @@ class GatewayClient(Client):
...
-        :param chain: Chain used by the network. Required if you use a custom URL for ``net`` param.
+        :param chain: Chain used by the network. Required if you use a custom URL for ``net`` param. Chain is deprecated
+                        and will be removed in the next releases
         :param session: Aiohttp session to be used for request. If not provided, client will create a session for
                         every request. When using a custom session, user is resposible for closing it manually.
         """
+        if chain is not None:
+            warnings.warn(
+                "Chain is deprecated and will be deleted in the next releases"
+            )
```
cptartur, line 72: I'm not sure if we should be adding these deprecation warnings to the constructors. If the user isn't using them anyway, they'll just have to remove this parameter once we deprecate chain already.

It's removal is only problem for the users using chain property.

starknet_py/net/account/account_client.py
```diff
@@ -55,39 +55,44 @@ def __init__(
...
     ):
         """
         :param address: Address of the account contract
         :param client: Instance of GatewayClient which will be used to add transactions
         :param signer: Custom signer to be used by AccountClient.
-                       If none is provieded, default
+                       If none is provided, default
                        :py:class:`starknet_py.net.signer.stark_curve_signer.StarkCurveSigner` is used.
         :param key_pair: Key pair that will be used to create a default `Signer`
+        :param chain: ChainId of the chain used by the client
```
cptartur, line 67: Perhaps here we should do a warning like `Chain will be a required parameter in the next version.`?

starknet_py/net/full_node_client.py
war-in: I can't see a difference with stacklevel=2

starknet_py/net/account/account_client.py
war-in: Actually, the chain won't be a required parameter. It will be optional, but chain or signer must be provided and we have an exception for that.

starknet_py/net/account/account_client.py
```diff
@@ -55,39 +55,48 @@ def __init__(
...
```
cptartur, line 67: Same as below.

cptartur (approved):

PR #293, opened 2022-08-08
Fix error when creating GatewayClient with custom gateway url

war-in:

It fixes a bug

* **What is the current behaviour?** (You can also link to an open issue here)

When providing net as dict with gateway_url and _feeder_gateway_url there was an error while computing the chain property.

There are no wrong errors and tests were added to avoid such a situation

No

### DIFF
--- a/starknet_py/tests/e2e/client/gateway_test.py
+++ b/starknet_py/tests/e2e/client/gateway_test.py
@@ -3,7 +3,7 @@ import pytest
 from starknet_py.net.client_models import TransactionStatusResponse, TransactionStatus
 from starknet_py.net.gateway_client import GatewayClient
 from starknet_py.net.models import StarknetChainId
-from starknet_py.net.networks import TESTNET, MAINNET, net_address_from_net
+from starknet_py.net.networks import TESTNET, MAINNET
 
 
 @pytest.mark.asyncio
@@ -54,12 +54,14 @@ async def test_get_transaction_status(invoke_transaction_hash, gateway_client):
 
 # pylint: disable=protected-access
 @pytest.mark.parametrize(
-    "net, chain",
-    ((TESTNET, StarknetChainId.TESTNET), (MAINNET, StarknetChainId.MAINNET)),
+    "net, net_address, chain",
+    (
+        (TESTNET, "https://alpha4.starknet.io", StarknetChainId.TESTNET),
+        (MAINNET, "https://alpha-mainnet.starknet.io", StarknetChainId.MAINNET),
+    ),
 )
-def test_creating_client_from_predefined_network(net, chain):
+def test_creating_client_from_predefined_network(net, net_address, chain):
     gateway_client = GatewayClient(net=net)
-    net_address = net_address_from_net(net)
 
     assert gateway_client.net == net
     assert gateway_client._feeder_gateway_client.url == f"{net_address}/feeder_gateway"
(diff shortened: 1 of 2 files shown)


starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -47,3 +50,53 @@ async def test_get_transaction_status(invoke_transaction_hash, gateway_client):
...
+    ((TESTNET, StarknetChainId.TESTNET), (MAINNET, StarknetChainId.MAINNET)),
+)
+def test_creating_client_from_predefined_network(net, chain):
+    gateway_client = GatewayClient(net=net)
+    net_address = net_address_from_net(net)
+
+    assert gateway_client.net == net
+    assert gateway_client._feeder_gateway_client.url == f"{net_address}/feeder_gateway"
+    assert gateway_client._gateway_client.url == f"{net_address}/gateway"
+    assert gateway_client.chain == chain
```
cptartur, line 67: I think in this case it'd be better to use values for urls directly rather than using `net_address_from_net`. If there was an error in the `net_address_from_net`, we wouldn't notice.

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -47,3 +50,53 @@ async def test_get_transaction_status(invoke_transaction_hash, gateway_client):
...
+
+
+def test_creating_client_with_custom_net(run_devnet):
+    gateway_client = GatewayClient(net=run_devnet, chain=StarknetChainId.TESTNET)
```
cptartur, line 71: What if we just used something like `net="customnetwork.net"`.

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -47,3 +50,53 @@ async def test_get_transaction_status(invoke_transaction_hash, gateway_client):
...
+    assert gateway_client._feeder_gateway_client.url == f"{run_devnet}/feeder_gateway"
+    assert gateway_client._gateway_client.url == f"{run_devnet}/gateway"
+    assert gateway_client.chain == StarknetChainId.TESTNET
+
+
+def test_creating_client_with_custom_net_dict(run_devnet):
+    net = {
+        "feeder_gateway_url": f"{run_devnet}/feeder_gateway",
+        "gateway_url": f"{run_devnet}/gateway",
+    }
```
cptartur, line 83: Same here

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -47,3 +50,53 @@ async def test_get_transaction_status(invoke_transaction_hash, gateway_client):
...
+
+    gateway_client = GatewayClient(net=net, chain=StarknetChainId.TESTNET)
+
+    assert gateway_client.net == net
+    assert gateway_client._feeder_gateway_client.url == net["feeder_gateway_url"]
+    assert gateway_client._gateway_client.url == net["gateway_url"]
+    assert gateway_client.chain == StarknetChainId.TESTNET
+
+
+def test_throwing_on_custom_net_dict_without_chain(run_devnet):
```
cptartur, line 93: What if we used a custom net but as a string and not a dict. Perhaps we should check if en error is also raised in that case.

cptartur (changes requested):

starknet_py/tests/e2e/client/gateway_test.py
war-in: We could do it

cptartur (approved):

PR #298, opened 2022-08-09
Support signing declare transactions

cptartur:

This PR adds initial implementation of support of signed declare transactions. We need to wait for release of new StarkNet to verify it's correctness though.

Signed declare transactions are not supported.

as above

Included new parameter in AccountClient.execute, but I've designed these changes so there aren't breaking existing stuff. Added deprecation warning to one parameter of execute.

## **Other information**

n/a

### DIFF
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -314,22 +314,45 @@ class AccountClient(Client):
 
     async def execute(
         self,
-        calls: Calls,
+        calls: Calls = None,
         max_fee: Optional[int] = None,
         auto_estimate: bool = False,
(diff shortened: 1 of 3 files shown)


Arcticae:
@cptartur Can we also add the support for the `StarkCurveSigner` to sign the declare tx as well?

cptartur:
`StarkCurveSigner` (or any signer that implements the `Signer`) should be able to sign declare transactions already? Unless there's some special flow for signing these, in which case we need a separate method.

cptartur:
After discussing with @Solpatium, we think deprecating modifying `execute` this way is not a good approach.

PR #305, opened 2022-08-12
Remove usage of deprecated features

### LINKED ISSUE #301: Remove usages of deprecated features through the library
We've recently added a deprecation warning to some of the existing features including e.g. chain_id. However our tests and library has not been completely updated to get rid of the usages of the deprecated features as indicated by the errors shown when running tests (`poe test`). Tests and library should be to make sure we aren't relying on deprecated features ourselves.

war-in:

Closes #301 

Our tests are using chain id which is deprecated

Deprecation warnings are reduced

No

## **Other information**

Can't resolve estimate_fee warning on providing max_fee=0. Will work on it


starknet_py/net/full_node_client.py
```diff
@@ -65,7 +65,12 @@ def __init__(
         """
         self.url = node_url
         self._client = RpcHttpClient(url=node_url, session=session)
-        self._chain = chain_from_network(net, chain)
+
+        try:
+            self._chain = chain_from_network(net, chain)
+        except ValueError:
+            self._chain = None
```
Solpatium, line 72: Why can't we just check if chain was provided and only then use chain_from_network?

war-in: Because if someone is using testnet or mainnet there is no need to provide a chain and still have a chain property correctly set

war-in: I can add this:
if chain is not None and net in ["mainnent", "testnet"]:
   self.chain = chain_form_network

Solpatium: I'd rather do this.

Arcticae (changes requested):

starknet_py/net/gateway_client.py
```diff
@@ -76,7 +76,12 @@ def __init__(
             gateway_url = net["gateway_url"]
 
         self._net = net
-        self._chain = chain_from_network(net, chain)
+
+        if net in ["testnet", "mainnet"]:
+            self._chain = chain_from_network(net, chain)
+        elif chain is not None:
+            self._chain = chain
```
Arcticae, line 83:
```suggestion
        if net in ["testnet", "mainnet"]:
            chain = chain_from_network(net, chain)
         self._chain = chain
```

starknet_py/net/full_node_client.py
```diff
@@ -65,7 +65,12 @@ def __init__(
         """
         self.url = node_url
         self._client = RpcHttpClient(url=node_url, session=session)
-        self._chain = chain_from_network(net, chain)
+
+        if net in ["testnet", "mainnet"]:
+            self._chain = chain_from_network(net, chain)
+        elif chain is not None:
+            self._chain = chain
```
Arcticae, line 72:
```suggestion
        if net in ["testnet", "mainnet"]:
            chain = chain_from_network(net, chain)
        self._chain = chain
```

Arcticae (changes requested):

Arcticae (approved):

PR #308, opened 2022-08-16
Fix/cairo-serializer-kwargs

### LINKED ISSUE #302: CairoSerializer should throw on unknown keyword arguments
In `CairoSerializer.from_python` from `starknet_py/utils/data_transformer/data_transformer.py` we don't check if unnecessary named arguments are passed. 

We only check that for positional params:
```
        if len(args) > len(type_by_name):
            raise TypeError(
                f"Provided {len(args)} positional arguments, {len(type_by_name)} possible."
            )
```

drknzz:

Closes #302 

Adds check for unnecessary keyword arguments in CairoSerializer

Adds tests for CairoSerializer


starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -344,6 +344,10 @@ def from_python(
                 f"Provided {len(args)} positional arguments, {len(type_by_name)} possible."
             )
 
+        for key, val in named_arguments.items():
+            if key not in type_by_name:
+                raise TypeError(f"Unnecessary named argument {key}={val} provided.")
+
```
war-in, line 350: That will throw only one wrong parameter but could throw all. Maybe go through the entire dict and collect all of them?

starknet_py/tests/e2e/conftest.py
```diff
@@ -32,6 +33,32 @@
     "0x60D7C88541F969520E46D39EC7C9053451CFEDBC2EEB847B684981A22CD452E"
 )
 
+CONTRACT = """
```
war-in, line 36: I'm not sure this should be here. Maybe another .cairo file in the mocked_contracts_dir?

starknet_py/tests/e2e/conftest.py
```diff
@@ -192,3 +219,15 @@ def deploy_erc20_contract(gateway_account_client, erc20_source_code) -> Contract
 @pytest.fixture(name="compiled_proxy")
 def compiled_proxy(request) -> str:
     return (directory_with_contracts / request.param).read_text("utf-8")
+
+
+@pytest.mark.asyncio
+@pytest.fixture(name="cairo_serializer")
```
war-in, line 225: scope="module"? Or do we always need it as a new object?

war-in (changes requested):

war-in (approved):

PR #319, opened 2022-08-24
Add sign_message and verify_message

drknzz:

### DIFF
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -39,6 +39,8 @@ from starknet_py.utils.crypto.facade import Call
 from starknet_py.utils.data_transformer.execute_transformer import execute_transformer
 from starknet_py.utils.sync import add_sync_methods
 from starknet_py.net.models.address import AddressRepresentation, parse_address
+from starknet_py.utils.typed_data.functions import get_message_hash
+from starknet_py.utils.typed_data.types import TypedData
 
 
 @add_sync_methods
@@ -402,6 +404,72 @@ class AccountClient(Client):
             address=address,
             signer=signer,
         )
+
+    def sign_message(self, typed_data: TypedData) -> List[int]:
+        """
+        Sign an TypedData object for off-chain usage with the starknet private key and return the signature
+        This adds a message prefix, so it can't be interchanged with transactions
+
+        :param typed_data: TypedData object to be signed
+        :return: The signature of the TypedData object
+        """
+        return self.signer.sign_message(typed_data, self.address)
+
+    def hash_message(self, typed_data: TypedData) -> int:
+        """
+        Hash a TypedData object with pedersen hash and return the hash
+        This adds a message prefix, so it can't be interchanged with transactions
+
+        :param typed_data: TypedData object to be hashed
+        :return: the hash of the TypedData object
+        """
+        return get_message_hash(typed_data, self.address)
+
+    async def verify_message(self, typed_data: TypedData, signature: List[int]) -> bool:
+        """
+        Verify a signature of a TypedData object
+
+        :param typed_data: TypedData object to be verified
+        :param signature: signature of the TypedData object
+        :return: true if the signature is valid, false otherwise
+        """
+        msg_hash = self.hash_message(typed_data)
+        return await self.verify_message_hash(msg_hash, signature)
+
+    async def verify_message_hash(self, msg_hash: int, signature: List[int]) -> bool:
+        """
+        Verify a signature of a given hash
+
+        :param msg_hash: hash to be verified
+        :param signature: signature of the hash
+        :return: true if the signature is valid, false otherwise
+        """
+        calldata = {"hash": msg_hash, "signature": signature}
+        compiled_calldata = compile_calldata(calldata)
+
+        invoke_tx = InvokeFunction(
+            contract_address=self.address,
+            entry_point_selector=get_selector_from_name("is_valid_signature"),
+            calldata=compiled_calldata,
+            signature=[],
+            max_fee=0,
+            version=0,
+        )
+        try:
+            await self.call_contract(invoke_tx=invoke_tx)
+            return True
+        # pylint: disable=bare-except
+        except:
+            return False
+
+
+def compile_calldata(calldata) -> List[int]:
+    def compile_val(val):
+        if isinstance(val, int):
+            return [val]
+        return [len(val)] + list(val)
+
+    return [x for value in calldata.values() for x in compile_val(value)]
 
 
 async def deploy_account_contract(
--- a/starknet_py/net/signer/base_signer.py
+++ b/starknet_py/net/signer/base_signer.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 from typing import List
 
 from starknet_py.net.models.transaction import Transaction
+from starknet_py.utils.typed_data.types import TypedData
 
 
 class BaseSigner(ABC):
@@ -19,13 +20,21 @@ class BaseSigner(ABC):
         """
 
     @abstractmethod
-    def sign_transaction(
-        self,
-        transaction: Transaction,
-    ) -> List[int]:
+    def sign_transaction(self, transaction: Transaction) -> List[int]:
         """
         Sign execute transaction and return a signature
 
         :param transaction: Execute transaction to sign
         :return: transaction signature
+        """
+
+    @abstractmethod
+    def sign_message(self, typed_data: TypedData, account_address: int) -> List[int]:
+        """
+        Sign an TypedData object for off-chain usage with the starknet private key and return the signature
+        This adds a message prefix, so it can't be interchanged with transactions
+
+        :param typed_data: TypedData object to be signed
+        :param account_address: account address
+        :return: the signature of the JSON object
         """
--- a/starknet_py/utils/typed_data/functions.py
+++ b/starknet_py/utils/typed_data/functions.py
@@ -1,0 +1,91 @@
+from typing import List, Union
+
+from starkware.cairo.common.hash_state import compute_hash_on_elements
+from starkware.starknet.public.abi import get_selector_from_name
+
+from starknet_py.cairo.felt import encode_shortstring
+from starknet_py.utils.typed_data.types import TypedData
+
+
+def get_hex(value: Union[int, str]):
+    if isinstance(value, str):
+        if value[:2] == "0x":
+            return value
+        return hex(encode_shortstring(value))
+    return hex(value)
+
+
+def get_dependencies(typed_data: TypedData, type_: str) -> List[str]:
+    if type_ not in typed_data.types:
+        return []
+    dependencies = set()
+
+    def get_deps(type_: str):
+        for item in typed_data.types[type_]:
+            item_type = item["type"]
+            if item_type in typed_data.types and item_type not in dependencies:
+                dependencies.add(item_type)
+                get_deps(item_type)
+
+    get_deps(type_)
+    return [type_, *list(dependencies)]
+
+
+def encode_type(typed_data: TypedData, type_: str) -> str:
+    [primary, *dependencies] = get_dependencies(typed_data, type_)
+    types = [primary, *sorted(dependencies)]
+
+    def format_dependency(dependency):
+        lst = [f"{t['name']}:{t['type']}" for t in typed_data.types[dependency]]
+        res = f"{dependency}("
+        for idx, item in enumerate(lst):
+            if idx > 0:
+                res += ","
+            res += item
+        return res + ")"
+
+    return "".join([format_dependency(x) for x in types])
+
+
+def get_type_hash(typed_data: TypedData, type_: str) -> int:
+    return get_selector_from_name(encode_type(typed_data, type_))
+
+
+def encode_value(typed_data: TypedData, type_: str, value: Union[int, str]) -> str:
+    if type_ in typed_data.types:
+        return [type_, get_struct_hash(typed_data, type_, value)]
+    return [type_, int(get_hex(value), 16)]
+
+
+def encode_data(typed_data: TypedData, type_: str, data: dict) -> List[List[str]]:
+    def encode(field):
+        if data[field["name"]] is None:
+            raise ValueError(f"Cannot encode data: missing data for {field['name']}")
+        value = data[field["name"]]
+        return encode_value(typed_data, field["type"], value)
+
+    types = ["felt"]
+    values = [get_type_hash(typed_data, type_)]
+
+    for item in typed_data.types[type_]:
+        typ, val = encode(item)
+        types.append(typ)
+        values.append(val)
+
+    return [types, values]
+
+
+def get_struct_hash(typed_data: TypedData, type_: str, data: dict) -> int:
+    encoded_data = encode_data(typed_data, type_, data)
+    return compute_hash_on_elements(encoded_data[1])
+
+
+def get_message_hash(typed_data: TypedData, account_address: int) -> int:
+    message = [
+        encode_shortstring("StarkNet Message"),
+        get_struct_hash(typed_data, "StarkNetDomain", typed_data.domain),
+        account_address,
+        get_struct_hash(typed_data, typed_data.primary_type, typed_data.message),
+    ]
+
+    return compute_hash_on_elements(message)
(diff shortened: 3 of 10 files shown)


starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,94 @@
...
+    types = [primary, *sorted(dependencies)]
+
+    def format_dependency(dependency):
+        lst = [f"{t['name']}:{t['type']}" for t in typed_data.types[dependency]]
+        res = f"{dependency}("
+        for idx, item in enumerate(lst):
+            if idx > 0:
+                res += ","
+            res += item
+        return res + ")"
```
Solpatium, line 48: `f"{dependency}({','.join(lst)})"`

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,94 @@
...
+    get_deps(type_)
+    return [type_, *list(dependencies)]
+
+
+def encode_type(typed_data: TypedData, type_: str) -> str:
+    [primary, *dependencies] = get_dependencies(typed_data, type_)
+    types = [primary, *sorted(dependencies)]
+
+    def format_dependency(dependency):
+        lst = [f"{t['name']}:{t['type']}" for t in typed_data.types[dependency]]
```
Solpatium, line 42: please use a better name

Solpatium (changes requested):

starknet_py/utils/typed_data/types.py
```diff
@@ -0,0 +1,18 @@
...
+
+class StarkNetDomain(TypedDict):
+    name: Optional[str]
+    version: Optional[str]
+    chainId: Optional[Union[str, int]]
+
+
+@dataclass(frozen=True)
+class TypedData:
+    types: dict
```
cptartur, line 15: Types have know typing `Dict[str, List[Dict[str, str]]]`so this could have better typing. 

I'd actually consider defining more data classes and making this typed more clearly. We could then use Marshmallow schema to accept typeddata as dict or typeddict from users (so it's convenient for them) and then parsing it to classes that are easy for us to operate on.

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,89 @@
...
+        typ, val = encode(item)
+        types.append(typ)
+        values.append(val)
+
+    return [types, values]
+
+
+def get_struct_hash(typed_data: TypedData, type_: str, data: dict) -> int:
+    encoded_data = encode_data(typed_data, type_, data)
+    return compute_hash_on_elements(encoded_data[1])
```
cptartur, line 78: This could be modelled more clearly following specification:
* parse data to get Struct
* then `return compute_hash_on_elements(encode_type(Struct), *values)`

You can consider defining an abstraction for Struct with helper methods to return it's values, types, etc.

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,89 @@
...
+    return get_selector_from_name(encode_type(typed_data, type_))
+
+
+def encode_value(typed_data: TypedData, type_: str, value: Union[int, str]) -> str:
+    if type_ in typed_data.types:
+        return [type_, get_struct_hash(typed_data, type_, value)]
+    return [type_, int(get_hex(value), 16)]
+
+
+def encode_data(typed_data: TypedData, type_: str, data: dict) -> List[List[str]]:
```
cptartur, line 58: This method is trying to do too much at the same time

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,89 @@
...
+            item_type = item["type"]
+            if item_type in typed_data.types and item_type not in dependencies:
+                dependencies.add(item_type)
+                get_deps(item_type)
+
+    get_deps(type_)
+    return [type_, *list(dependencies)]
+
+
+def encode_type(typed_data: TypedData, type_: str) -> str:
```
cptartur, line 37: This should accept some better representation of Struct than the entire TypedData. In shouldn't be the concern of this method to also parse the data.

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,89 @@
...
+def get_hex(value: Union[int, str]):
+    if isinstance(value, str):
+        if value[:2] == "0x":
+            return value
+        return hex(encode_shortstring(value))
+    return hex(value)
+
+
+def get_dependencies(typed_data: TypedData, type_: str) -> List[str]:
+    if type_ and type_[-1] == "*":
```
cptartur, line 19: What's the purpose of this check?

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,89 @@
...
+
+def get_hex(value: Union[int, str]):
+    if isinstance(value, str):
+        if value[:2] == "0x":
+            return value
+        return hex(encode_shortstring(value))
+    return hex(value)
+
+
+def get_dependencies(typed_data: TypedData, type_: str) -> List[str]:
```
cptartur, line 18: I'm not sure if this is a good name for the metod.

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,89 @@
...
+    types = [primary, *sorted(dependencies)]
+
+    def make_dependency_str(dependency):
+        lst = [f"{t['name']}:{t['type']}" for t in typed_data.types[dependency]]
+        return f"{dependency}({','.join(lst)})"
+
+    return "".join([make_dependency_str(x) for x in types])
+
+
+def get_type_hash(typed_data: TypedData, type_: str) -> int:
```
cptartur, line 48: Why not just `type_hash`

starknet_py/utils/typed_data/functions.py
```diff
@@ -0,0 +1,89 @@
+from typing import List, Union
```
cptartur, line 1: For me this module is too complicated to comprehend. I think you should introduce more abstraction to make it easier to comprehend.

Also, variables like `typed_data` are getting passed around bym multiple methods. If they are really necessary in every one of them, this should be converted into a class with necessary values assigned to `self`. But again, I think there should be more abstraction.

starknet_py/net/account/account_client.py
```diff
@@ -403,6 +407,63 @@ async def create_account(
...
+        """
+        Hash a TypedData object with pedersen hash and return the hash
+        This adds a message prefix, so it can't be interchanged with transactions
+
+        :param typed_data: TypedData object to be hashed
+        :return: the hash of the TypedData object
+        """
+        return get_message_hash(typed_data, self.address)
+
+    async def verify_message(self, typed_data: TypedData, signature: List[int]) -> bool:
```
cptartur, line 430: I'd be more convenient for the user for the TypedData here to be typeddict and not dataclass. That way if the user has for example a json returned from somewhere it can be passed directly to this method. Currently they'd have to instantiate the dataclass themselves.

cptartur (changes requested):
I haven't done any in depth review of the `functions` module, except for the most "high level" methods: I think we need more abstractions in this module, because currently it's relying a lot on passing dict arounds and is quite difficult to comprehend.

starknet_py/net/networks.py
```diff
@@ -1,10 +1,7 @@
 from typing import Union
-import sys
 
-if sys.version_info >= (3, 8):
-    from typing import TypedDict
-else:
-    from typing_extensions import TypedDict
```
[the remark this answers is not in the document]
drknzz: To avoid code duplication of this conditional import, it has been moved to `starknet_py/utils/typing.py`.

starknet_py/utils/typing.py
```diff
@@ -0,0 +1,7 @@
+import sys
+
+# pylint: disable=unused-import
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
```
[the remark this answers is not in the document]
drknzz: Avoidance of conditional TypedDict import code duplication.
From now on TypedDict can be imported ordinarily from this module.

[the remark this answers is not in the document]
cptartur: I don't know if I like this approach, this can be confusing. What do you think @Arcticae @Solpatium

starknet_py/net/account/account_client.py
```diff
@@ -403,6 +407,63 @@ async def create_account(
...
+
+        :param msg_hash: hash to be verified
+        :param signature: signature of the hash
+        :return: true if the signature is valid, false otherwise
+        """
+        calldata = [msg_hash, len(signature), *signature]
+
+        invoke_tx = InvokeFunction(
+            contract_address=self.address,
+            entry_point_selector=get_selector_from_name("is_valid_signature"),
```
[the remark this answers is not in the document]
drknzz: I think so.

[This post](https://github.com/argentlabs/argent-x/discussions/14) says:

> Wallets have to generate signatures for 2 different types of data: transactions that will be sent to an Account Contract (AC) via the sequencer, and off-chain messages that will be returned to the requesting dapp for verification by calling the `is_valid_signature` method on the AC.

starknet_py/net/account/account_client.py
cptartur: Why wasn't this addessed

starknet_py/net/signer/base_signer.py
```diff
@@ -19,13 +20,21 @@ def public_key(self) -> int:
...
+    def sign_transaction(self, transaction: Transaction) -> List[int]:
         """
         Sign execute transaction and return a signature
 
         :param transaction: Execute transaction to sign
         :return: transaction signature
         """
+
+    @abstractmethod
+    def sign_message(self, typed_data: TypedData, account_address: int) -> List[int]:
```
cptartur, line 32: This should also accept typed_data as a typeddict instead of dataclass

starknet_py/tests/e2e/conftest.py
```diff
@@ -194,6 +197,20 @@ def compiled_proxy(request) -> str:
     return (directory_with_contracts / request.param).read_text("utf-8")
 
 
+@pytest.fixture(name="typed_data")
```
cptartur, line 200: This could be a parametrized fixture but this can be addressed in separate PR once #356 is merged.

starknet_py/utils/typed_data.py
```diff
@@ -0,0 +1,115 @@
...
+    Dataclass representing a TypedData object
+    """
+
+    types: Dict[str, List[Parameter]]
+    primary_type: str
+    domain: StarkNetDomain
+    message: dict
+
+    def _encode_value(self, type_: str, value: Union[int, str]) -> str:
+        if type_[-1] == "*":
```
cptartur, line 43: I'd defined a method like `is_pointer` so it's more explicit

starknet_py/utils/typed_data.py
```diff
@@ -0,0 +1,115 @@
...
+            return compute_hash_on_elements(
+                [self.struct_hash(type_[:-1], data) for data in value]
+            )
+        if type_ in self.types:
+            return self.struct_hash(type_, value)
+        return int(get_hex(value), 16)
+
+    def _encode_data(self, type_: str, data: dict) -> List[str]:
```
cptartur, line 51: I'd use name like `type_name` or something along the lines here and in other methods.

starknet_py/utils/typed_data.py
```diff
@@ -0,0 +1,115 @@
...
+        if type_ not in self.types:
+            # type_ is a primitive type, has no dependencies
+            return []
+
+        dependencies = set()
+
+        def collect_deps(type_: str) -> None:
+            for param in self.types[type_]:
+                # strip the pointer
+                fixed_type = param.type[:-1] if param.type[-1] == "*" else param.type
```
cptartur, line 69: I'd define a helper method of this

starknet_py/utils/typed_data.py
```diff
@@ -0,0 +1,115 @@
...
+                    dependencies.add(fixed_type)
+                    # recursive call
+                    collect_deps(fixed_type)
+
+        # collect dependencies into a set
+        collect_deps(type_)
+        return [type_, *list(dependencies)]
+
+    def _encode_type(self, type_: str) -> str:
+        [primary, *dependencies] = self._get_dependencies(type_)
```
cptartur, line 80:
```suggestion
        primary, *dependencies = self._get_dependencies(type_)
```
I think brackets are not needed here.

cptartur (changes requested):
I like the improvements to the typed_data module. However user interface in AccountClient and signer still needs improvements. 

What I'd like to have is a method in the typed_data like `parse_typed_data` that accepts a typeddict and returns a TypedData dataclass instance.

I'd also use this typeddict as an input type for both account_client and signer methods related to signed messages. With the current approach users either have to construct whole dataclass themselves or use our schema which we aren't exposing as a part of the public API.

starknet_py/net/models/typed_data.py
```diff
@@ -0,0 +1,37 @@
+"""
+TypedDict structures for TypedData
+"""
+
+from typing import Dict, List, Union
+
+from typing import TypedDict
```
cptartur, line 7: Combine these?

starknet_py/net/models/typed_data.py
```diff
@@ -0,0 +1,37 @@
...
+
+class TypedData(TypedDict):
+    """
+    TypedDict representing a TypedData object
+    """
+
+    types: Dict[str, List[Parameter]]
+    primary_type: str
+    domain: StarkNetDomain
+    message: dict
```
cptartur, line 37: I believe this would be `Dict[Str, Any]` but I'm not sure about this.

starknet_py/net/networks.py
```diff
@@ -1,10 +1,5 @@
+from typing import TypedDict
 from typing import Union
```
cptartur, line 2: Combine this

starknet_py/net/signer/base_signer.py
```diff
@@ -19,13 +20,21 @@ def public_key(self) -> int:
...
+        """
+        Sign an TypedData object for off-chain usage with the starknet private key and return the signature
```
cptartur, line 34: Update docstring

starknet_py/tests/e2e/conftest.py
```diff
@@ -194,6 +196,22 @@ def compiled_proxy(request) -> str:
...
+    params=["typed_data_example.json", "typed_data_struct_array_example.json"],
+)
+def typed_data(request) -> TypedData:
+    file_name = getattr(request, "param")
+
+    directory = Path(os.path.dirname(__file__))
+    file_path = directory / "account" / file_name
+
+    with open(file_path, "r", encoding="utf-8") as file:
+        typed_data = json.load(file)
```
cptartur, line 210: I think you can do `file_path.read_text("utf-8")`, so you don't have to manage opening the file.

starknet_py/net/account/account_client.py
```diff
@@ -576,6 +580,65 @@ async def get_contract_nonce(
...
+            version=0,
+            nonce=None,
+        )
+        try:
+            await self.call_contract(invoke_tx=invoke_tx)
+            return True
+        except ClientError as ex:
+            if re.search(r"Signature\s.+,\sis\sinvalid", ex.message):
+                return False
+            raise ex
```
cptartur, line 640: I'm not sure if we should just be re-raising whatever exception we got here. Perhaps `raise Exception("some message") from ex` would be a more ergonomic approach for the end user? 

cc @Solpatium

cptartur (commented):
Good work so far! I think addition of helper methods in TypedData made it a bit more easier to comprehend. 

Couple of comments though.

starknet_py/net/account/account_client.py
cptartur: Alright, we can leave it as it is then

starknet_py/net/models/typed_data.py
cptartur: I meant the `message` parameter 😅. Types was typed correctly with `List[Parameter]`.

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -225,11 +225,10 @@ async def test_sign_and_verify_offchain_message(gateway_account_client, typed_da
 
 
 @pytest.mark.asyncio
-async def test_get_class_hash_at(map_contract, account_clients):
-    for account_client in account_clients:
-        class_hash = await account_client.get_class_hash_at(
-            map_contract.address, block_hash="latest"
-        )
```
cptartur: Why was this removed?

cptartur (approved):
Good job!

PR #329, opened 2022-08-30
Bump python to 3.9.12

cptartur:

Cairo-lang dropped support for lower python versions. This PR bumps minimum python version to 3.9.12, removes checks for python 3.7 and 3.8, adds checks for python 3.10.

n/a

3.7 and 3.8 not supported

Yes, breaks package for users on python 3.7 and 3.8

## **Other information**

n/a


pyproject.toml
```diff
@@ -14,7 +14,7 @@ repository = "https://github.com/software-mansion/starknet.py"
 documentation = "https://starknetpy.rtfd.io/"
 
 [tool.poetry.dependencies]
-python = ">=3.7.2,<3.10"
+python = ">=3.9.12, <=3.10"
```
Solpatium, line 17: Why do we require 3.9.12? I think we can accept any 3.9.x

Solpatium (approved):

PR #303, opened 2022-08-11
Support signed declare

cptartur:
### **Please check if the PR fulfills these requirements**
- [ ] The formatter, linter, and tests all run without an error
- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

## **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Closes #289 

This PR doesn't include tests for the new method as it will have to be added once signed declare is actually supported in StarkNet.

* **What is the current behavior?** (You can also link to an open issue here)

Only unsigned declare is supported

## **What is the new behavior (if this is a feature change)?**

as above

## **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

Yes
* transactions module was moved to net
* transactions module is deprecated
* deprecated AccountClient.sign_transaction

## **Other information**

n/a

PR #336, opened 2022-08-30
Add get_nonce to client

drknzz:

### DIFF
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -450,13 +450,15 @@ class AccountClient(Client):
             supported_tx_version=version,
         )
 
-    async def get_nonce(
+    async def get_contract_nonce(
         self,
         contract_address: int,
         block_hash: Optional[Union[Hash, Tag]] = None,
         block_number: Optional[Union[int, Tag]] = None,
     ) -> int:
-        return self._get_nonce()
+        return await self.client.get_contract_nonce(
+            contract_address, block_hash, block_number
+        )
 
 
 async def deploy_account_contract(
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -426,6 +426,14 @@ class AccountClient(Client):
             signer=signer,
         )
 
+    async def get_nonce(
+        self,
+        contract_address: int,
+        block_hash: Optional[Union[Hash, Tag]] = None,
+        block_number: Optional[Union[int, Tag]] = None,
+    ) -> int:
+        return self._get_nonce()
+
 
 async def deploy_account_contract(
     client: Client, public_key: int
--- a/starknet_py/net/account/account_client.py
+++ b/starknet_py/net/account/account_client.py
@@ -193,6 +193,9 @@ class AccountClient(Client):
         return await self.client.get_class_by_hash(class_hash=class_hash)
 
     async def _get_nonce(self) -> int:
+        if self.supported_tx_version == 1:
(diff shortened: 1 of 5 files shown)


starknet_py/net/account/account_client.py
```diff
@@ -450,6 +450,14 @@ async def create_account(
             supported_tx_version=version,
         )
 
+    async def get_nonce(
+        self,
+        contract_address: int,
+        block_hash: Optional[Union[Hash, Tag]] = None,
+        block_number: Optional[Union[int, Tag]] = None,
+    ) -> int:
+        return self._get_nonce()
```
Solpatium, line 459: `get_nonce` checks nonce of an address provided as the argument. Internal _get_nonce should call either this method (in case of tx version = 1 ) or use the logic it is using now (tx version = 0).

starknet_py/net/client.py
```diff
@@ -271,3 +271,19 @@ async def get_class_by_hash(self, class_hash: Hash) -> DeclaredContract:
         :param class_hash: Class hash
         :return: ContractClass object
         """
+
+    @abstractmethod
+    async def get_nonce(
```
Solpatium, line 276: Let's call it `get_contract_nonce`. This avoids confusion in case of AccountClient.get_nonce which suggests it is account's nonce.

starknet_py/net/full_node_client.py
```diff
@@ -398,6 +398,14 @@ async def get_pending_transactions(self) -> List[Transaction]:
 
         return PendingTransactionsSchema().load(res, unknown=EXCLUDE)
 
+    async def get_nonce(
+        self,
+        contract_address: int,
+        block_hash: Optional[Union[Hash, Tag]] = None,
+        block_number: Optional[Union[int, Tag]] = None,
+    ) -> int:
+        raise NotImplementedError
```
Solpatium, line 407: We can use it already https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json#L629

starknet_py/net/gateway_client.py
```diff
@@ -363,6 +363,26 @@ async def get_code(
...
+            block_hash=block_hash, block_number=block_number
+        )
+        params = {
+            **{"contractAddress": convert_to_felt(contract_address)},
+            **block_identifier,
+        }
+
+        nonce = await self._feeder_gateway_client.call(
+            method_name="get_nonce", params=params
+        )
```
Solpatium, line 382: This should use new endpoint for getting nonce: https://www.starknet.io/docs/0.10.0/hello_starknet/cli.html#get-nonce. The nonce is now moved to the protocol level, starknet itself will maintain and increment it.

Solpatium (changes requested):

cptartur (approved):
I'd merge it now because i need nonces in #303 and add tests in separate PR.

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -88,3 +88,14 @@ def test_creating_client_with_custom_net_dict():
     assert gateway_client.net == net
     assert gateway_client._feeder_gateway_client.url == net["feeder_gateway_url"]
     assert gateway_client._gateway_client.url == net["gateway_url"]
+
+
+@pytest.mark.asyncio
+async def test_get_contract_nonce():
+    gateway_client = GatewayClient(net="https://external.integration.starknet.io")
```
cptartur, line 95: Use `gateway_client` from fixture like in other tests.

PR #343, opened 2022-08-31
0.5.0 migration guide

drknzz:


docs/migration_guide.rst
```diff
@@ -1,3 +1,84 @@
...
+
+Currently, compatible Python versions are ``python3.8`` and ``python3.9``.
+
+New Transaction version
+-----------------------
+
+Cairo 0.10.0 brings new version of the transaction.
+The differences:
+
+- ``contract_address`` field is now called ``account_contract_address``
```
war-in, line 65: It hasn't chanded. It is mentioned in the notion, but there is no difference in the package

docs/migration_guide.rst
```diff
@@ -1,3 +1,84 @@
+0.4.8 Migration guide
+=====================
+
+0.4.8 of starknet.py brings fresh changes including a breaking change in contracts' code.
```
war-in, line 4: Maybe we should mention that this breaking change doesn't come from starknet.py?

docs/migration_guide.rst
```diff
@@ -1,19 +1,68 @@
...
 
-New transaction version
+- new cairo syntax,
+- new transaction version (1),
+- new ``__validate__`` endpoint in accounts.
+
+``starknet.py`` 0.5.0 has an experimental support for new features and tries to minimize number of breaking changes for
+users who want to use the old transaction version (0). Please note that support for this transaction version will be
+removed in the future.
+
```
Solpatium, line 13: Add a note that informs users that they can use the previous version if they don't need any new features.

Solpatium (changes requested):

docs/migration_guide.rst
```diff
@@ -1,19 +1,76 @@
...
+.. note::
+
+    There is no need to upgrade ``starknet.py`` to the newest version because the old one is still compatible with StarkNet.
```
cptartur, line 16: Are we sure about this line?

Solpatium (approved):

cptartur (commented):
One question, otherwise looks very good

PR #345, opened 2022-09-01
Fix/default nonce invoke function

Solpatium:
InvokeFunction doesn't set it by default. 

I haven't tested it yet.


Solpatium:
We decided this can remain as a breaking change.

PR #349, opened 2022-09-02
Feat/add missing fields to models

war-in:

### DIFF
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -90,7 +90,6 @@ class Transaction(ABC):
     signature: List[int]
     max_fee: int
     version: int
-    nonce: int
 
     def __post_init__(self):
         if self.__class__ == Transaction:
@@ -106,6 +105,7 @@ class InvokeTransaction(Transaction):
     contract_address: int
     calldata: List[int]
     entry_point_selector: int
+    nonce: Optional[int] = None
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


starknet_py/net/schemas/rpc.py
```diff
@@ -105,12 +105,14 @@ class TransactionSchema(Schema):
     hash = Felt(data_key="transaction_hash", required=True)
     signature = fields.List(Felt(), data_key="signature", load_default=[])
     max_fee = Felt(data_key="max_fee", load_default=0)
+    version = Felt(data_key="version", load_default=0)
```
cptartur, line 108: I'm not sure about this. Are there cases where the versions is missing according to the specs? So we need this `load_default`.

starknet_py/net/client_models.py
```diff
@@ -150,14 +154,28 @@ class TransactionReceipt:
...
+    def version(self) -> int:
+        warnings.warn(
+            "Field version is deprecated as it is always incorrectly set to 0. It will be removed in the future",
+            category=DeprecationWarning,
+        )
+        return 0
+
+    @version.setter
+    def version(self, v) -> None:
+        pass
```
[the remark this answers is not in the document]
war-in: Unfortunately, it has to be done this way. It is used by TransactionReceipt __init__ and can't raise any exceptions

starknet_py/net/schemas/rpc.py
war-in: Yeah, you're right. Will fix that

cptartur (approved):

PR #356, opened 2022-09-06
Add new account tests

cptartur:
### **Please check if the PR fulfills these requirements**
- [ ] The formatter, linter, and tests all run without an error
- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

## **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

This PR adds parametrized `account_clients` and `map_contracts` which allows using existing tests with both new and old version of account client. 

This is a work in progress, as to better test this feature, devnet needs to be updated and necessary configurations need to be added.

* **What is the current behavior?** (You can also link to an open issue here)

It is required to use for loops in tests.

## **What is the new behavior (if this is a feature change)?**

Tests are correctly parametrized.

## **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

No

## **Other information**

n/a

PR #357, opened 2022-09-06
Adapt tests to run on integration

war-in:

It changes e2e tests to be able to run on integration

Only part of the tests can be run on integration

All tests can be run

No


starknet_py/tests/e2e/client/prepare_net_for_gateway_test.py
```diff
@@ -0,0 +1,50 @@
+from starknet_py.contract import Contract
+from starknet_py.net import AccountClient
+from starknet_py.transactions.declare import make_declare_tx
+
+
+def prepare_net_for_tests(account_client: AccountClient, compiled_contract: str):
```
cptartur, line 6: If #356 is merged, this could probably be changed to async function so don't have to use `_sync` methods.

starknet_py/tests/e2e/docs/guide/test_deploying_new_contracts.py
```diff
@@ -45,13 +45,15 @@ async def test_deploying_new_contracts(gateway_client):
     deployment_result = await Contract.deploy(
         client, compilation_source=contract, constructor_args=constructor_args
     )
+    await deployment_result.wait_for_acceptance()
```
cptartur, line 48: This shouldn't be included in docs I think

starknet_py/tests/e2e/client/conftest.py
```diff
@@ -1,57 +1,54 @@
...
-    assert re.search("^0x[a-fA-F0-9]{1,63}$", deploy_transaction_hash) is not None
-    assert re.search("^0x[a-fA-F0-9]{1,63}$", invoke_transaction_hash) is not None
-    assert re.search("^0x[a-fA-F0-9]{1,63}$", declare_transaction_hash) is not None
-
     prepared_data = {
-        "genesis_block": genesis_block,
-        "block": block,
-        "contract_address": int(contract_address, 16),
-        "deploy_transaction_hash": int(deploy_transaction_hash, 16),
+        "contract_address": contract_address,
```
Solpatium, line 44: What's the contract address?

starknet_py/tests/e2e/client/conftest.py
```diff
@@ -1,57 +1,54 @@
...
+    (
+        contract_address,
+        deploy_transaction_hash,
+        block_with_deploy_number,
+        block_with_deploy_hash,
+        invoke_transaction_hash,
+        block_with_invoke_number,
+        declare_transaction_hash,
+        block_with_declare_number,
+    ) = await prepare_net_for_tests(
```
Solpatium, line 39: Don't use tuples for returning so many values. Use dataclass instead.

starknet_py/tests/e2e/client/conftest.py
war-in: It's an address of deployed balance contract

starknet_py/tests/e2e/conftest.py
```diff
@@ -141,19 +168,55 @@ def address_and_private_key(pytestconfig):
...
 @pytest.fixture(scope="module")
-def account_clients(gateway_account_client, rpc_account_client):
-    return gateway_account_client, rpc_account_client
+def new_address_and_private_key(pytestconfig):
+    net = pytestconfig.getoption("--net")
+
+    account_details = {
+        # TODO configure for other envs
+        "devnet": None,
+        "testnet": None,
```
Solpatium, line 192: How does it work?

starknet_py/tests/e2e/contract_interaction/interaction_test.py
```diff
@@ -219,7 +214,11 @@ async def test_warning_when_max_fee_equals_to_zero(map_contract):
         DeprecationWarning,
         match=r"Transaction will fail with max_fee set to 0. Change it to a higher value.",
     ):
-        await map_contract.functions["put"].invoke(10, 20, max_fee=0)
+        # try except have to be added because when running on integration it will throw an error (max_fee=0)
```
Solpatium, line 219: What happens on other envs?

starknet_py/tests/e2e/conftest.py
war-in: It is not. In the current test setup, we don't use new_account_client, because transaction version = 1 is not guaranteed to work correctly on devnet

war-in: Working on updating tests to use accounts deployed by us

starknet_py/tests/e2e/contract_interaction/interaction_test.py
war-in: on testnet result is the same, but devnet accepts transactions with max_fee=0

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -49,44 +48,35 @@ async def test_get_deploy_transaction(
...
-            contract_address=contract_address,
-            calldata=invoke_transaction_calldata,
-            entry_point_selector=invoke_transaction_selector,
-            hash=invoke_transaction_hash,
-            signature=[],
-            max_fee=0,
-            version=0,
-            nonce=None,
-        )
+        assert invoke_transaction_calldata[0] in transaction.calldata
```
Solpatium, line 78: It checks just a single value from calldata, right?

Solpatium (changes requested):
Great job, just a few questions :)

starknet_py/tests/e2e/contract_interaction/interaction_test.py
Solpatium: Let's change "when running on integration" to "when running on a real environment"

starknet_py/tests/e2e/client/client_test.py
war-in: Yes, because transaction.calldata has many values and we are interested only in invoke_transaction_calldata[0]

starknet_py/tests/e2e/client/conftest.py
Solpatium: Do we use any specific functions from it? If so we should name it "balance_contract_address".

starknet_py/tests/e2e/conftest.py
Solpatium: Oh I see, so this would be a seperate PR?

war-in: Yes, I think it will be clearer than adding this here

starknet_py/tests/e2e/client/conftest.py
war-in: No, we don't. I've checked that

starknet_py/tests/e2e/client/client_test.py
Solpatium: Why are we only interested in the first one?

war-in: because we pass [1234] as constructor argument and transaction.calldata is something like [0, 5, ..., 1234].
So our argument is here, but there are additional values too

war-in: And we can't explicitly do `assert invoke_transaction_calldata[0] in transaction.calldata`

pyproject.toml
```diff
@@ -42,6 +42,9 @@ pytest-xdist = "^2.5.0"
 web3 = {extras=["tester"], version="6.0.0b4"}
 
 
+[tool.poetry.group.dev.dependencies]
```
cptartur, line 45: If poetry changed pyproject.toml format, perhaps all dev dependencies could be moved to this category and re-locked?

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -49,44 +48,35 @@ async def test_get_deploy_transaction(
...
-        sender_address=DEFAULT_DECLARE_SENDER_ADDRESS,
-        hash=declare_transaction_hash,
-        signature=[],
-        max_fee=0,
-        version=0,
-        nonce=None,
+    sender_address = (
+        gateway_account_client.address
+        if transaction.version == 1
+        else DEFAULT_DECLARE_SENDER_ADDRESS
```
cptartur, line 61: Could this logic easily be moved to a fixture?

starknet_py/tests/e2e/client/client_test.py
cptartur: Are we sure constructor calldata will always be at 0th place? Perhaps `any(1234 in data for data in calldata)` would be a safer check?

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -38,9 +38,15 @@ async def test_node_get_transaction_by_block_id_and_index(
...
-    deploy_transaction_hash, contract_address, rpc_client, class_hash
+    deploy_transaction_hash,
+    block_with_deploy_number,
+    contract_address,
+    rpc_client,
+    class_hash,
 ):
-    tx = await rpc_client.get_transaction_by_block_id(block_number=1, index=0)
+    tx = await rpc_client.get_transaction_by_block_id(
+        block_number=block_with_deploy_number, index=0
```
cptartur, line 48: How are we sure that this transactions will be at index 0 in integration/testnet?

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -15,14 +15,14 @@
...
-    block_with_deploy_hash,
+    block_with_deploy_number,
     deploy_transaction_hash,
     contract_address,
     rpc_client,
     class_hash,
 ):
     tx = await rpc_client.get_transaction_by_block_id(
-        block_hash=block_with_deploy_hash, index=0
+        block_number=block_with_deploy_number, index=0
```
cptartur, line 25: Same here

starknet_py/tests/e2e/conftest.py
```diff
@@ -33,6 +37,20 @@
...
+    "0X126FAB6AE8ACA83E2DD00B92F94F3402397D527798E18DC28D76B7740638D23"
+)
+
+
+@pytest.fixture(scope="module")
+def event_loop():
+    policy = asyncio.get_event_loop_policy()
+    loop = policy.new_event_loop()
+    yield loop
+    loop.close()
```
cptartur, line 52: I guess we could add a comment here that this was added because the problems with async fixtures in case we run into problems with them in the future.

cptartur (commented):
I like the changes introduced by this PR a lot.

One think I'm worried about however is that because of the modifications of the `Client` tests we don't validate whole dataclasses returned by the client. 

I think we should create an issue about researching a solution that allows us to validate if typing in schemas match typing in dataclasses.

starknet_py/tests/e2e/client/full_node_test.py
war-in: Full node tests should not be run in integration/testnet

Solpatium (approved):

PR #358, opened 2022-09-06
Fix result selector on Custom RPC call

ExoMonk:

Bug Fix on Custom RPC call results -> `starknet_call`

Calling `starknet_call` will always result into a TypeError as we check twice "result" key

`starknet_call` is now working.

No

## **Other information**

-

### DIFF
--- a/starknet_py/net/full_node_client.py
+++ b/starknet_py/net/full_node_client.py
@@ -229,7 +229,7 @@ class FullNodeClient(Client):
                 **block_identifier,
             },
         )
-        return [int(i, 16) for i in res["result"]]
+        return [int(i, 16) for i in res]
 
     async def send_transaction(
         self, transaction: InvokeFunction


Solpatium:
What RPC do you use?

ExoMonk:
> What RPC do you use?

-> jsonRPC (Pathfinder API) using FullNodeClient

Solpatium:
Which pathfinder version do you use?

Solpatium:
We unwrap result in http client already, so we do it again in here. It means there is also an issue with tests/devnet. 
@cptartur please investigate

Solpatium:
Tests passed before because devnet returns it this way: https://discord.com/channels/793094838509764618/928219862327754812/1016738446989082624. We'll need to create a PR to fix it.

cptartur:
Devnet issue will be addressed once 0.10 support is released. Merging this PR now however

PR #360, opened 2022-09-09
Get rid of using pre-deployed accounts

war-in:

It updates the tests

Pre-deployed accounts are used while testing on devnet

Accounts are created by us and feed with devnet tokens

No

### DIFF
--- a/starknet_py/tests/e2e/conftest.py
+++ b/starknet_py/tests/e2e/conftest.py
@@ -10,12 +10,15 @@ from pathlib import Path
 
 import pytest
 import pytest_asyncio
+from starkware.crypto.signature.signature import get_random_private_key
 
 from starknet_py.net import KeyPair, AccountClient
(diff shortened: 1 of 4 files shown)


starknet_py/tests/e2e/conftest.py
```diff
@@ -211,13 +260,21 @@ def new_gateway_account_client(new_address_and_private_key, gateway_client):
...
+    scope="module",
+    params=[
+        "gateway_account_client",
+        "new_gateway_account_client",
+        "rpc_account_client",
+    ],
+)
 def account_client(request):
-    # FIXME add rpc client for devnet tests
     return request.getfixturevalue(request.param)
```
cptartur, line 272: Ideally this fixture would automatically include or exclude a rpc client based on network. This can be however addressed at some later time, please create an issue about researching a solution for this.

cptartur (approved):

PR #363, opened 2022-09-12
Docs for signing off-chain messages

drknzz:

Add section in docs for signing off-chain messages
Fix creating typed data dictionary through TypedData TypedDict


starknet_py/tests/e2e/docs/guide/test_sign_offchain_message.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Create an AccountClient instance
+    client = GatewayClient("testnet")
+    account_client = AccountClient(
+        client=client,
+        address="0x7536539dbba2a49ab688a1c86332625f05f660a94908f362d29212e6071432d",
+        chain=StarknetChainId.TESTNET,
+        key_pair=KeyPair.from_private_key(
+            2640601739596882773716581229189050365310622467205517187262055898295170952602
+        ),
```
war-in, line 56: I would not use a real testnet address and private key. This should be set to some dummy value ("0x2137") and then changed to the parametrized one

war-in (changes requested):

starknet_py/tests/e2e/docs/guide/test_sign_offchain_message.py
cptartur: Agreed, this test should be runnable on devnet (and other nets) like all other docs tests.

starknet_py/tests/e2e/docs/guide/test_sign_offchain_message.py
```diff
@@ -0,0 +1,69 @@
...
+        "contents": "Hello, Bob!",
+    }
+
+    # Create a TypedData dictionary
+    typed_data = TypedData(
+        types=types,
+        primaryType="Mail",
+        domain=starknet_domain,
+        message=message,
+    )
```
cptartur, line 46: Could we replace these steps with the creation of a single standard dict to show it is possible to use `sign_message` without explicitly instantiating TypedData typeddict?

Something like
```python
typed_data = {
  "types": {...},
  "primary_type": "Mail"
  "domain": {...},
  "message: {...}
}
```

Current approach suggests it's necessary to create all these objects.

cptartur (changes requested):

war-in (approved):

starknet_py/net/account/account_client.py
```diff
@@ -629,7 +629,7 @@ async def _verify_message_hash(self, msg_hash: int, signature: List[int]) -> boo
             nonce=None,
         )
         try:
-            await self.call_contract(invoke_tx=invoke_tx)
+            await self.call_contract(invoke_tx=invoke_tx, block_hash="latest")
```
cptartur, line 632: Why was this added?

starknet_py/tests/e2e/conftest.py
```diff
@@ -52,6 +52,9 @@
     "0X126FAB6AE8ACA83E2DD00B92F94F3402397D527798E18DC28D76B7740638D23"
 )
 
+directory_with_contracts = Path(os.path.dirname(__file__)) / "mock_contracts_dir"
+mocks_dir = Path(os.path.dirname(__file__)) / "mocks"
```
cptartur, line 56: If you want to name your directory `mocks` perhaps we should consider renaming `mock_contracts_dir` to just `contracts`?

cptartur: cc @war-in

starknet_py/tests/e2e/docs/guide/test_sign_offchain_message.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # We can calculate the message hash
+    msg_hash = account_client.hash_message(typed_data=typed_data)
+
+    # Sign the message
+    signature = account_client.sign_message(typed_data=typed_data)
+
+    # Verify the message
+    assert await account_client.verify_message(
+        typed_data=typed_data, signature=signature
```
cptartur, line 75: I think we shouldn't be using `assert` in these examples in case someone uses this code in production, as asserts can be disabled there.

cc @war-in

war-in: You're right! I would move assert to the end of the file after # add to docs: end

starknet_py/tests/e2e/conftest.py
cptartur: I just think having `mocks/` and `mock_contracts_dir/` is kind of confusing. For me it's fine to have "mock" contracts in `contracts/`, but if you don't like it, we don't have to go with that approach.

starknet_py/net/account/account_client.py
drknzz: Without it, full node client had a problem in rpc_call as devnet does not support other blocks than latest.

cptartur (approved):

PR #366, opened 2022-09-13
Remove some deprecation warnings when running tests

war-in:

Removes some warning while running tests.
Also removes the signature parameter from call

-

-

No


starknet_py/tests/e2e/client/conftest.py
```diff
@@ -26,11 +26,13 @@
...
 
-async def prepare_network(gateway_account_client: AccountClient) -> PreparedNetworkData:
+async def prepare_network(
+    new_gateway_account_client: AccountClient,
+) -> PreparedNetworkData:
     contract_compiled = Path(contracts_dir / "balance_compiled.json").read_text("utf-8")
 
     prepared_data = await prepare_net_for_tests(
-        gateway_account_client, compiled_contract=contract_compiled
+        new_gateway_account_client, compiled_contract=contract_compiled
```
cptartur, line 35: I wonder if this should be changed now, this way all transactions in our client tests are version 1 instead of 0.

war-in: Yes, but we are fetching them only, so version doesn't matter here

cptartur (approved):

starknet_py/net/account/account_client.py
```diff
@@ -619,17 +619,13 @@ async def _verify_message_hash(self, msg_hash: int, signature: List[int]) -> boo
         """
         calldata = [msg_hash, len(signature), *signature]
```
Solpatium, line 620: We can also update `_get_call_payload`

Solpatium (approved):

PR #368, opened 2022-09-14
Integration testing improved

war-in:

We do not have to change anything (in the conftest) to effectively test on integration

We have to remove `"rpc_account_client"` from `account_client` fixture to do not have failing tests while testing on integration

We do not have to do anything to test on integration

No


starknet_py/tests/e2e/conftest.py
```diff
@@ -314,13 +317,21 @@ def new_gateway_account_client(
...
-    scope="module",
-    params=[
+def net_to_accounts() -> List[str]:
+    accounts = [
         "gateway_account_client",
         "new_gateway_account_client",
         "rpc_account_client",
-    ],
+    ]
+    if any(net in sys.argv for net in ["--net=integration", "--net=testnet"]):
```
cptartur, line 326: Were we also supporting writing these arguments as `--net integration` or only with the `=` sign? I'd verify that.

PR #374, opened 2022-09-20
UDC deployer

war-in:

It adds `AccountClient`'s method `deploy_contract` which deploys with `class_hash`

The only way of deployment is through `Deploy` transaction

...

No


cptartur (commented):
Looks fine to me. I'm unsure if relying on address being at specific place in response is a good idea. Perhaps we should consider parsing the received event?

starknet_py/net/account/account_client.py
```diff
@@ -469,6 +477,70 @@ async def execute(
...
+        self,
+        class_hash: Hash,
+        salt: Optional[int] = None,
+        unique: bool = True,
+        abi: Optional[list] = None,
+        constructor_calldata: Optional[Union[List[any], dict]] = None,
+        deployer_address: Optional[AddressRepresentation] = None,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+        version: Optional[int] = None,
```
cptartur, line 490: I'd consider combining some of these parameters into a TypedDict.

starknet_py/tests/e2e/deploy/deploy_test.py
```diff
@@ -106,3 +108,126 @@ async def test_constructor_without_arguments(gateway_account_client):
...
+
+@pytest.mark.asyncio
+async def test_default_deploy_with_class_hash(
+    deployer_address, account_client, map_class_hash
+):
+    res = await account_client.deploy_contract(
+        class_hash=map_class_hash, deployer_address=deployer_address, max_fee=MAX_FEE
+    )
+
+    assert isinstance(res, int)
```
cptartur, line 121: I'd also check if its `!= 0`

starknet_py/utils/contructor_args_translator.py
```diff
@@ -0,0 +1,31 @@
...
+        None,
+    )
+
+    # Constructor might not accept any arguments
+    if not constructor_abi or not constructor_abi["inputs"]:
+        return []
+
+    if not constructor_args:
+        raise ValueError(
+            "Provided contract has a constructor and no args were provided."
```
cptartur, line 20:
```suggestion
            "Provided contract has a constructor and no arguments were provided."
```

starknet_py/tests/e2e/deploy/deploy_test.py
```diff
@@ -106,3 +108,126 @@ async def test_constructor_without_arguments(gateway_account_client):
...
+            await new_gateway_account_client.sign_declare_transaction(
+                compilation_source=constructor_with_arguments_source, max_fee=MAX_FEE
+            )
+        )
+    ).class_hash
+
+    value = 10
+    tuple_value = (1, (2, 3))
+    arr = [1, 2, 3]
+    struct = {"value": 12, "nested_struct": {"value": 99}}
```
cptartur, line 189: Can't we parametrize the test with these?

starknet_py/tests/e2e/deploy/deploy_test.py
```diff
@@ -106,3 +108,126 @@ async def test_constructor_without_arguments(gateway_account_client):
...
+    )
+    contract_2 = Contract(address=contract_2_address, abi=abi, client=account_client)
+
+    assert contract_1.address != contract_2.address
+
+    result_1 = await contract_1.functions["get"].call(block_hash="latest")
+    result_2 = await contract_1.functions["get"].call(block_hash="latest")
+
+    assert result_1 == (value, tuple_value, sum(arr), struct)
+    assert result_2 == (value, tuple_value, sum(arr), struct)
```
cptartur, line 233: This test is quite big. Couldn't it be separated into smaller tests?

starknet_py/net/account/account_client.py
war-in: #375

starknet_py/tests/e2e/deploy/deploy_test.py
war-in: There will be too many arguments in the test function

starknet_py/net/account/account_client.py
```diff
@@ -477,27 +477,25 @@ async def execute(
     async def deploy(self, transaction: Deploy) -> DeployTransactionResponse:
         return await self.client.deploy(transaction=transaction)
 
-    async def deploy_contract(
+    async def deploy_through_udc(
```
cptartur, line 480: I'm not quite sure if that's a good name. I'd prefer if we avoided abbreviations in method names.

cptartur (commented):
Overall the implementation looks fine, im still unsure about including it in AccountClient interface though. I know it makes sense to put it there, but currently using `deploy_through_udc` requires passing a lot of parameters every time you use it and it's not very practical. 

Perhaps another approach should be considered, moving the logic to the separate module and somehow integrating it with `AccountClient`?

starknet_py/net/account/account_client.py
```diff
@@ -469,6 +471,28 @@ async def execute(
...
+        deployer_address: Optional[AddressRepresentation] = None,
+        salt: Optional[int] = None,
+        unique: Optional[bool] = True,
+    ) -> Deployer:
+        """
+        Creates a Deployer instance which is used to deploy contracts through Universal Deployer Contract (UDC)
+
+        :param deployer_address: Address of the UDC. Must be set when using a custom network
+        :param salt: The salt for a contract to be deployed. Random value is selected if it is not provided
+        :param unique: Boolean determining if the salt should be connected with the account's address
```
Solpatium, line 485: Default missing here

starknet_py/net/deployer/contract_deployment.py
```diff
@@ -0,0 +1,89 @@
...
+            calldata=calldata,
+        )
+
+        transaction = await self.deployer.account.sign_invoke_transaction(
+            calls=call, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+
+        return transaction
+
+    async def send_transaction(self, deploy_invoke_transaction: InvokeFunction) -> int:
```
Solpatium, line 70: I think we should accept the same args as with `prepare_transaction` here. Current api is not ergonomic.

Solpatium: Also someone might provide an invalid InvokeFunction tx here.

starknet_py/net/deployer/deployer.py
```diff
@@ -0,0 +1,33 @@
+from typing import Optional, List
```
Solpatium, line 1: We should change the name of the package to `udc_deployer`

starknet_py/utils/contructor_args_translator.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def translate_constructor_args(abi: list, constructor_args: any) -> List[int]:
+    constructor_abi = next(
+        (member for member in abi if member["type"] == "constructor"),
+        None,
+    )
+
+    # Constructor might not accept any arguments
+    if not constructor_abi or not constructor_abi["inputs"]:
```
Solpatium, line 15: So we check two cases here:
1. Contract doesn't have a constructor
2. There is a constructor with no arguments

Right?

starknet_py/net/deployer/contract_deployment.py
Solpatium: Also when we send something we usually don't wait for tx to be accepted as here. This is surprising for developers.

starknet_py/net/account/account_client.py
```diff
@@ -469,6 +471,28 @@ async def execute(
     async def deploy(self, transaction: Deploy) -> DeployTransactionResponse:
         return await self.client.deploy(transaction=transaction)
 
+    def create_deployer(
```
Solpatium, line 474: I think we can remove this method to avoid problems with typing, pyright can't handle string types :|

starknet_py/net/deployer/deployer.py
```diff
@@ -0,0 +1,33 @@
...
+        address: AddressRepresentation,
+        salt: Optional[int] = None,
+        unique: bool = True,
+    ):
+        self.account = account
+        self.address = address
+        self.salt = salt
+        self.unique = unique
+
+    def for_contract(self, class_hash: Hash, abi: Optional[List] = None):
```
Solpatium, line 25: `make_deployment`?

starknet_py/net/deployer/contract_deployment.py
```diff
@@ -0,0 +1,89 @@
...
+)
+
+
+class ContractDeployment:
+    """
+    ContractDeployment used to prepare and send deploy invoke transactions
+    """
+
+    def __init__(
+        self, deployer: "Deployer", class_hash: Hash, abi: Optional[List] = None
```
Solpatium, line 21: Can we do something to avoid string types? Perhaps nesting this class in Deployer? As `Deployer.Deployment`?

Solpatium (changes requested):

starknet_py/net/account/account_client.py
```diff
@@ -469,6 +471,28 @@ async def execute(
...
+    ) -> Deployer:
+        """
+        Creates a Deployer instance which is used to deploy contracts through Universal Deployer Contract (UDC)
+
+        :param deployer_address: Address of the UDC. Must be set when using a custom network
+        :param salt: The salt for a contract to be deployed. Random value is selected if it is not provided
+        :param unique: Boolean determining if the salt should be connected with the account's address
+        :returns: Deployer
+        """
+        deployer_address = deployer_address_from_network(
```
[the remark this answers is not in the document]
war-in: `deployer_address_from_network` works as `chain_from_network` if mainnet/testnet is used it returns fixed value, but for custom net it returns `deployer_address` passed as argument

starknet_py/net/deployer/contract_deployment.py
war-in: Agree with first and second, but third one is tricky. We have to wait for tx being accepted to fetch its receipt. Furthermore, wait_for_tx does not wait for the transaction to be accepted, but only for the PENDING status

starknet_py/utils/contructor_args_translator.py
war-in: Not sure, because it was moved from the Contract to avoid conflicts

starknet_py/net/deployer/contract_deployment.py
war-in: Will try

starknet_py/net/account/account_client.py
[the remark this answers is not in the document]
Solpatium: Yes and I would expect that when I provide an argument it would be used instead of the default.

starknet_py/net/deployer/contract_deployment.py
Solpatium: Can't we expose a method for getting this info when tx is accepted already?

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,146 @@
...
+        :param address: Address of the UDC. Must be set when using a custom network
+        :param salt: The salt for a contract to be deployed. Random value is selected if it is not provided
+        :param unique: Boolean determining if the salt should be connected with the account's address. Default to True
+        """
+        address = (
+            address if isinstance(address, int) or address is None else int(address, 16)
+        )
+        address = address or deployer_address_from_network(
+            net=account.net, deployer_address=address
+        )
```
cptartur, line 43: What if address is None and unique is set to False? Will an exception be raised?

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,146 @@
...
+
+        return event[0]
+
+    async def _get_deploy_event(self, transaction_hash: Hash) -> Optional[Event]:
+        receipt = await self.account.get_transaction_receipt(tx_hash=transaction_hash)
+
+        for event in receipt.events:
+            if get_selector_from_name("ContractDeployed") == event.keys[0]:
+                return event
+        return None
```
cptartur, line 91:
```suggestion
next(filter(lambda event: get_selector_from_name("ContractDeployed") == event.keys[0], receipt.events), None)
```
I'm not sure if this is simpler than the for loop but certainly is 1 line vs 4 😅

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,146 @@
...
+        address = address or deployer_address_from_network(
+            net=account.net, deployer_address=address
+        )
+
+        self.account = account
+        self.address = address
+        self.salt = salt
+        self.unique = unique
+
+    def make_deployment(self, class_hash: Hash, abi: Optional[List] = None):
```
cptartur, line 50: This should have a return type

starknet_py/net/udc_deployer/deployer.py
war-in: Unique is not connected with the address so yes, an exception will be raised

starknet_py/net/udc_deployer/deployer.py
Solpatium: It is very dense and harder to read.

cptartur (approved):
Looks fine to me now. We should add proper docs for this module and a section in guide before merging. 

Also, currently addresses of the UDCs on testnet and mainnet are mock values, and we can't exactly merge it that way: Do we want to merge the initial support now and extend it once UDCs get deployed or are we going to wait with merging this PR until then?

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,154 @@
...
+
+    async def _get_deploy_event(self, transaction_hash: Hash) -> Optional[Event]:
+        receipt = await self.account.get_transaction_receipt(tx_hash=transaction_hash)
+
+        for event in receipt.events:
+            if get_selector_from_name("ContractDeployed") == event.keys[0]:
+                return event
+        return None
+
+    class ContractDeployment:
```
Solpatium, line 96: I am not sure if this is needed now.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,154 @@
...
+            :param class_hash: The class_hash of the contract to be deployed
+            :param abi: ABI of the contract to be deployed
+            """
+            self.deployer = deployer
+            self.class_hash = class_hash
+            self.abi: List = abi or []
+
+        async def prepare_transaction(
+            self,
+            constructor_calldata: Optional[Union[List[any], dict]] = None,
```
Solpatium, line 115: calldata

Solpatium (changes requested):

starknet_py/net/udc_deployer/deployer.py
war-in: Right, an exception would be raised only if the address would be `None` and the account.net would be a custom network

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
```diff
@@ -0,0 +1,96 @@
...
+    # read about them in the API section
+    deployer = Deployer(
+        account=account_client,
+        deployer_address=deployer_address,
+        salt=salt,
+        unique=False,
+    )
+
+    # If contract we want to deploy does not have constructor, or the constructor
+    # does not have arguments, abi does not have to be passed to the `.prepare_contract_deployment` method
```
cptartur, line 50:
```suggestion
    # does not have arguments, abi is not a required parameter of `.prepare_contract_deployment` method
```

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,129 @@
...
+            calldata=calldata,
+        )
+
+        transaction = await self.account.sign_invoke_transaction(
+            calls=call, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+
+        return transaction
+
+    async def find_deployed_contract_address(self, transaction_hash: Hash) -> int:
```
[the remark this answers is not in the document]
Solpatium: I like find better because it suggests searching, which is what we do here.

starknet_py/tests/e2e/conftest.py
```diff
@@ -469,6 +471,112 @@ async def cairo_serializer(gateway_account_client: AccountClient) -> CairoSerial
...
+async def deployer_address(gateway_client: AccountClient) -> int:
+    """
+    Returns an address of the UDC
+    """
+    deploy_tx = make_deploy_tx(
+        compilation_source=(mock_dir / "contracts/universal_deployer.cairo").read_text(
+            "utf-8"
+        )
+    )
+    res = await gateway_client.deploy(deploy_tx)
```
war-in, line 484: For now, it uses an old 'Deploy` transaction, but there is an issue in devnet to create pre-deployed UDC, so we will be able to replace this code.

starknet_py/net/models/deployer_addresses.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def deployer_address_from_network(
+    net: Network, deployer_address: Optional[int] = None
+) -> int:
+    # FIXME: change to real addresses once deployed
+    mapping = {
+        MAINNET: 123,
+        TESTNET: 456,
+    }
```
war-in, line 13: Change when deployers will be deployed on mainnet and testnet

starknet_py/tests/e2e/mock/contracts/universal_deployer.cairo
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.4.0 (utils/presets/UniversalDeployer.cairo)
+
```
drknzz, line 3: I think we should paste here whole oz's license text like I've done it [here](https://github.com/software-mansion/starknet.py/blob/update/compiled-contracts/starknet_py/tests/e2e/mock/contracts/oz_proxy.cairo).

starknet_py/utils/contructor_args_translator.py
```diff
@@ -0,0 +1,33 @@
+# pyright: reportGeneralTypeIssues=false
```
drknzz, line 1: Maybe we could get rid of it?

starknet_py/utils/contructor_args_translator.py
```diff
@@ -0,0 +1,33 @@
...
+from starkware.starknet.public.abi_structs import identifier_manager_from_abi
+
+from starknet_py.utils.data_transformer import FunctionCallSerializer
+
+
+def translate_constructor_args(abi: List, constructor_args: any) -> List[int]:
+    constructor_abi = next(
+        (member for member in abi if member["type"] == "constructor"),
+        None,
+    )
```
drknzz, line 14: Could we make a local variable here

starknet_py/net/models/deployer_addresses.py
```diff
@@ -0,0 +1,23 @@
+from typing import Optional
+
+from starknet_py.net.networks import Network, MAINNET, TESTNET
+
+
+def deployer_address_from_network(
```
drknzz, line 6: I'm not a fan of this function. From what I see it is used only once in the actual implementation and always `None` as the `deployer_address` will be passed. Maybe we could change it somehow

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,128 @@
...
+        :param deployer_address: Address of the UDC. Must be set when using a custom network
+        :param salt: The salt for a contract to be deployed. Random value is selected if it is not provided
+        :param unique: Boolean determining if the salt should be connected with the account's address. Default to True
+        """
+        if deployer_address is None:
+            deployer_address = deployer_address_from_network(
+                net=account.net, deployer_address=deployer_address
+            )
+        else:
+            deployer_address = int_from_hex(deployer_address)
```
drknzz, line 48: Here, the `deployer_address_from_network` is always called with `deployer_address=None`

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,128 @@
...
+        self.deployer_address = deployer_address
+        self.salt = salt
+        self.unique = unique
+
+    async def prepare_contract_deployment(
+        self,
+        *,
+        class_hash: Hash,
+        abi: Optional[List] = None,
+        calldata: Optional[Union[List[any], dict]] = None,
```
drknzz, line 60: The `List[any]` could be just `List`, or `list` maybe, as you've used before

starknet_py/utils/contructor_args_translator.py
```diff
@@ -0,0 +1,33 @@
+# pyright: reportGeneralTypeIssues=false
+
+from typing import List
+
+from starkware.starknet.public.abi_structs import identifier_manager_from_abi
+
+from starknet_py.utils.data_transformer import FunctionCallSerializer
+
+
+def translate_constructor_args(abi: List, constructor_args: any) -> List[int]:
```
drknzz, line 10: Can constructor_args be actually `any`?

starknet_py/tests/e2e/mock/contracts/universal_deployer.cairo
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.4.0 (utils/presets/UniversalDeployer.cairo)
```
drknzz, line 2: We could actually use v0.5.0 (I don't know if anything has changed since v0.4.0, I think not though)

drknzz (changes requested):
Nice job overall 🚀 , left some changes to think about

starknet_py/utils/contructor_args_translator.py
war-in: I'm not sure wdym 😅

starknet_py/tests/e2e/mock/contracts/universal_deployer.cairo
war-in: why not

starknet_py/utils/contructor_args_translator.py
drknzz: I meant to assign `(member for member in abi if member["type"] == "constructor")` to local variable, and then use it here; sorry, should've been more specific 🙈

drknzz (approved):

starknet_py/net/models/deployer_addresses.py
war-in: I don't think that this function is useful anymore. There is an issue in devnet to change the pedeployed deployer address to `0x041a78e741e5aF2fEc34B695679bC6891742439f7AFB8484Ecd7766661aD02BF` to match testnet
https://github.com/Shard-Labs/starknet-devnet/issues/326

cptartur: Hmm, but you still need to resolve the address for mainnet/testnet right?

war-in: nope, there is a link in the devnet issue, please check it

cptartur: Let's remove it then

docs/guide.rst
```diff
@@ -92,6 +92,22 @@ Here's how you can deploy new contracts:
     :dedent: 4
 
 
+Deploying new contracts with Universal Deployer Contract (UDC)
```
Solpatium, line 95: I think we should create a section about deploying and show this and deploy_account as 2 possible paths.

Solpatium: Can be done as a separate PR.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+
+        if not abi and calldata:
+            raise ValueError("calldata was provided without an abi")
+
+        calldata = translate_constructor_args(abi=abi or [], constructor_args=calldata)
+
+        calldata, _ = universal_deployer_serializer.from_python(
+            value_types=deploy_contract_abi["inputs"],
+            classHash=int_from_hex(class_hash),
```
Solpatium, line 77: I think we usually require ints right away, right?

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,121 @@
...
+        # Ignore typing, because event is a NamedTuple
+        return event.address  # pyright: ignore
+
+    async def _get_deploy_event(self, transaction_hash: Hash) -> Optional[Event]:
+        receipt = await self.account.get_transaction_receipt(tx_hash=transaction_hash)
+
+        for event in receipt.events:
+            if get_selector_from_name("ContractDeployed") == event.keys[0]:
+                return event
+        return None
```
Solpatium, line 121: What if there are 2 events like this in the tx?

starknet_py/tests/e2e/conftest.py
```diff
@@ -469,6 +471,113 @@ async def cairo_serializer(gateway_account_client: AccountClient) -> CairoSerial
     return CairoSerializer(identifier_manager=contract.data.identifier_manager)
 
 
+@pytest_asyncio.fixture(scope="module")
+async def deployer_address(gateway_client: AccountClient) -> int:
```
Solpatium, line 475: Do we still need this? https://github.com/Shard-Labs/starknet-devnet/pull/324

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
```diff
@@ -0,0 +1,95 @@
...
+    # add to docs: end
+    testnet_account_client = AccountClient(
+        address=123,
+        client=GatewayClient(net="testnet"),
+        key_pair=KeyPair.from_private_key(123),
+        chain=StarknetChainId.TESTNET,
+    )
+
+    # add to docs: start
+    # If the account is deployed on testnet or mainnet there is no need to explicitly specify
```
Solpatium, line 29: "If you use mainnet/testnet/devnet there is no need..."

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
```diff
@@ -0,0 +1,95 @@
...
+    # address of the deployer (default one will be used)
+    deployer = Deployer(account=testnet_account_client)
+
+    # If custom net is used (e.g. starknet-devnet) address of the deployer contract is required
```
Solpatium, line 33: devnet supports it as well now

Solpatium (approved):
LGTM with comments

starknet_py/tests/e2e/conftest.py
war-in: Only for a while. They are deploying it to the wrong address https://github.com/Shard-Labs/starknet-devnet/issues/326

Solpatium: Hmmmm do you know if they plan to change that?

war-in: I hope so, but I don't understand the last message from the linked issue 😅

Solpatium: I think we might want to wait for devnet to avoid confusion 🤔 Please contact them to see what's the plan.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -64,20 +64,52 @@ async def prepare_contract_deployment(
         :param calldata: Constructor args of the contract to be deployed
         :param max_fee: Max amount of Wei to be paid when executing transaction
         :param auto_estimate: Use automatic fee estimation, not recommend as it may lead to high costs
-        :return: InvokeFunction
+        :return: NamedTuple with invoke_tx and address fields
```
cptartur, line 67: Typing of this function should be updated

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -86,36 +118,15 @@ async def prepare_contract_deployment(
             calldata=calldata,
         )
 
-        transaction = await self.account.sign_invoke_transaction(
-            calls=call, max_fee=max_fee, auto_estimate=auto_estimate
+        deployer_address = self.deployer_address if self.unique else 0
+        salt = pedersen_hash(self.account.address, salt) if self.unique else salt
```
cptartur, line 122: Why is hash calculated here?

starknet_py/net/udc_deployer/deployer.py
cptartur, line 67: Maybe let's make a docstring a little bit more descriptive, so it says it's the address of a contract that will be deployed or something like that.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -64,20 +64,52 @@ async def prepare_contract_deployment(
...
         """
+        call, address = await self.create_deployment_call(
+            class_hash=class_hash, salt=salt, abi=abi, calldata=calldata
+        )
+
+        transaction = await self.account.sign_invoke_transaction(
+            calls=call, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+
+        return namedtuple(
```
cptartur, line 77: Let's use a NamedTuple from [typing](https://docs.python.org/3/library/typing.html#typing.NamedTuple) so we can have better typing. 

We can define the tuples outside methods and then use them for both type annotations and creating actual tuples.

cptartur (changes requested):

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
```diff
@@ -0,0 +1,93 @@
...
+        account=account_client,
+        deployer_address=deployer_address,
+        unique=False,
+    )
+
+    # If contract we want to deploy does not have constructor, or the constructor
+    # does not have arguments, abi is not a required parameter of `deployer.prepare_contract_deployment` method
+    deploy_invoke_transaction = await deployer.prepare_contract_deployment(
+        class_hash=map_class_hash, salt=salt, max_fee=int(1e16)
+    )
```
cptartur, line 52:
```suggestion
    deploy_invoke_transaction, address = await deployer.prepare_contract_deployment(
        class_hash=map_class_hash, salt=salt, max_fee=int(1e16)
    )
```

starknet_py/net/udc_deployer/deployer.py
Solpatium, line 67: @war-in @cptartur what if we didn't expose `prepare_contract_deployment` here at all and just asked users to provide the call they get here to the execute or sign_invoke_transaction? I think prepare_contract_deployment is pretty useless and duplicates functionality from account. 
It just feels like `create_deployment_call` is the right abstraction here.

Solpatium: We could change create_deployment_call to accept an optional `Account` and would use its address.

Solpatium: The benefit is that it emphasised that it is just a normal call inside invoke, it becomes obvious how you could atomically do some things after deploying the contract.

cptartur: I don't think it's practical to do so: we're generally using transactions in other parts of the library. As far as i understand, creating calls was only added to the udc for the users that want to create and interact with the account in the same tx - i'd say not the most standard flow.

Solpatium: But see what prepare_contract_deployment does now: takes output of `create_deployment_call` and calls `sign_invoke_transaction`. It doesn't do much.

In case user wants to execute the deploy the flow will be the same as now:
```
deploy_call = udc.create_deployment_call(...) # now it is sign_invoke_transaction
account.execute(deploy_call)
```

In case user doesn't want to execute it right away there is just one more function call to do.
```
deploy_call = udc.create_deployment_call(...)
tx = account.sign_invoke_transaction(deploy_call)
```

I really like better composability and emphasis of treating UDC deploy just like any regular call. By hiding this call in tx users won't realise the possibilities. If it is not ergonomic for users then we can add it in the future without any issue.
Also having `auto_estimate` in UDC as a parameter feels like leaking logic from account.

starknet_py/net/udc_deployer/deployer.py
war-in: Because it is calculated inside the UDC contract. To get the same address I have to do the same transformations

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,133 @@
...
+            typename="InvokeAndAddress", field_names=["invoke_tx", "address"]
+        )(transaction, address)
+
+    async def create_deployment_call(
+        self,
+        *,
+        class_hash: Hash,
+        salt: Optional[int] = None,
+        abi: Optional[List] = None,
+        calldata: Optional[Union[List, dict]] = None,
```
cptartur, line 90:
```suggestion
        self,
        class_hash: Hash,
        *,
        salt: Optional[int] = None,
        abi: Optional[List] = None,
        calldata: Optional[Union[List, dict]] = None,
```

I think we can still allow passing class_hash as a positional argument as it is always required: I think it's more usable that way.

starknet_py/net/udc_deployer/deployer.py
cptartur: Okay, we can do what you suggest and merge.

However I still think we need simplified flow for less advanced users, as the current one is quite a lot more complicated that just `Client.deploy` as it was previously.

Let's create an issue for that.

starknet_py/net/udc_deployer/deployer.py
Solpatium: I see your point, but I am not sure if it is worth it. My fear is that they might change class_hash to something else that we cannot predict and it is much easier to deprecate the old one with keyword arguments. We had a situation with transaction parameter somewhere that we couldn't imagine being changed and yet it happened. On the other hand I don't know what else could be used here in the future 🤔 
@war-in what do you think?

cptartur: I understand the issue with deprecation and it makes sense: however how likely is that the class hash will change to something else at this point?

war-in: Changing class_hash to something else is quite a big issue. I think we can change it to the positional parameter

Solpatium: I am fine with that

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,97 @@
...
+class Deployer:
+    """
+    Deployer used to deploy contracts through Universal Deployer Contract (UDC)
+    """
+
+    def __init__(
+        self,
+        *,
+        deployer_address: AddressRepresentation = DEFAULT_DEPLOYER_ADDRESS,
+        account_address: Optional[AddressRepresentation] = None,
```
Solpatium, line 34: Maybe we should move it to `create_deployment_call`?

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,97 @@
...
+        :param deployer_address: Address of the UDC. Is set to the address of
+            the default UDC (same address on real nets and devnet) by default.
+            Must be set when using custom network other than devnet.
+        :param account_address: Address of an account which will send the transaction.
+            If passed, address of the contract to be deployed will be connected with it.
+        """
+
+        self.deployer_address = parse_address(deployer_address)
+        self.account_address = account_address
+        self.unique = account_address is not None
```
Solpatium, line 46: We don't need this as a property here

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
```diff
@@ -0,0 +1,92 @@
...
+    testnet_account_client = AccountClient(
+        address=123,
+        client=GatewayClient(net="testnet"),
+        key_pair=KeyPair.from_private_key(123),
+        chain=StarknetChainId.TESTNET,
+    )
+
+    # docs: start
+    # If you use mainnet/testnet/devnet there is no need to explicitly specify
+    # address of the deployer (default one will be used)
```
Solpatium, line 30: We should wait for release with devnet fix before we release these docs.

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
```diff
@@ -0,0 +1,92 @@
...
+            "dict": {"value": 12, "nested_struct": {"value": 99}},
+        },
+    )
+
+    # Once call is prepared it can be signed and send with an Account
+    invoke_tx = await account_client.sign_invoke_transaction(
+        deploy_call, max_fee=int(1e16)
+    )
+    resp = await account_client.send_transaction(transaction=invoke_tx)
+    await account_client.wait_for_tx(resp.transaction_hash)
```
Solpatium, line 86: Let's show flow with execute first, before the signing flow and emphasise it is a regular call.

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # After waiting for a transaction
```
Solpatium, line 88: I think it would be great to show that you can create a multicall with invoke. A good example would be a contract that is configured after deployment.

Solpatium (changes requested):

starknet_py/net/udc_deployer/deployer.py
cptartur: It depends if we want the UDC instance to be reusable for different account addresses or created per address: I think in standard use case, making all deployments from the same account is more probable.

war-in: I agree with @cptartur

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
war-in: Devnet just merged mentioned fix and we are using it now

starknet_py/net/udc_deployer/deployer.py
Arcticae: It's a bit weird from user perspective, we should be able to create the UDC with a minimal set of required properties. In this case `account_address` is a separate concept from a UDC.

Arcticae (commented):
Please make an issue for that, so we know in advance of ABI/address changes

starknet_py/utils/data_transformer/universal_deployer_serializer.py
```diff
@@ -0,0 +1,39 @@
+from starkware.starknet.public.abi_structs import identifier_manager_from_abi
+
+from starknet_py.utils.data_transformer.data_transformer import CairoSerializer
+
+universal_deployer_abi = [
```
Arcticae, line 5: ABI correctness should be checked, in conjunction with the addresses being correct (we should check if the contract is there, and has expected ABI). 
It can be either performed in a separate action which could be executed daily, or in test suite.

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
Arcticae: @war-in plz make an issue from that

docs/guide.rst
```diff
@@ -87,6 +87,21 @@ Here's how you can deploy new contracts:
...
+--------------------------------------------------------------
+
+The best way of deploying contracts on Starknet is to use UDC. starknet.py assumes you use an implementation compatible
```
Arcticae, line 93: This statement is not true

docs/guide.rst
```diff
@@ -87,6 +87,21 @@ Here's how you can deploy new contracts:
...
+with `OpenZeppelin's UDC implementation <https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/utils/presets/UniversalDeployer.cairo>`_.
```
Arcticae, line 94: I think providing the interface would be more appropriate here, since it should be 1 function only, right? And if the contract shares that method, it'll be compatible. It doesn't even have to be a UDC

starknet_py/net/udc_deployer/deployer.py
cptartur: `account_address` can be used (with unique=True) to influence the address of the deployed contract, I though standard use case of the udc class instance would be configuring this parameter once and then making different deployments. 

We can move it to `create_deployment_call` if you think it makes more sense there, but for me it'd be more ergonomic.

docs/guide.rst
cptartur: Maybe it's most "universal" as it doesn't require any special methods in the contract unlike deploy_account.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,101 @@
...
+        *,
+        deployer_address: AddressRepresentation = DEFAULT_DEPLOYER_ADDRESS,
+        account_address: Optional[AddressRepresentation] = None,
+    ):
+        """
+        :param deployer_address: Address of the UDC. Is set to the address of
+            the default UDC (same address on real nets and devnet) by default.
+            Must be set when using custom network other than devnet.
+        :param account_address: Address of an account which will send the transaction.
+            If passed, address of the contract to be deployed will be connected with it.
```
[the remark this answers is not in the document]
Arcticae: Docs should explain how the parameter is used, now it says it's "connected" which makes the explanation a bit cryptic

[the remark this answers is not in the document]
Arcticae: And "Address of an account which will send the transaction" is not completely true, since we can pass a different address and it will still work - right?

docs/guide.rst
Arcticae: Still, i think we should not make those kind of statements in the docs - syscall is the only supported way and this is a wrapper around that

starknet_py/net/udc_deployer/deployer.py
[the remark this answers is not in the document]
cptartur: It will work in a sense that the transaction will pass, but the address returned by the UDC will be different than that of the deployed contract: you'd then have to search for it through a block explorer or something.

Though I agree explanation could be improved a bit.

docs/guide.rst
cptartur: Syscall and deploy_account transaction - it can deploy any contract not only accounts. It's not very practical though.

Solpatium: We can say that using UDC is a way of deploying contracts if you already have an account.

starknet_py/net/udc_deployer/deployer.py
Solpatium: My logic is that we want to expose logic of function from UDC here, so we can expose same parameters (in our case unique becomes account_address).

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,101 @@
...
+            pedersen_hash(parse_address(self.account_address), salt)
+            if self.account_address is not None
+            else salt
+        )
+        address = compute_address(
+            class_hash=class_hash,
+            constructor_calldata=constructor_calldata,
+            salt=salt,
+            deployer_address=deployer_address,
+        )
```
Solpatium, line 99: Let's move it to a separate function and add more tests to it to make sure it works properly.

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
Solpatium: Yes, but is it released? Users don't use dependency from git.

Solpatium (changes requested):
Please add more tests for calculating the expected address 🙏 
We can use cairo function to compare results

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
war-in: https://github.com/software-mansion/starknet.py/issues/500

starknet_py/tests/e2e/deploy/deployer_test.py
```diff
@@ -96,3 +100,43 @@ async def test_constructor_arguments_contract_deploy(
...
+
+    assert isinstance(class_hash, int)
+    assert class_hash != 0
+
+
+@pytest.mark.asyncio
+async def test_if_udc_has_expected_abi(gateway_client):
+    code = await gateway_client.get_code(contract_address=DEFAULT_DEPLOYER_ADDRESS)
+
+    assert code.abi == universal_deployer_abi
```
Solpatium, line 119: This is not a test of our codebase, but rather assertion of the environment. I am sure pytest has something like `before_module` to run it there.

war-in: https://github.com/software-mansion/starknet.py/pull/374/commits/93401336b78f2abb712d8875bbcec1c9af31dcd3

starknet_py/net/udc_deployer/deployer.py
war-in: Still, I think it should be as reusable as possible. In most cases, users will be using this with only one account and passing the same `account_address` will be annoying

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
war-in: Not yet. I've asked Shard-Labs and I'm waiting for the response

starknet_py/net/udc_deployer/deployer.py
Solpatium: I don't think this will be very annoying. It is not something that you do often and if you do you can always create a wrapper for it.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,106 @@
...
+            Must be set when using custom network other than devnet.
+        :param account_address: Should be equal to the address of the account which will send the transaction.
+            If passed, it will be used to modify the salt, otherwise, salt will not be affected.
+        """
+
+        self.deployer_address = parse_address(deployer_address)
+        self.account_address = account_address
+        self._unique = account_address is not None
+
+    async def create_deployment_call(
```
Solpatium, line 48: I think this doesn't have to be asynchronous anymore, right?

starknet_py/tests/e2e/docs/guide/test_deploying_with_udc.py
Solpatium: I think we can merge it and wait with release until they release.

Solpatium (approved):

starknet_py/net/udc_deployer/deployer.py
cptartur: You can also argue that if someone (for whatever reason) want's to reuse the same UDC instance for different addresses they can do `udc.account_address = new_address`.

If we make this module too directly modelled after UDC interface what will be the benefit of using it instead of just `Contract` aside from address calculation.

cptartur: I don't really se the usacase where someone want's to make multiple deploy calls through the same udc but with different addresses - they'd have to then go through different AccountClients anyway.

Solpatium: >You can also argue that if someone (for whatever reason) want's to reuse the same UDC instance for different addresses they can do udc.account_address = new_address.

No, never do that 😆 

Ok, I don't have a strong opinion on this. It can stay in constructor.

cptartur: I don't recommend doing that, but if it's available as a public parameter anyways and for performance (or whatever) reasons someone cannot create multiple UDC objects, they might as well do that.

Solpatium: No, just never do that :)

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        Creates deployment call to the UDC contract
+
+        :param class_hash: The class_hash of the contract to be deployed
+        :param salt: The salt for a contract to be deployed. Random value is selected if it is not provided
+        :param abi: ABI of the contract to be deployed
+        :param calldata: Constructor args of the contract to be deployed
+        :return: NamedTuple with call and address of the contract to be deployed
+        """
+        if not abi and calldata:
```
Arcticae, line 65: Please provide a way to deploy with plain cairo calldata.

Arcticae (changes requested):

Arcticae (approved):

PR #382, opened 2022-09-21
Add pyright

drknzz:


pyproject.toml
```diff
@@ -90,7 +92,7 @@ build-backend = "setuptools.build_meta"
 
 
 [tool.black]
-target-version = ["py37"]
+target-version = ["py38"]
```
cptartur, line 95: Good catch 👍

pyproject.toml
```diff
@@ -100,3 +102,11 @@ markers = [
     "run_on_testnet: marks test that will only run on testnet (when --net=testnet)",
     "run_on_devnet: marks test that will only run on devnet (when --net=devnet)"
 ]
+
+[tool.pyright]
+include = ["starknet_py/"]
+exclude = [
+    "**/__pycache__",
+    "starknet_py/tests",
+    "**/*_test.py",
```
cptartur, line 111: Why aren't we typechecking tests?

starknet_py/net/client_models.py
```diff
@@ -178,7 +178,7 @@ class DeclareTransactionResponse(SentTransactionResponse):
     Dataclass representing a result of declaring a contract on starknet
     """
 
-    class_hash: int = None
```
cptartur, line 181: I wonder if we should be having default values here? Perhaps this should be resolved on schema level instead?

starknet_py/net/client_models.py
```diff
@@ -187,7 +187,7 @@ class DeployTransactionResponse(SentTransactionResponse):
     Dataclass representing a result of deploying a contract to starknet
     """
 
-    contract_address: int = None
```
cptartur, line 190: Same here

starknet_py/net/gateway_client.py
```diff
@@ -96,7 +98,7 @@ async def get_block(
         res = await self._feeder_gateway_client.call(
             method_name="get_block", params=block_identifier
         )
-        return StarknetBlockSchema().load(res, unknown=EXCLUDE)
+        return StarknetBlockSchema().load(res, unknown=EXCLUDE)  # pyright: ignore
```
cptartur, line 101: Why are these ignores needed? Perhaps something like 
```py
wrapped_res: StarknetBlock = StarknetBlockSchema().load(res, unknown=EXCLUDE)
return wrapped_res
```
or a usage of `typing.cast` would be a better solution?

starknet_py/net/http_client.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, url, session: Optional[aiohttp.ClientSession] = None):
     def http_session(self) -> ClientSession:
         if self.session is not None:
             # noinspection PyTypeChecker
-            return nullcontext(self.session)
+            return nullcontext(self.session)  # pyright: ignore
```
cptartur, line 25: We should add a issue about covering this line with tests

starknet_py/net/http_client.py
```diff
@@ -37,8 +37,8 @@ async def request(
                 session=session,
                 address=address,
                 http_method=http_method,
-                params=params,
-                payload=payload,
+                params=params,  # pyright: ignore
+                payload=payload,  # pyright: ignore
```
cptartur, line 41: Why are these needed here?

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -255,16 +278,20 @@ def _to_python_named(self, cairo_type, name, values):
...
             cairo_type.typ,
             cairo_type.name or "",
             value,
         )
         return value
 
-    def to_python(self, cairo_type: UsedCairoType, name: str, values: CairoData):
+    def to_python(
+        self, cairo_type: TypeTuple.Item, name: str, values: CairoData
+    ) -> PythonResult[Any]:
```
cptartur, line 294: Couldn't some more specific typing be used here instead of Any?

starknet_py/net/l1/contracts.py
```diff
@@ -36,7 +37,7 @@ def __init__(
         )
 
     def l2_to_l1_messages(
-        self, msg_hash: bytes, block_number: EthBlockIdentifier = None
+        self, msg_hash: bytes, block_number: typing.Optional[EthBlockIdentifier] = None
```
cptartur, line 40: Why not import `Optional` and use it without `typing.`?

war-in (approved):

cptartur (changes requested):

pyproject.toml
Solpatium: There is no point in doing this currently.

starknet_py/net/client_models.py
war-in: SentTransactionResponse has default value so DeclareTransactionResponse must have it too

pyproject.toml
cptartur: This should be mentioned in separate issue or in #390 then.

starknet_py/net/gateway_client.py
Solpatium: It's more code, I am ok with having ignores here.

starknet_py/net/http_client.py
Solpatium: Becuse they don't use None as accepted type even though they default to it below.

starknet_py/net/client_models.py
cptartur: Could't it be changed? I'm not 100% sure what's the point of having keeping defaults in dataclasses instead of schemas.

starknet_py/net/l1/contracts.py
Solpatium: This module used typing as module before, I didn't want to change more lines than necessary.

starknet_py/net/gateway_client.py
cptartur: This doesn't seem ideal, but i guess we could keep it. However `typing.cast` doesn't add that much code imo.

starknet_py/net/l1/contracts.py
cptartur: I'd change it anyways, it'd be more in line with what are we doing in pretty much every other file in the package.

starknet_py/net/client_models.py
Solpatium: image

pyproject.toml
Solpatium: @drknzz 👍

starknet_py/net/client_models.py
Solpatium: You can change this in a seperate PR if you want. This PR has a different purpose.

pyproject.toml
Solpatium: I am not sure if we need to typecheck tests at all tbh

starknet_py/net/http_client.py
Solpatium: https://github.com/software-mansion/starknet.py/issues/391

starknet_py/net/client_models.py
Solpatium: Let's discuss this in thread above.

pyproject.toml
Solpatium: I created an issue, but added a low priority https://github.com/software-mansion/starknet.py/issues/392

starknet_py/net/client_models.py
cptartur: Okay, this can be resolved in a separate PR

cptartur (approved):

PR #393, opened 2022-09-23
Fix type errors

Solpatium:


starknet_py/net/full_node_client.py
```diff
@@ -428,14 +443,16 @@ def get_block_identifier(
...
 
+    call = cast(Call, tx)
     return {
-        "contract_address": convert_to_felt(tx.to_addr),
-        "entry_point_selector": convert_to_felt(tx.selector),
-        "calldata": [convert_to_felt(i) for i in tx.calldata],
+        "contract_address": convert_to_felt(call.to_addr),
+        "entry_point_selector": convert_to_felt(call.selector),
+        "calldata": [convert_to_felt(i) for i in call.calldata],
     }
```
cptartur, line 458: These cast seem quite redundant to me. Isn't it possible to make pyright somehow work with isinstance calls?

starknet_py/net/schemas/gateway.py
```diff
@@ -298,7 +296,7 @@ def make_dataclass(self, data, **kwargs):
         return BlockStateUpdate(
             **data,
             storage_diffs=storage_diffs,
-            deployed_contracts=deployed_contracts,
+            deployed_contracts=deployed_contracts,  # pyright: ignore
```
cptartur, line 299: Why was this ignore added?

cptartur (changes requested):

starknet_py/net/full_node_client.py
Solpatium: This works for some reason:
```
    if isinstance(tx, InvokeFunction):
        invoke = cast(InvokeFunction, tx)
        return {
            "contract_address": convert_to_felt(invoke.contract_address),
            "entry_point_selector": convert_to_felt(invoke.entry_point_selector),
            "calldata": [convert_to_felt(i) for i in invoke.calldata],
        }
    return {
        "contract_address": convert_to_felt(tx.to_addr),
        "entry_point_selector": convert_to_felt(tx.selector),
        "calldata": [convert_to_felt(i) for i in tx.calldata],
    }
```

This **does not**:
```
    if isinstance(tx, InvokeFunction):
        return {
            "contract_address": convert_to_felt(tx.contract_address),
            "entry_point_selector": convert_to_felt(tx.entry_point_selector),
            "calldata": [convert_to_felt(i) for i in tx.calldata],
        }
    return {
        "contract_address": convert_to_felt(tx.to_addr),
        "entry_point_selector": convert_to_felt(tx.selector),
        "calldata": [convert_to_felt(i) for i in tx.calldata],
    }
```

>  /Users/jakubptak/work/starknet/starknet_python_sdk/starknet_py/net/full_node_client.py:447:52 - error: Cannot access member "contract_address" for type "Call"
    Member "contract_address" is unknown (reportGeneralTypeIssues)
>  /Users/jakubptak/work/starknet/starknet_python_sdk/starknet_py/net/full_node_client.py:448:56 - error: Cannot access member "entry_point_selector" for type "Call"
    Member "entry_point_selector" is unknown (reportGeneralTypeIssues)

starknet_py/common.py
```diff
@@ -19,6 +17,7 @@ def create_compiled_contract(
     if not compiled_contract and not compilation_source:
```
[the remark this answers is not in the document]
Solpatium: No, because it is `and`. One of them can and will be null. pyright can't do logical deduction. Changed it to allow it understand it.

starknet_py/net/schemas/gateway.py
Solpatium: Changed it above this

starknet_py/net/client.py
```diff
@@ -153,7 +151,7 @@ async def wait_for_tx(
                     TransactionStatus.ACCEPTED_ON_L1,
                     TransactionStatus.ACCEPTED_ON_L2,
                 ):
-                    return result.block_number, status
+                    return cast(int, result.block_number), status
```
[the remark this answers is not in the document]
Solpatium: I added an assertion as a sanity check. I don't want to add more code validating this because having a block_number is an invariant there. I think using assertions in cases like this is a perfect fit.

starknet_py/net/client.py
```diff
@@ -163,15 +161,15 @@ async def wait_for_tx(
                             return result.block_number, status
                 elif status == TransactionStatus.REJECTED:
                     raise TransactionRejectedError(
-                        message=result.rejection_reason,
+                        message=result.rejection_reason or "No message.",
```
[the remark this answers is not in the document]
Solpatium: Changed

starknet_py/net/client.py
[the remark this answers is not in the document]
cptartur: I'm not 100% sure about using `assert` in production code, but if that resolves the issue, I think we can keep it.

starknet_py/net/full_node_client.py
cptartur: Well that's weird, but I guess we cannot do anything about it.

starknet_py/net/client.py
```diff
@@ -171,7 +170,7 @@ async def wait_for_tx(
                 elif status != TransactionStatus.RECEIVED:
                     # This will never get executed with current possible transactions statuses
                     raise TransactionFailedError(
-                        message=result.rejection_reason,
+                        message=cast(str, result.rejection_reason),
```
cptartur, line 173: Perhaps we should add a "getter" method for transaction rejection reason that always returns not None value, based on the transaction status? This way we can use it without this kind of type casting?

starknet_py/transaction_exceptions.py
```diff
@@ -1,13 +1,16 @@
...
     Base exception for transaction failure
     """
 
     def __init__(
         self,
-        message: str = "Unknown starknet error",
+        message: Optional[str] = None,
     ):
-        self.message = message
+        self.message = cast(str, message or "Unknown starknet error")
```
cptartur, line 13:
```suggestion
        if message is None:
            message = "Unknown starknet error"
        self.message = message
```

Solpatium: Why? If this is an empty string then it might as well be turned into "Unknown starknet error"

cptartur: To make pyright happy without having to do typecasting. I didn't consider the fact that we'd also like to replace `""` with default message.

PR #395, opened 2022-09-26
Fix remaining types

Solpatium:
Only `typed_data.py` has global `reportGeneralTypeIssues` enabled now. It will be addressed after logic fixes.

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -9,7 +9,8 @@ from typing import (
     TypeVar,
     Union,
     Dict,
-    NamedTuple, Any,
+    NamedTuple,
+    Any,
 )
 from typing import TypedDict
 
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -98,7 +98,7 @@ class SentTransaction:
 @add_sync_methods
 @dataclass(frozen=True)
 class InvokeResult(SentTransaction):
-    # We ensure these not None in __post_init__
+    # We ensure these are not None in __post_init__
     contract: ContractData = None  # pyright: ignore
     invoke_transaction: InvokeFunction = None  # pyright: ignore
 
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -1,5 +1,3 @@
-# pyright: reportGeneralTypeIssues=false, reportOptionalMemberAccess=false
-
 from __future__ import annotations
 
 import dataclasses
@@ -11,7 +9,7 @@ from typing import (
     TypeVar,
     Union,
     Dict,
-    NamedTuple,
+    NamedTuple, Any,
 )
 from typing import TypedDict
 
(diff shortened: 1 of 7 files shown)


starknet_py/contract.py
```diff
@@ -99,8 +98,9 @@ async def wait_for_acceptance(
 @add_sync_methods
 @dataclass(frozen=True)
 class InvokeResult(SentTransaction):
-    contract: ContractData = None
-    invoke_transaction: InvokeFunction = None
+    # We ensure these not None in __post_init__
```
cptartur, line 101:
```suggestion
    # We ensure these are not None in __post_init__
```

starknet_py/contract.py
```diff
@@ -99,8 +98,9 @@ async def wait_for_acceptance(
...
+    contract: ContractData = None  # pyright: ignore
+    invoke_transaction: InvokeFunction = None  # pyright: ignore
```
cptartur, line 103: This is not validated in `__post_init__`

starknet_py/contract.py
```diff
@@ -130,7 +131,7 @@ def __init__(
         client: Client,
         payload_transformer: FunctionCallSerializer,
         contract_data: ContractData,
-        max_fee: int,
+        max_fee: Optional[int],
```
cptartur, line 134: Why was this changed?

starknet_py/net/gateway_client.py
```diff
@@ -396,14 +394,14 @@ def get_block_identifier(
...
+            "entry_point_selector": hex(tx.selector),
             "calldata": [str(i) for i in tx.calldata],
         }
     return {
-        "contract_address": hex(tx.to_addr),
-        "entry_point_selector": hex(tx.selector),
+        "contract_address": hex(tx.contract_address),
+        "entry_point_selector": hex(tx.entry_point_selector),
         "calldata": [str(i) for i in tx.calldata],
     }
```
cptartur, line 407: I wonder why it only works in that order. 🤔

starknet_py/utils/iterable.py
```diff
@@ -0,0 +1,12 @@
+from typing import TypeVar, Iterable, Union
+
+T = TypeVar("T")
+
+# pyright: reportGeneralTypeIssues=false
+def ensure_iterable(value: Union[T, Iterable[T]]) -> Iterable[T]:
```
cptartur, line 6: `Union[T, Iterable[T]]` it would make more sense to type it as `Any` because it is basically that.

cptartur (changes requested):

starknet_py/contract.py
Solpatium: Because in fact it accepts it.

starknet_py/utils/iterable.py
Solpatium: How would you type the return value without losing type in that case without providing generic type directly?

starknet_py/contract.py
Solpatium: Yes is is
```
    def __post_init__(self):
        assert self.contract is not None
        assert self.invoke_transaction is not None
```

starknet_py/contract.py
Solpatium: And code depends on it

starknet_py/utils/iterable.py
cptartur: I though maybe we could still use `Iterable[T]` as return type. In that case, we can keep it as it is.

cptartur (approved):

starknet_py/utils/iterable.py
Solpatium: Python would have no way of knowing what the T should be there

war-in (approved):

PR #396, opened 2022-09-26
Add custom exceptions for CairoSerializer

### LINKED ISSUE #324: Wrap all the exceptions raised by StarkNet.py with custom ones.
Currently, StarkNet.py uses generic errors throughout the code.

This behaviour is not ideal because:
-  it doesn't allow for catching errors from this package separately,
- it doesn't differentiate between unexpected and expected errors.

The new exceptions could be introduced as a non-breaking change by using inheritance (possibly multiple inheritance).

### LINKED ISSUE #385: Document exceptions raised by methods

Where possible, we should include `::raises Exception` to method docstrings so it's explicit what exceptions are being raised where.

as above

none

**Additional context**

n/a

war-in:

Closes #324 
Closes #385 

CairoSerializer raises ValueError and TypeError

CairoSerializer raises CairoSerializerException and InvalidDataException

No

### DIFF
--- a/starknet_py/utils/data_transformer/data_transformer.py
+++ b/starknet_py/utils/data_transformer/data_transformer.py
@@ -38,8 +38,8 @@ from starknet_py.cairo.felt import (
     encode_shortstring,
 )
 from starknet_py.utils.data_transformer.errors import (
-    CairoSerializerException,
-    InvalidDataException,
+    InvalidValueException,
+    InvalidTypeException,
 )
 
 ABIFunctionEntry = Dict
@@ -102,7 +102,7 @@ def read_from_cairo_data(
     name: str, values: CairoData, n: int
 ) -> Tuple[CairoData, CairoData]:
     if len(values) < n:
-        raise InvalidDataException(
+        raise InvalidValueException(
             f"Output {name} expected {n} values, {len(values)} values are available."
         )
 
@@ -142,7 +142,7 @@ class FeltTransformer(TypeTransformer[TypeFelt, int]):
             return [value]
 
         if not isinstance(value, int):
-            raise TypeError(f"{name} should be int.")
+            raise InvalidTypeException(f"{name} should be int.")
         cairo_vm_range_check(value)
         return [value]
 
@@ -161,7 +161,7 @@ class StructTransformer(TypeTransformer[TypeStruct, StructTransformerResult]):
         definition = self.identifier_manager.get(cairo_type.scope).identifier_definition
 
         if not isinstance(definition, StructDefinition):
-            raise InvalidDataException(
+            raise InvalidValueException(
                 f"Invalid definition found for {cairo_type.scope}."
             )
 
@@ -175,12 +175,12 @@ class StructTransformer(TypeTransformer[TypeStruct, StructTransformerResult]):
             return [value & ((1 << 128) - 1), value >> 128]
 
         if not isinstance(value, dict):
-            raise TypeError(f"Expected {name} to be a dict.")
+            raise InvalidTypeException(f"Expected {name} to be a dict.")
 
         result = []
         for member_name, member in definition.members.items():
             if member_name not in value:
-                raise InvalidDataException(f"{name}[{member_name}] not provided.")
+                raise InvalidValueException(f"{name}[{member_name}] not provided.")
 
             values = self.resolve_type(member.cairo_type).from_python(
                 member.cairo_type, f"{name}.{member_name}", value[member_name]
@@ -212,7 +212,7 @@ class StructTransformer(TypeTransformer[TypeStruct, StructTransformerResult]):
 class TupleTransformer(TypeTransformer[TypeTuple, Tuple]):
     def from_python(self, cairo_type, name, value) -> CairoData:
         if len(value) != len(cairo_type.members):
-            raise InvalidDataException(
+            raise InvalidValueException(
                 f"Input {name} length mismatch: {len(value)} != {len(cairo_type.members)}."
             )
 
@@ -236,7 +236,7 @@ class TupleTransformer(TypeTransformer[TypeTuple, Tuple]):
 
     def _from_python_named(self, cairo_type, name, values):
         if not isinstance(values, dict) and not TupleTransformer.isnamedtuple(values):
-            raise InvalidDataException(
+            raise InvalidValueException(
                 f"Input {name} is a named tuple and must be dict or NamedTuple"
             )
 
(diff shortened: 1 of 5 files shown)


starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -405,19 +418,25 @@ def to_python(self, value_types: List[dict], values: CairoData) -> NamedTuple:
...
         for name, cairo_type in type_by_name.items():
-            transformed, values = self.resolve_type(cairo_type).to_python(
-                cairo_type, name, values
-            )
+            try:
+                transformed, values = self.resolve_type(cairo_type).to_python(
+                    cairo_type, name, values
+                )
+            except ValueError as err:
+                raise InvalidDataException(str(err)) from err
```
cptartur, line 434: Why was this added? If all errors were replaced by custom ones, are these lines necessary?

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -405,19 +418,25 @@ def to_python(self, value_types: List[dict], values: CairoData) -> NamedTuple:
...
+
             result[name] = transformed
 
         if len(values) > 0:
-            raise ValueError(
+            raise CairoSerializerException(
```
cptartur, line 439: This could probably also be `InvalidDataException`.

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -169,12 +175,12 @@ def from_python(self, cairo_type, name, value) -> CairoData:
             return [value & ((1 << 128) - 1), value >> 128]
 
         if not isinstance(value, dict):
-            raise TypeError(f"Expected {name} to be a dict.")
+            raise InvalidDataException(f"Expected {name} to be a dict.")
```
cptartur, line 178: This is a breaking change

starknet_py/utils/data_transformer/data_transformer.py
war-in: Yes, because not all of them were replaced. Felt still returns ValueError, because I didn't want to change any errors outside the module

starknet_py/utils/data_transformer/data_transformer.py
war-in: You're right, I'll return to the old TypeError here because it is catched in the CairoSerializer.from_python

starknet_py/utils/data_transformer/data_transformer.py
cptartur: Are we sure only "default" exception that can be thrown in this module is `ValueError`? Perhaps this should be general `except Exception as err`?

starknet_py/utils/data_transformer/data_transformer.py
cptartur: Can't we just throw an exception inheriting a proper type here? That's way it's not breaking.

war-in: It is CairoSerializerException and I would like avoid to use it outside CairoSerializer

starknet_py/utils/data_transformer/data_transformer.py
war-in: Yes, it is only the ValueError. Other errors are general python errors and we shouldn't catch them

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -138,7 +142,7 @@ def from_python(self, cairo_type, name, value) -> CairoData:
             return [value]
 
         if not isinstance(value, int):
-            raise TypeError(f"{name} should be int.")
```
cptartur, line 141: This is also a breaking change

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -386,11 +396,14 @@ def from_python(
...
 
-            values = self.resolve_type(cairo_type).from_python(
-                cairo_type, name, named_arguments[name]
-            )
+            try:
+                values = self.resolve_type(cairo_type).from_python(
+                    cairo_type, name, named_arguments[name]
+                )
+            except (TypeError, ValueError) as err:
+                raise InvalidDataException(str(err)) from err
```
cptartur, line 406: This is still kind of problematic, because some `TypeErrors` would be that way replaced by `InvalidDataException` which extends `ValueError`

starknet_py/utils/data_transformer/data_transformer_test.py
```diff
@@ -592,7 +596,7 @@ def test_wrong_types(cairo_type, value):
         }
     ]
 
-    with pytest.raises(TypeError):
```
cptartur, line 595: This should raise `CairoSerializerException`.

cptartur (changes requested):

cptartur (approved):

starknet_py/utils/data_transformer/errors.py
```diff
@@ -0,0 +1,6 @@
+class CairoSerializerException(TypeError):
+    """Exception thrown by CairoSerializer"""
+
+
+class InvalidDataException(ValueError):
```
Solpatium, line 5: So this is an error for `to_python` methods? Maybe we could name it `DeserializationError` and the other one `SerializationError`?

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -386,11 +396,16 @@ def from_python(
...
-                cairo_type, name, named_arguments[name]
-            )
+            try:
+                values = self.resolve_type(cairo_type).from_python(
+                    cairo_type, name, named_arguments[name]
+                )
+            except ValueError as err:
+                raise InvalidDataException(str(err)) from err
+            except TypeError as err:
+                raise CairoSerializerException(str(err)) from err
```
Solpatium, line 408: What kind of exceptions do we expect here?

starknet_py/utils/data_transformer/errors.py
Solpatium, line 5: I would add an abstract class to inherit from that would mean any error coming from data_transformer. This way protostar can expect this one error.

Solpatium: I see that it is not the case. Users won't understand their meaning.

starknet_py/utils/data_transformer/data_transformer.py
war-in: I'm not sure if I understand you correctly 😅 We expect errors from transformer classes (TupleTransformer, FeltTransformer). I've changed errors in the data_transformer.py, but I don't want to do it in e.g. Felt.py (it throws ValueErrors)

starknet_py/utils/data_transformer/data_transformer_test.py
```diff
@@ -617,7 +620,7 @@ def test_arg_provided_twice():
 def test_missing_arg():
     abi = [{"name": "first", "type": "felt"}, {"name": "second", "type": "felt"}]
 
-    with pytest.raises(TypeError) as excinfo:
+    with pytest.raises(CairoSerializerException) as excinfo:
```
Solpatium, line 623: We should still catch exact exceptions here, not the general one

starknet_py/utils/data_transformer/data_transformer.py
Solpatium: Those exceptions might not be very informative for users, but we can keep it this way.

Solpatium (approved):
Just one comment

PR #404, opened 2022-09-28
Repair pyright in unit test

### LINKED ISSUE #392: Fix types in tests

war-in:

Partially resolves #392 


starknet_py/compile/compiler_test.py
```diff
@@ -31,7 +37,7 @@ def test_compile_direct_load():
 
 def test_compile_file_load():
     output_file_str = Compiler(
-        contract_source=[test_file_path.resolve().absolute()]
+        contract_source=[cast(CairoFilename, test_file_path.resolve().absolute())]
```
cptartur, line 40: I'm not sure about cast required for pyright to work with `NewType`s. There should be a solution that doesn't require casting.

starknet_py/utils/sync/sync_test.py
```diff
@@ -76,12 +76,12 @@ async def test_asynchronous_versions():
...
     contract = Contract("1")
-    repository = contract.get_repository_sync()
-    function = repository.get_function_sync()
+    repository = contract.get_repository_sync()  # pyright: ignore
+    function = repository.get_function_sync()  # pyright: ignore
 
     assert function.call_sync() == 1
     assert function.get_name() == "function X"
-    assert contract.example_class_method_sync() == 2
+    assert contract.example_class_method_sync() == 2  # pyright: ignore
```
cptartur, line 84: Why are ignores needed here?

war-in: Because pyright does not recognize _sync methods :/

cptartur: I'd consider including some explanations here and in other places, why pyright was disabled. This probably also applies to code already merged

starknet_py/compile/compiler_test.py
war-in: It is already resolved. I've changed CairoFilename from NewType to str 😅

starknet_py/net/l1/messaging_test.py
```diff
@@ -138,7 +140,7 @@ def get_tx_receipt(_tx_hash):
             }
         )
 
-    w3_mock_receipt.eth.get_transaction_receipt = get_tx_receipt
+    w3_mock_receipt.eth.get_transaction_receipt = get_tx_receipt  # pyright: ignore
```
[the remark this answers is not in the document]
war-in: Small differences in return types. I don't think it is worth resolving

Solpatium (approved):

PR #409, opened 2022-09-30
RPC v0.2.0

war-in:


starknet_py/net/schemas/rpc.py
```diff
@@ -111,13 +111,19 @@ class TransactionSchema(Schema):
 
 
 class InvokeTransactionSchema(TransactionSchema):
-    contract_address = Felt(data_key="contract_address", required=True)
-    entry_point_selector = Felt(data_key="entry_point_selector", required=True)
+    contract_address = Felt(data_key="contract_address", default=None)
+    sender_address = Felt(data_key="sender_address", default=None)
+    entry_point_selector = Felt(data_key="entry_point_selector", default=None)
```
drknzz, line 116: What's the difference between `load_default=None` and `default=None`?

war-in: Not sure, but we should use `load_default` like everywhere

drknzz (approved):
All good, although I am not sure the naming `new_rpc_account_client` is the best. Just a note though. 👌

starknet_py/net/client_models.py
```diff
@@ -278,7 +278,7 @@ class BlockStateUpdate:
 class StateDiff:
     deployed_contracts: List[DeployedContract]
     storage_diffs: List[StorageDiff]
-    declared_contracts: List[int]
```
cptartur, line 281: This is breaking, but I think it's fine to do

starknet_py/net/full_node_client.py
```diff
@@ -167,23 +168,12 @@ async def estimate_fee(
...
-                    "signature": [convert_to_felt(i) for i in tx.signature],
-                    "nonce": (
-                        convert_to_felt(tx.nonce) if tx.nonce is not None else None
-                    ),  # TODO: verify this is correct
-                    "type": "INVOKE",
-                    "contract_address": convert_to_felt(tx.contract_address),
-                    "entry_point_selector": convert_to_felt(tx.entry_point_selector),
-                    "calldata": [convert_to_felt(i) for i in tx.calldata],
-                },
+                "request": {**transaction_params},
```
cptartur, line 176: This could probably be inlined? `"request: _create_brodcasted_txn(...)`?

starknet_py/net/full_node_client.py
```diff
@@ -454,3 +446,53 @@ def _get_call_payload(tx: Union[InvokeFunction, Call]) -> dict:
         "entry_point_selector": convert_to_felt(tx.selector),
         "calldata": [convert_to_felt(i) for i in tx.calldata],
     }
+
+
+def _create_broadcasted_txn(transaction: Union[InvokeFunction, Declare]) -> Dict:
+    common_params = {
+        "type": "INVOKE",
```
cptartur, line 453: Is `"type"` a common param? I don't think a `Declare` transaction version should be `INVOKE`.

starknet_py/net/full_node_client.py
```diff
@@ -454,3 +446,53 @@ def _get_call_payload(tx: Union[InvokeFunction, Call]) -> dict:
...
+                "program": contract_class["program"],
+                "entry_points_by_type": contract_class["entry_points_by_type"],
+                "abi": contract_class["abi"],
+            },
+            "sender_address": convert_to_felt(transaction.sender_address),
+        }
+
+        return {**common_params, **declare_params}
+
+    return {}
```
cptartur, line 498: I don't think we should default to empty dict. If transaction is neither DECLARE nor INVOKE, this method should fail.

starknet_py/tests/e2e/account/account_client_test.py
```diff
@@ -201,7 +201,8 @@ async def test_rejection_reason_in_transaction_receipt(account_client, map_contr
 
     transaction_receipt = await account_client.get_transaction_receipt(res.hash)
 
-    assert "Actual fee exceeded max fee." in transaction_receipt.rejection_reason
+    if isinstance(account_client.client, GatewayClient):
+        assert "Actual fee exceeded max fee." in transaction_receipt.rejection_reason
```
cptartur, line 205: Why was `isinstance` check added?

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -311,11 +311,19 @@ async def test_get_class_hash_at(clients, contract_address, class_hash):
...
+    assert contract_class.entry_points_by_type is not None
+
+
+@pytest.mark.asyncio
+async def test_get_class_by_hash_rpc(rpc_client, class_hash):
+    contract_class = await rpc_client.get_class_by_hash(
+        class_hash=class_hash, block_hash="latest"
+    )
+    assert contract_class.program != ""
+    assert contract_class.entry_points_by_type is not None
```
cptartur, line 326: Perhaps these should be in their respective files for gateway and rpc client tests?

starknet_py/tests/e2e/contract_interaction/interaction_test.py
```diff
@@ -204,7 +205,8 @@ async def test_wait_for_tx_throws_on_transaction_rejected(account_client, map_co
     with pytest.raises(TransactionRejectedError) as err:
         await account_client.wait_for_tx(transaction.hash)
 
-    assert "Entry point 0x123 not found in contract" in err.value.message
+    if isinstance(account_client.client, GatewayClient):
+        assert "Entry point 0x123 not found in contract" in err.value.message
```
cptartur, line 209: Same here

starknet_py/net/schemas/rpc.py
```diff
@@ -111,17 +111,19 @@ class TransactionSchema(Schema):
```
cptartur, line 111: We should include support for L1_HANDLER transactions.

cptartur (changes requested):

starknet_py/net/full_node_client.py
```diff
@@ -472,3 +462,54 @@ def _get_call_payload(tx: Union[InvokeFunction, Call]) -> dict:
...
+        "nonce": convert_to_felt(transaction.nonce)
+        if transaction.nonce is not None
+        else transaction.nonce,
+    }
+
+    if transaction.tx_type == TransactionType.INVOKE_FUNCTION:
+        invoke_specific_params = {
+            "type": "INVOKE",
+            "calldata": [convert_to_felt(i) for i in transaction.calldata],
+        }
```
cptartur, line 481: I get it avoids some duplication but I think it'd be cleared for other developers if we just had
```python
if transaction.version == 0:
  { ... }  # whole v0 transaction
else:
  { ... }  # whole v1 transaction
```
without combining tx version specific and common params at the end.

starknet_py/net/schemas/rpc.py
```diff
@@ -111,17 +112,19 @@ class TransactionSchema(Schema):
...
 
-    @pre_load
-    def preprocess(self, data, **kwargs):
-        return data
-
     @post_load
     def make_transaction(self, data, **kwargs) -> InvokeTransaction:
+        data["contract_address"] = data.get("contract_address") or data.get(
+            "sender_address"
+        )
```
cptartur, line 125: Perhaps we should also make sure that `data["contract_address"]` is not None? (not that it's allowed in spec)

starknet_py/net/schemas/rpc.py
```diff
@@ -155,6 +169,7 @@ class TypesOfTransactionsSchema(OneOfSchema):
         "DEPLOY": DeployTransactionSchema,
         # FIXME add proper handling/serialization
         "DEPLOY_ACCOUNT": None,
```
cptartur, line 171: Do we have an issue for this fixme? If not, let's create one.

starknet_py/net/schemas/rpc.py
```diff
@@ -147,6 +150,17 @@ def make_dataclass(self, data, **kwargs) -> DeployTransaction:
         return DeployTransaction(**data)
 
 
+class L1HandlerTransactionSchema(TransactionSchema):
+    contract_address = Felt(data_key="contract_address", required=True)
+    calldata = fields.List(Felt(), data_key="calldata", required=True)
+    entry_point_selector = Felt(data_key="entry_point_selector", required=True)
+    nonce = Felt(data_key="nonce", load_default=None)
```
cptartur, line 157: Should we allow `None` for these transactions? It seems from the spec that they will always have a nonce, though I'm not 100% sure.

starknet_py/net/schemas/rpc.py
```diff
@@ -21,6 +21,7 @@
     DeployTransactionResponse,
```
cptartur, line 21: DeclaredContractSchema should include `abi`.

cptartur (changes requested):

starknet_py/net/schemas/rpc.py
war-in: The issue is to support v0.2.1 😅

starknet_py/net/schemas/rpc.py
```diff
@@ -111,17 +112,19 @@ class TransactionSchema(Schema):
 
 
 class InvokeTransactionSchema(TransactionSchema):
-    contract_address = Felt(data_key="contract_address", required=True)
-    entry_point_selector = Felt(data_key="entry_point_selector", required=True)
+    contract_address = Felt(data_key="contract_address", load_default=None)
+    sender_address = Felt(data_key="sender_address", load_default=None)
+    entry_point_selector = Felt(data_key="entry_point_selector", load_default=None)
     calldata = fields.List(Felt(), data_key="calldata", required=True)
     nonce = Felt(data_key="nonce", load_default=None)
```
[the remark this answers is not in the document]
war-in: I'm not sure if it is necessary, because v0 will be removed by the end of the year (or am I wrong?)

starknet_py/net/full_node_client.py
war-in: I don't think this is a good idea, because `common_params` are for the `Declare` transaction too

starknet_py/net/schemas/rpc.py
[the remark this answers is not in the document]
cptartur: I guess they will, lets keep it that way then.

starknet_py/net/full_node_client.py
cptartur: Problem is this function is quite complicated to comprehend because parts of transaction come from 3 different places. 

Perhaps there's another way of improving it? Or if you think it's the bast way, we can keep it as is.

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -144,45 +144,24 @@ async def test_get_storage_at_incorrect_address_gateway_client(gateway_client):
...
-                "transaction_hash": "0x7e1ed66dbccf915857c6367fc641c24292c063e54a5dd55947c2d958d94e1a9",
-                "type": "L1_HANDLER",
-            },
-        }
-
-        transaction = await gateway_client.get_transaction(tx_hash=0x1)
+async def test_get_deploy_account_transaction(
+    gateway_client, deploy_account_transaction_hash
+):
+    # TODO Modify this test to also use FullNodeClient (move to client_test)
```
cptartur, line 150: Do we have an issue for this? If not, let's create one.

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -202,21 +181,9 @@ async def test_get_l1_handler_transaction_without_nonce(gateway_client):
                 "type": "L1_HANDLER",
             },
         }
+        mocked_transaction_call.return_value = return_value
```
cptartur, line 184: Why was this split?

starknet_py/tests/e2e/client/gateway_test.py
war-in: https://github.com/software-mansion/starknet.py/issues/418

I've added a checkbox

starknet_py/net/schemas/rpc.py
```diff
@@ -111,17 +112,21 @@ class TransactionSchema(Schema):
...
-        return data
-
     @post_load
     def make_transaction(self, data, **kwargs) -> InvokeTransaction:
+        data["contract_address"] = data.get("contract_address") or data.get(
+            "sender_address"
+        )
+        del data["sender_address"]
+
+        assert data["contract_address"] is not None
```
cptartur, line 128: I don't think we should be doing asserts and maybe we shouldn't be checking this at all?

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -369,3 +391,41 @@ async def test_custom_session(map_contract, network):
...
+                    "0x3635c9adc5dea00000",
+                    "0x0",
+                    "0x7cb4539b69a2371f75d21160026b76a7a7c1cacb",
+                ],
+                "transaction_hash": "0x7e1ed66dbccf915857c6367fc641c24292c063e54a5dd55947c2d958d94e1a9",
+                "type": "L1_HANDLER",
+            },
+        }
+        mocked_transaction_call_gateway.return_value = return_value
+        mocked_transaction_call_rpc.return_value = return_value["transaction"]
```
cptartur, line 425: This is not exactly an RPC formatted response, but I guess for the mocking sake it's good enough.

cptartur (approved):
Looks good, there's still a question about assert in L1HandlerSchema, but after that is addressed (however), ready to merge.

PR #424, opened 2022-10-06
0.10.1 initial support

cptartur:

This PR aims to ensure that all existing modules still work with changes to the feeder gateway and cairo lang in 0.10.1.

StarkNet.py does not work with 0.10.1

Minimal functionality is preserved

Yes

* entry_point_selector was removed from gateway responses (at least for tx version 1) so InvokeTransaction dataclass `entry_point_selector` has been changed to `Optional[int]`

## **Other information**

n/a

### DIFF
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -77,6 +77,7 @@ class TransactionType(Enum):
     INVOKE = "INVOKE"
     DEPLOY = "DEPLOY"
     DECLARE = "DECLARE"
+    DEPLOY_ACCOUNT = "DEPLOY_ACCOUNT"
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


cptartur:
Some updates to the tests are needed for them to pass on 0.10.1.

starknet_py/net/client_models.py
```diff
@@ -103,7 +104,7 @@ class InvokeTransaction(Transaction):
 
     contract_address: int
     calldata: List[int]
-    entry_point_selector: int
+    entry_point_selector: Optional[int] = None
```
cptartur, line 107: This is a breaking change. Alternative solution is to use a default value of `0`, but I'm not sure if this is better.

war-in: I don't think that defaulting to 0 is a good idea. I would leave it as optional @Solpatium

war-in (approved):
Just repair pylint 😄

starknet_py/net/client_models.py
Solpatium: The starknet model is changing and our job should be to model it properly. This field is not optional, so starknet.py should follow. Please add a comment when this is optional.

PR #447, opened 2022-10-10
Bump starknet-devnet

Solpatium:
e2e tests should be faster now


pyproject.toml
```diff
@@ -41,7 +41,7 @@ pytest-mock = "^3.6.1"
 pytest-xdist = "^2.5.0"
 web3 = {extras=["tester"], version="6.0.0b4"}
 pyright = "^1.1.270"
-starknet-devnet = "0.3.1"
+starknet-devnet = "^0.3.3"
```
war-in, line 44: It would be better to use only one devnet version `"0.3.3"`

Issue #438, opened 2022-10-10
Add `abi` to result of `gateway_client.get_class_by_hash` when RPC v0.2.0 is supported


drknzz:
Current `DeclaredContract` lacks `abi`, which is returned as shown [here](https://starknet.io/docs/hello_starknet/cli.html#get-class-by-hash).

```
@dataclass
class DeclaredContract:
    """
    Dataclass representing contract declared to starknet
    """

    program: dict
    entry_points_by_type: EntryPointsByType
```

Also, by then, Contract.from_address should support FullNodeClient.

PR #448, opened 2022-10-11
Refactor Contract.from_address

### LINKED ISSUE #425: Outdated Proxy Resolver
The current [OpenZeppelin](https://github.com/OpenZeppelin/cairo-contracts/blob/v0.4.0/src/openzeppelin/upgrades/presets/Proxy.cairo) and [Argent](https://github.com/argentlabs/argent-contracts-starknet/blob/cb478d1e7d3f1455b617372daf4fd7512f23e496/contracts/upgrade/Proxy.cairo) Proxy contracts use `class hash` instead of `address` in their storage.

Currently `Contract.from_address` operates on addresses in storage and resolves proxy chains by using `ProxyChecks` with outdated `implementation_address` method.

drknzz:

Closes #425 

## Notes
- `contracts_compiled` directory with compiled contracts which names start with an underscore ( so `_example_contract_name.json`) - these contracts will not be compiled when `poe compile_contracts` is run
- added test for OpenZeppelin proxy contract with `implementation_address` - test uses contract compiled with starknet-compile 0.8.1 as it was the latest that supported `delegate_call` / `delegate_l1_handler` syscalls
- added custom ProxyCheck test


starknet_py/contract.py
```diff
@@ -590,51 +618,12 @@ def _make_functions(
...
-
-        if not is_proxy:
-            return contract
-
-        return await self._make_contract_recursively(
-            address=address,
-            step=step + 1,
-        )
+    @staticmethod
+    def _prepare_proxy_config(proxy_config: ProxyConfig) -> ProxyConfig:
```
war-in, line 622: Should it be `@staticmethod` for sure?

starknet_py/proxy_check.py
```diff
@@ -16,9 +13,11 @@ async def is_proxy(self, contract: "Contract") -> bool:
         """
 
     @abstractmethod
-    async def implementation_address(self, contract: "Contract") -> int:
+    async def implementation(self, contract: "Contract") -> int:
         """
-        :return: Address of contract being proxied by `contract` given as a parameter
+        :return: Implementation (either class hash or contract address)
+            of contract being proxied by `contract` given as a parameter
```
cptartur, line 19: Allowing this to be either class hash or contract address complicates the handling of the proxies. I think we could extend the class hierarchy a little bit and add something along the lines of `ProxyCheckAddress` and `ProxyCheckClassHash` as the subclasses of the `ProxyCheck` and then handle them accordingly based on their type (with `isinstance` checks).

starknet_py/contract.py
```diff
@@ -398,44 +399,71 @@ async def from_address(
...
+            )
 
-        return Contract(address=address, abi=contract.data.abi, client=client)
+        try:
+            contract_class = await proxy_client.get_class_by_hash(implementation)
+        except ClientError as err:
+            if "Class with hash" not in err.message:
+                raise err
+            class_hash = await proxy_client.get_class_hash_at(implementation)
+            contract_class = await proxy_client.get_class_by_hash(class_hash)
```
cptartur, line 464: This is quite complicated. See my comment about the ProxyCheck structure below.

cptartur (changes requested):

starknet_py/contract.py
```diff
@@ -590,51 +595,89 @@ def _make_functions(
...
-        max_steps: int,
-        proxy_checks: List[ProxyCheck],
-    ):
-        self._address = address
-        self._client = client
-        self._max_steps = max_steps
-        self._proxy_checks = proxy_checks
-        self._processed_addresses = set()
+    @staticmethod
+    async def _from_address_proxy_step(
```
Solpatium, line 610: I don't like that so much logic related to resolving proxy is leaking here. I think I would rather leave `ContractFromAddressFactory` (although name could be improved, like ContractAbiResolver) or create a separate module for that.

starknet_py/contract.py
```diff
@@ -590,51 +595,89 @@ def _make_functions(
...
+        implementation = await proxy_check.implementation(proxy_contract)
+        if implementation != 0:
+            break
+
+    if implementation == 0:
+        raise ProxyResolutionError(proxy_checks)
+    return implementation
+
+
+def get_actual_client(client: Client) -> Client:
```
Solpatium, line 682: Any functions that are not part of the public API should be prefixed with `_`

starknet_py/contract.py
```diff
@@ -590,51 +595,89 @@ def _make_functions(
...
+        proxy_client = proxy_contract.client
+
+        try:
+            contract_class = await proxy_client.get_class_by_hash(implementation)
+        except ClientError as err:
+            if "is not declared" not in err.message:
+                raise err
+            contract_class = await get_class_by_address(
+                address=implementation, client=proxy_client
+            )
```
Solpatium, line 629: This logic doesn't make sense for OpenZeppelin, just adds an overhead.

starknet_py/contract.py
```diff
@@ -398,44 +408,39 @@ async def from_address(
...
-            proxy_config = default_config if proxy_config else {}
-        else:
-            proxy_config = {**default_config, **proxy_config}
-
-        proxy_client = client.client if isinstance(client, AccountClient) else client
-        if not isinstance(proxy_client, GatewayClient):
+        address = parse_address(address)
+        proxy_config = prepare_proxy_config(proxy_config)
+        actual_client = get_actual_client(client)
+        if not isinstance(actual_client, GatewayClient):
```
Solpatium, line 430: Can't we support any client now that we use contract classes?

starknet_py/contract.py
```diff
@@ -590,51 +595,89 @@ def _make_functions(
...
+    proxy_contract: Contract, proxy_config: Contract.ProxyConfig
+) -> int:
+    implementation, proxy_checks = 0, proxy_config.get("proxy_checks", [])
+    for proxy_check in proxy_checks:
+        implementation = await proxy_check.implementation(proxy_contract)
+        if implementation != 0:
+            break
+
+    if implementation == 0:
+        raise ProxyResolutionError(proxy_checks)
```
Solpatium, line 678: What if someone provides a contract without a proxy and provides config = True? In this case shouldn't we see that there is no proxy and return abi of the original contract?

starknet_py/proxy_check.py
Solpatium: I think we should resolve the abi for an address, why would we need to resolve for class hash?

starknet_py/proxy_check.py
```diff
@@ -27,26 +26,31 @@ class ArgentProxyCheck(ProxyCheck):
...
         return "get_implementation" in contract.functions
 
-    async def implementation_address(self, contract: "Contract") -> int:
-        res = await contract.functions["get_implementation"].call()
-        return res[0]
+    async def implementation(self, contract: "Contract") -> int:
+        try:
+            (result,) = await contract.functions["get_implementation"].call()
+            return result
+        except KeyError:
```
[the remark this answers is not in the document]
Solpatium: You might be thinking about "it's better to ask for forgiveness than to ask for permission" rule, but it mostly applies when trying to get access some resources that we don't control.

Solpatium (changes requested):

starknet_py/net/client_models.py
```diff
@@ -320,6 +320,7 @@ class DeclaredContract:
 
     program: dict
     entry_points_by_type: EntryPointsByType
+    abi: Optional[List[Dict[str, Any]]] = None
```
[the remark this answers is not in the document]
drknzz: It is not, but classes from `client_models.py` are shared between gateway and full node clients, and the latter does not have it in RPC v0.1.0

There is an [issue](https://github.com/software-mansion/starknet.py/issues/438) for this.

starknet_py/utils/data_transformer/data_transformer.py
```diff
@@ -368,22 +368,22 @@ def from_python(
...
-                f"Provided {len(args)} positional arguments, {len(type_by_name)} possible."
+                f"Provided {len(args)} positional arguments, {len(value_types)} possible."
             )
 
-        key_diff = set(named_arguments.keys()).difference(set(type_by_name))
-
+        named_arguments = {**kwargs}
+        key_diff = set(named_arguments.keys()).difference(
+            set(val["name"] for val in value_types)
+        )
```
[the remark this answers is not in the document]
drknzz: `CairoSerializer._abi_to_types` removes parameters that are lengths of an array, effectively changing its input length to smaller.
Passing abi with length of an array would incorrectly throw exceptions.

starknet_py/contract.py
drknzz: There is no way of getting `abi` from a contract using RPC v0.1.0, so only GatewayClient can be used for now.

starknet_py/tests/e2e/proxy_check/proxy_test.py
```diff
@@ -1,140 +1,171 @@
...
+    get_storage_var_address,
+)
 
 from starknet_py.contract import Contract
+from starknet_py.net.client_errors import ContractNotFoundError
+from starknet_py.proxy_check import ProxyResolutionError, ProxyCheck
+
 
-MAX_FEE = int(1e20)
+async def put_and_get_map_contract(map_contract: Contract, key: int, val: int) -> int:
```
Solpatium, line 13: I would change this into `assert_map_works_properly` and set&get some custom value in here with an assertion.

starknet_py/contract_abi_resolver.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        Returns abi of a contract directly from given address.
+        """
+        contract_class = await ContractAbiResolver._get_class_by_address(
+            address=address, client=client
+        )
+        return contract_class.abi or []
+
+    @staticmethod
+    async def resolve_proxy_step(
```
Solpatium, line 74: `resolve_abi`?

starknet_py/contract_abi_resolver.py
```diff
@@ -0,0 +1,131 @@
...
+    ADDRESS = "address"
+
+
+class ContractAbiResolver:
+    """
+    Class for resolving abi of a contract
+    """
+
+    @staticmethod
+    async def resolve_direct(
```
Solpatium, line 61: `get_abi_for_address`

starknet_py/contract_abi_resolver.py
```diff
@@ -0,0 +1,131 @@
...
+class ImplementationType(Enum):
+    """
+    Enum representing transaction types
+    """
+
+    CLASS_HASH = "class_hash"
+    ADDRESS = "address"
+
+
+class ContractAbiResolver:
```
Solpatium, line 55: Instead of passing proxy config to multiple methods you can just make them regular object methods and accept parameter. This is a more pythonic way.

starknet_py/contract_abi_resolver.py
```diff
@@ -0,0 +1,131 @@
+# Needed because of string typed Contract
```
Solpatium, line 1: I think we can get rid of this. I suggest not accepting Contract at all & just use a client. For Argent we can perform a call using just a simple call. 
I don't want the low level interface to depend on high level interface, coupling them bidirectionally.

starknet_py/proxy_check.py
```diff
@@ -2,51 +2,72 @@
 # pyright: reportUndefinedVariable=false
 
 from abc import ABC, abstractmethod
```
Solpatium, line 4: Since we are introducing breaking changes anyway we should move this and resolver to proper directory within starknet_py

starknet_py/proxy_check.py
```diff
@@ -2,51 +2,72 @@
...
 
-    async def implementation_address(self, contract: "Contract") -> int:
-        res = await contract.functions["get_implementation"].call()
-        return res[0]
+    async def implementation_hash(self, contract: "Contract") -> Optional[int]:
+        if "get_implementation" not in contract.functions:
+            return None
+
+        (result,) = await contract.functions["get_implementation"].call()
+        return result
```
Solpatium, line 41: If argent is using address for proxies this will return this address and claim that it is a class_hash. We do need to check if the value is address or class_hash.

starknet_py/contract_abi_resolver.py
```diff
@@ -0,0 +1,131 @@
...
+        for proxy_check in proxy_checks:
+            implementation = await proxy_check.implementation_hash(proxy_contract)
+            if implementation is not None:
+                return implementation, ImplementationType.CLASS_HASH
+
+            implementation = await proxy_check.implementation_address(proxy_contract)
+            if implementation is not None:
+                return implementation, ImplementationType.ADDRESS
+
+        raise ProxyResolutionError(proxy_checks)
```
Solpatium, line 112: Looks like this exception is not a part of Proxies themselves, but `ContractAbiResolver` itself I suggest moving this exception to this file.

starknet_py/tests/e2e/proxy_check/proxy_test.py
```diff
@@ -1,140 +1,171 @@
...
+        address=deployment_result.deployed_contract.address,
+        client=gateway_account_client,
     )
-    implementation = "implementation" if argent_proxy else "new_implementation"
 
-    params = {
-        "proxy1": {implementation: proxy2.address},
-        "proxy2": {implementation: proxy1.address},
-    }
+    assert all(f in contract.functions for f in ("put", "get"))
```
Solpatium, line 37: nitpick: assert knows how to compare different structures when using != and ==. You can use that to compare sets. This way when there is something missing it will let you know what. With current code you'll only know that returned value was false.

```
assert set(contract.functions.keys()) == {"put", "get"}
```

Solpatium (changes requested):
Almost there! :)

starknet_py/contract.py
```diff
@@ -398,44 +386,42 @@ async def from_address(
...
+        )
+        direct_contract = Contract(address=address, abi=direct_abi, client=client)
+        if not proxy_config:
+            return direct_contract
 
-        return Contract(address=address, abi=contract.data.abi, client=client)
+        abi = await ContractAbiResolver.resolve_proxy_step(
+            proxy_contract=direct_contract, proxy_config=proxy_config
+        )
+        return Contract(address=address, abi=abi, client=client)
```
cptartur, line 424: I think this flow can still be improved. Contract shouldn't need to be aware of details about proxies. Instead it should be possible to do something along the lines of `ContractAbiResolve.resolve(...)` (or even name this module something different).

starknet_py/contract_abi_resolver.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+
+    proxy_checks: List[ProxyCheck]
+    """
+    List of classes implementing :class:`starknet_py.proxy_check.ProxyCheck` ABC,
+    that will be used for checking if contract at the address is a proxy contract.
+    """
+
+
+def prepare_proxy_config(proxy_config: Union[bool, ProxyConfig]) -> ProxyConfig:
```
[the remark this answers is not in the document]
cptartur: I'd prefer we have straightforward implementation here: proxy_config can be spread into simple arguments instead of union here, and we can have necessary logic for parsing `proxy_config` in Contract, like you suggested in the second point.

starknet_py/contract_abi_resolver.py
```diff
@@ -0,0 +1,131 @@
...
+        address: Address,
+        client: Client,
+    ) -> Abi:
+        """
+        Returns abi of a contract directly from given address.
+        """
+        contract_class = await ContractAbiResolver._get_class_by_address(
+            address=address, client=client
+        )
+        return contract_class.abi or []
```
cptartur, line 71: What if we return `[]` as abi? Doesn't that make `Contract` instance created using it useless? Perhaps we shoud raise if we cannot get the abi or add some handling to usages of this method in `Contract`?

cc @Solpatium

starknet_py/contract_abi_resolver.py
cptartur, line 1: This should be placed in some separate module along with proxy_checks instead of top level directory.

cptartur (changes requested):
I like where this is going but the flow of resolving proxies can still be improved.

Ideally I'd like this to be whole, self contained module, which can perform an `resolve()` operation, so that we don't have to worry about the proper usage in `Contract.

starknet_py/contract_abi_resolver.py
Solpatium: When is is it even possible to get None for abi? I think an empty list should always be returned from api 🤔

cptartur: I think it's a problem with 0.1.0 version of the rpc spec we currently support: It doesn't have abis at all. 
Even though, if the api returns empty list, it's still problematic in the same way as default `[]` is.

drknzz: Gateway does currently always return `abi`, but since it is not in RPC, `abi` is marked optional in `DeclaredContract`.
The statement `or []` in return was purely for typechecker not to throw errors.

cptartur: I'd do some more explicit handling, using default `[]` will cause issues down the road.

pyproject.toml
```diff
@@ -67,6 +67,7 @@ docs_open = { shell = "open docs/_build/html/index.html" }
 lint = "pylint starknet_py"
 format = "black starknet_py"
 format_check = "black --check starknet_py"
+format_diff = "black --diff starknet_py"
```
cptartur, line 70: Why was this added?

starknet_py/contract.py
```diff
@@ -398,44 +386,43 @@ async def from_address(
...
 
-        contract = await ContractFromAddressFactory(
-            address=address,
-            client=proxy_client,
-            max_steps=proxy_config.get("max_steps", 1),
-            proxy_checks=proxy_config.get("proxy_checks", []),
-        ).make_contract()
+        abi = await ContractAbiResolver(
+            address=address, client=client, proxy_config=proxy_config
+        ).resolve()
```
cptartur, line 423: 👍

starknet_py/contract.py
```diff
@@ -398,44 +386,43 @@ async def from_address(
...
         :return: an initialized Contract instance
         """
-        default_config: Contract.ProxyConfig = {
-            "max_steps": 5,
-            "proxy_checks": [ArgentProxyCheck(), OpenZeppelinProxyCheck()],
-        }
-        if isinstance(proxy_config, bool):
-            proxy_config = default_config if proxy_config else {}
+        address = parse_address(address)
+        if isinstance(proxy_config, bool) and not proxy_config:
```
cptartur, line 406:
```suggestion
        if proxy_config is False:
```

starknet_py/contract.py
```diff
@@ -398,44 +386,43 @@ async def from_address(
...
+            proxy_config = ProxyConfig()
         else:
-            proxy_config = {**default_config, **proxy_config}
+            proxy_arg = (
+                ProxyConfig() if isinstance(proxy_config, bool) else proxy_config
```
cptartur, line 410:
```suggestion
                ProxyConfig() if proxy_config is True else proxy_config
```

starknet_py/contract.py
```diff
@@ -398,44 +386,43 @@ async def from_address(
...
+            )
+            proxy_config = prepare_proxy_config(proxy_arg)
```
cptartur, line 412: I think we can move this logic to a private helper function

starknet_py/contract.py
```diff
@@ -398,44 +386,43 @@ async def from_address(
...
 
-        proxy_client = client.client if isinstance(client, AccountClient) else client
-        if not isinstance(proxy_client, GatewayClient):
+        actual_client = client.client if isinstance(client, AccountClient) else client
+        if not isinstance(actual_client, GatewayClient):
+            # TODO: Add support for FullNodeClient once abi is available in RPC
             raise TypeError(
                 "Contract.from_address only supports GatewayClient or AccountClients using GatewayClient"
             )
```
cptartur, line 419: Let's move it to a helper like `assert_can_get_abi` or something like that

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def prepare_proxy_config(proxy_config: ProxyConfig) -> ProxyConfig:
+    if "max_steps" in proxy_config:
+        warnings.warn(
+            "ProxyConfig.max_steps is deprecated. Contract.from_address always makes at most 1 step.",
+            category=DeprecationWarning,
+        )
+    proxy_checks = [OpenZeppelinProxyCheck(), ArgentProxyCheck()]
+    return {"proxy_checks": proxy_checks, **proxy_config}
```
cptartur, line 35: Should we be always appending default `proxy_checks` to the user provided? What if the user doesn't want to resolve for e.g. OZ proxies?

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,150 @@
...
+        :param proxy_config: Proxy config for resolving proxy
+        """
+        self.address = address
+        self.client = client
+        self.proxy_config = ProxyConfig() if proxy_config is None else proxy_config
+
+    async def resolve(self) -> Abi:
+        """
+        Returns abi of either direct contract or contract proxied by direct contract depending on proxy_config.
+        """
```
cptartur, line 70: Let's document custom exception in this docstring, like `:raises ...`

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,150 @@
...
+        if self.proxy_config:
+            return await self.resolve_abi()
+        return await self.get_abi_for_address()
+
+    async def get_abi_for_address(self) -> Abi:
+        """
+        Returns abi of a contract directly from address.
+        """
```
cptartur, line 78: Same here

starknet_py/tests/e2e/conftest.py
```diff
@@ -434,7 +436,53 @@ def compiled_proxy(request) -> str:
...
+        compiled_contract=compiled_proxy,
+        constructor_args={
+            implementation_key: declare_result.class_hash,
+            "selector": get_selector_from_name("put"),
+            "calldata": [69, 420],
+        },
+        client=gateway_account_client,
+    )
+    await deployment_result.wait_for_acceptance()
+    return deployment_result
```
cptartur, line 485: test module now contains a `utils.py` file. Let's make this a normal function and move it there since this fixture isn't really used as a fixture.

starknet_py/tests/e2e/conftest.py
```diff
@@ -486,3 +534,7 @@ def default_gateway_gas_price() -> int:
     the GatewayClient.
     """
     return StarknetGeneralConfig().min_gas_price
+
+
+def read_compiled_contract(file_name: str) -> str:
+    return (contracts_compiled_dir / file_name).read_text("utf-8")
```
cptartur, line 540: Let's add a docstring for this just for the sake of consistency.

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -4,9 +4,8 @@ MOCK_DIRECTORY=starknet_py/tests/e2e/mock
 CONTRACTS_DIRECTORY="$MOCK_DIRECTORY"/contracts
 CONTRACTS_COMPILED_DIRECTORY="$MOCK_DIRECTORY"/contracts_compiled
 
-# delete all artifacts
-rm -rf $CONTRACTS_COMPILED_DIRECTORY
-mkdir $CONTRACTS_COMPILED_DIRECTORY
+# delete all artifacts except those starting with _
+find $CONTRACTS_COMPILED_DIRECTORY -not -regex '.*/_.*json$' -regex '.*json$' -delete
```
cptartur, line 8: Why are we omitting these with `_`?

starknet_py/tests/e2e/mock/contracts/openzeppelin/upgrades/library_custom.cairo
```diff
@@ -0,0 +1,132 @@
...
+    ) {
+        with_attr error_message("Proxy: implementation hash cannot be zero") {
+            assert_not_zero(new_implementation);
+        }
+
+        Proxy_implementation_hash_custom.write(new_implementation);
+        Upgraded.emit(new_implementation);
+        return ();
+    }
+}
```
cptartur, line 132: Newline

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,150 @@
...
+class ProxyResolutionError(Exception):
+    """
+    Error while resolving proxy using ProxyChecks.
+    """
+
+    def __init__(self, proxy_checks: List[ProxyCheck]):
+        proxy_checks_classes = [proxy_check.__class__ for proxy_check in proxy_checks]
+        self.message = (
+            f"Couldn't resolve proxy using given ProxyChecks ({proxy_checks_classes})"
+        )
```
cptartur, line 149: Is this message covered by any tests?

cptartur (changes requested):
Proxy resolver module looks much better now 👍 

I still have some comments though

Solpatium (changes requested):

starknet_py/proxy/contract_abi_resolver.py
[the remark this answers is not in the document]
drknzz: I know, it was working like this originally. Should I change it? If so how should it behave?

starknet_py/tests/e2e/mock/compile_contracts.sh
drknzz: See PR description

starknet_py/contract.py
Solpatium: 👍

starknet_py/proxy/contract_abi_resolver.py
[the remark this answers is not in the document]
Solpatium: Yes, you are right. It can stay this way.

starknet_py/contract.py
```diff
@@ -398,44 +386,36 @@ async def from_address(
...
-        else:
-            proxy_config = {**default_config, **proxy_config}
-
-        proxy_client = client.client if isinstance(client, AccountClient) else client
-        if not isinstance(proxy_client, GatewayClient):
+        address = parse_address(address)
+        proxy_config = Contract._create_proxy_config(proxy_config)
+
+        if not Contract._abi_compatible_client(client):
+            # TODO: Add support for FullNodeClient once abi is available in RPC
```
cptartur, line 409: Let's create an issue for this before we merge.

starknet_py/proxy/contract_abi_resolver.py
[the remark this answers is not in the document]
cptartur: Maybe let's make it more explicit and do `return {"proxy_checks": proxy_config if proxy_config is not None else proxy_checks}`

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,150 @@
...
+        self.address = address
+        self.client = client
+        self.proxy_config = ProxyConfig() if proxy_config is None else proxy_config
+
+    async def resolve(self) -> Abi:
+        """
+        Returns abi of either direct contract or contract proxied by direct contract depending on proxy_config.
+        """
+        if self.proxy_config:
+            return await self.resolve_abi()
```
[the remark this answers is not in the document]
cptartur: Maybe let's just use a more explicit check here?

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,155 @@
...
+
+    async def resolve_abi(self) -> Abi:
+        """
+        Returns abi of a contract that is being proxied by contract at address.
+        :raises ContractNotFoundError: when contract could not be found at address
+        :raises ProxyResolutionError: when given ProxyChecks were not sufficient to resolve proxy
+        :raises AbiNotFoundError: when abi is not present in proxied contract class at address
+        """
+        impl = await self._get_implementation_from_proxy()
+        implementation, implementation_type = impl
```
cptartur, line 97: Can we combine those lines?

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,155 @@
...
+            contract_class = await self.client.get_class_by_hash(implementation)
+        else:
+            contract_class = await self._get_class_by_address(address=implementation)
+
+        if contract_class.abi is None:
+            raise AbiNotFoundError()
+        return contract_class.abi
+
+    async def _get_implementation_from_proxy(self) -> Tuple[int, ImplementationType]:
+        implementation, proxy_checks = None, self.proxy_config.get("proxy_checks", [])
```
cptartur, line 109: Let's split this to separate lines

starknet_py/proxy/proxy_check.py
```diff
@@ -0,0 +1,104 @@
...
+    ) -> Optional[int]:
+        call = self._get_implementation_call(address=address)
+        try:
+            (implementation_address,) = await client.call_contract(invoke_tx=call)
+            await client.get_class_hash_at(contract_address=implementation_address)
+        except ClientError as err:
+            if re.search(
+                r"(Entry point 0x[0-9a-f]+ not found in contract)|(is not declared)",
+                err.message,
+                re.I,
```
cptartur, line 48: Nitpick, but maybe `re.IGNORECASE` is more readable?

starknet_py/proxy/proxy_check.py
```diff
@@ -0,0 +1,104 @@
...
+            ):
+                return None
+            raise err
+        return implementation_address
+
+    async def implementation_hash(
+        self, address: Address, client: Client
+    ) -> Optional[int]:
+        call = ArgentProxyCheck._get_implementation_call(address=address)
```
cptartur, line 57: Why is `ArgentProxyCheck._get_implementation_call` used here and above it's `self._get_implementation_call`?

starknet_py/proxy/proxy_check.py
```diff
@@ -0,0 +1,104 @@
...
+            selector=get_selector_from_name("get_implementation"),
+            calldata=[],
+        )
+
+
+class OpenZeppelinProxyCheck(ProxyCheck):
+    async def implementation_address(
+        self, address: Address, client: Client
+    ) -> Optional[int]:
+        return await self._get_storage_at(
```
cptartur, line 84: It's confusing to name the method the same as `Client` method. I'd rename it to something like `_get_storage_at_or_none` or some entirely different name.

starknet_py/tests/e2e/proxy/proxy_test.py
```diff
@@ -0,0 +1,173 @@
...
+
+from starknet_py.contract import Contract
+from starknet_py.net.client import Client
+from starknet_py.net.client_errors import ContractNotFoundError
+from starknet_py.net.models import Address
+from starknet_py.proxy.contract_abi_resolver import ProxyResolutionError
+from starknet_py.proxy.proxy_check import ProxyCheck
+
+
+async def map_works_properly(map_contract: Contract, key: int, val: int) -> bool:
```
cptartur, line 14:
```suggestion
async def is_map_working_properly(map_contract: Contract, key: int, val: int) -> bool:
```

starknet_py/tests/e2e/proxy/proxy_test.py
```diff
@@ -0,0 +1,173 @@
...
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "deploy_proxy_to_contract",
+    [
+        ("oz_proxy_compiled.json", "map_compiled.json"),
+        ("argent_proxy_compiled.json", "map_compiled.json"),
+    ],
+    indirect=True,
+)
```
cptartur, line 51: Can't we use a parametrized fixture instead of indirect? It's more confusing this way.

starknet_py/tests/e2e/proxy/proxy_test.py
```diff
@@ -0,0 +1,173 @@
...
+    )
+
+    assert proxied_contract.functions.keys() == {"put", "get"}
+    assert proxied_contract.address == proxy_contract.address
+    assert await map_works_properly(map_contract=proxied_contract, key=69, val=13)
+
+
+@pytest.mark.asyncio
+async def test_contract_from_address_unsupported_client(rpc_account_client):
+    with pytest.raises(TypeError, match=r".only supports GatewayClient."):
```
cptartur, line 72: I didn't know you could do regexes in pytest.raises. Good find 👍

starknet_py/tests/e2e/proxy/proxy_test.py
```diff
@@ -0,0 +1,173 @@
...
+            client=gateway_account_client,
+        )
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "deploy_proxy_to_contract",
+    [("oz_proxy_custom_compiled.json", "map_compiled.json")],
+    indirect=True,
+)
```
cptartur, line 93: Same here

starknet_py/tests/e2e/proxy/proxy_test.py
```diff
@@ -0,0 +1,173 @@
...
+            client=gateway_account_client,
+            proxy_config=True,
+        )
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "deploy_proxy_to_contract",
+    [("oz_proxy_custom_compiled.json", "map_compiled.json")],
+    indirect=True,
```
cptartur, line 111: Same here

starknet_py/proxy/contract_abi_resolver.py
drknzz: No, because I'd really like to keep the full names of the variables and in one line case this abomination is output when running `poe format_diff`:
```
-        implementation, implementation_type = await self._get_implementation_from_proxy()
+        (
+            implementation,
+            implementation_type,
+        ) = await self._get_implementation_from_proxy()
```

cptartur: Okay, let's keep it that way then.

starknet_py/proxy/proxy_check.py
```diff
@@ -0,0 +1,104 @@
...
+            await client.get_class_hash_at(contract_address=implementation_address)
+        except ClientError as err:
+            if re.search(
+                r"(Entry point 0x[0-9a-f]+ not found in contract)|(is not declared)",
+                err.message,
+                re.IGNORECASE,
+            ):
+                return None
+            raise err
+        return implementation_address
```
Solpatium, line 52: You repeat the same code below, just with the replacement of get_class_hash_at and err message. I suggest adding `get_implementation` that would perform the call and handle the specific error there. Then you could do something like this:
```
implementation_address = await self.get_implementation(address)
try:
   await client.get_class_hash_at(contract_address=implementation_address)
   return implementation_address
catch ClientError as err:
   if "is not declared" in err.message:
     return None
   raise err
```

There might be more code in general, but we avoid repetition and make the code cleaner.

Solpatium (approved):
LGTM with comments

starknet_py/contract.py
```diff
@@ -590,51 +569,14 @@ def _make_functions(
...
-
-        if not is_proxy:
-            return contract
-
-        return await self._make_contract_recursively(
-            address=address,
-            step=step + 1,
-        )
+    @staticmethod
+    def _abi_compatible_client(client) -> bool:
```
cptartur, line 580:
```suggestion
    def _is_abi_compatible_client(client) -> bool:
```

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -0,0 +1,157 @@
...
+class ContractAbiResolver:
+    """
+    Class for resolving abi of a contract
+    """
+
+    def __init__(
+        self,
+        address: Address,
+        client: Client,
+        proxy_config: ProxyConfig,
```
cptartur, line 59: So if you don't want to resolve proxies at all, you have to pass `proxy_config=[]`? Maybe that's been already discussed but perhaps more ergonomic making it `Optional[ProxyConfig]=None` and defaulting to empty list if it's None

cc @Solpatium

Solpatium: The most important aspect of ContractAbiResolver is being able to resolve ABI behind a proxy. I don't think making it easier is needed here. If users complain we can change it of course.

cptartur: Okay, let's keep it this way then.

cptartur (approved):

PR #449, opened 2022-10-12
Deprecate `Deploy` transaction

### LINKED ISSUE #420: Deprecate methods utilising `DEPLOY` transactions
While `DEPLOY` transactions will probably still be supported for a while, we should either start removing methods that rely on `DEPLOY` transactions or replace them with `DEPLOY_ACCOUNT` transactions.

war-in:

Closes #420 

No

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -460,6 +460,12 @@ class Contract:
         :raises: `ValueError` if neither compilation_source nor compiled_contract is provided.
         :return: DeployResult instance
         """
+        warnings.warn(
+            "In the future versions of StarkNet, Deploy transaction will not be supported."
+            "To deploy a contract use cairo syscall or Universal Deployer Contract",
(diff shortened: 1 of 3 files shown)


starknet_py/contract.py
```diff
@@ -460,6 +460,12 @@ async def deploy(
         :raises: `ValueError` if neither compilation_source nor compiled_contract is provided.
         :return: DeployResult instance
         """
+        warnings.warn(
+            "In the future versions of StarkNet, Deploy transaction will not be supported."
+            "To deploy a contract use cairo syscall or Universal Deployer Contract",
+            category=DeprecationWarning,
+        )
```
cptartur, line 467: I'm not sure if we should be mentioning the UDC just yet. Maybe let's include it in the warning once it's deployed on testnet and our support is released.

We can create issue to do this.

Issue #347, opened 2022-09-02
Move contents `client_models` to `net.models`, deprecate `client_models`.


cptartur:
**Is your feature request related to a problem? Please describe.**

Refactor starknet.py so all our client related models are in `net.models` module.

**Describe the solution you'd like**

Move contents of `client_models.py` to separate files in `net.models` and deprecate it.

**Describe alternatives you've considered**

none

**Additional context**

n/a

cptartur:
Blocked mostly by #376

cptartur:
We should consider defining our own dataclasses for transactions once we move them.

tomek0123456789:
Will be implemented in #1047

PR #471, opened 2022-10-19
Split client_models into multiple files in models directory

Solpatium:
Closes https://github.com/software-mansion/starknet.py/issues/347

### DIFF
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -1,355 +1,45 @@
-from abc import ABC
-from dataclasses import dataclass, field
-from enum import Enum
-from typing import List, Any, Dict, Optional, Union, Iterable
-from typing_extensions import Literal
-
-from starkware.starknet.services.api.gateway.transaction import (
-    InvokeFunction as IF,
-    Transaction as T,
-    ContractClass as CD,
-    Deploy as D,
-    Declare as DCL,
+# pylint: disable=unused-import
+from starknet_py.net.models.blocks import (
+    BlockStatus,
+    StarknetBlock,
+    GatewayBlock,
+    BlockSingleTransactionTrace,
+    BlockTransactionTraces,
+    StorageDiff,
+    StateDiff,
+    BlockStateUpdate,
+    Hash,
+    Tag,
 )
-
-
-from starknet_py.utils.docs import as_our_module
-
-InvokeFunction = as_our_module(IF)
-StarknetTransaction = as_our_module(T)
-ContractClass = as_our_module(CD)
-Deploy = as_our_module(D)
-Declare = as_our_module(DCL)
-
-
-Hash = Union[int, str]
-Tag = Literal["pending", "latest"]
-
-
-@dataclass
-class Call:
-    to_addr: int
-    selector: int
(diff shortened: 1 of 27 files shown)


starknet_py/net/models/transaction_payloads.py
```diff
@@ -0,0 +1,166 @@
+from abc import ABC
```
cptartur, line 1: I think `payloads` is misleading. These are transactions returned by starknet, not ones that can be sent.

starknet_py/net/gateway_client.py
```diff
@@ -7,26 +7,28 @@
...
+    BlockTransactionTraces,
     BlockStateUpdate,
-    InvokeFunction,
-    StarknetTransaction,
     Hash,
     Tag,
-    DeclaredContract,
+)
+from starknet_py.net.models.transaction import (
+    Transaction,
```
cptartur, line 20: Using `Transaction` for both making and receiving transactions will cause problems. That's why it was named `StarknetTransaction` before (though it's not an ideal name).

Also this renaming is kind of breaking?

starknet_py/net/signer/base_signer.py
```diff
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import List
 
-from starknet_py.net.models.transaction import Transaction
+from starknet_py.net.models import Transaction
```
cptartur, line 4: I'd consider getting rid of imports from `models.__init__.py` completely. Things like this can cause circular import problems.

starknet_py/net/gateway_client.py
cptartur: Problem is parts of starknet.py se `Transaction` and parts use `StarknetTransaction` for the same object, so it's unclear to me how can we resolve this.

starknet_py/net/models/transaction_payloads.py
Solpatium: responses?

starknet_py/net/signer/base_signer.py
Solpatium: https://github.com/software-mansion/starknet.py/issues/475

starknet_py/net/models/transaction_payloads.py
cptartur: I think that's better

starknet_py/net/client_models.py
```diff
@@ -1,355 +1,53 @@
...
-
-@dataclass
-class TransactionStatusResponse:
-    block_hash: Optional[int]
-    transaction_status: TransactionStatus
+warnings.warn(
+    "Importing from starknet_py.net.client_models has been deprecated. "
+    "Use imports from specific modules in starknet_py.net.models.",
+    category=DeprecationWarning,
+)
```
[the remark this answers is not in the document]
Solpatium: Yes, I am sure

cptartur:
I don't know how realistic is resolving the conflicts in this PR. Perhaps a new one should be opened merged as soon as possible to avoid this situation.

PR #485, opened 2022-10-27
Sphinx docs extension

### LINKED ISSUE #378: Create a sphinx extension for parsing and snipping code added to docs

drknzz:
Closes #378 

## What is the current behavior?

Code snippets in docs have to be changed either manually or by running a script.

## What is the new behavior?

Own Sphinx Extension removes the need of specifying line numbers that should be included in the docs.
Instead `:start-after:` and `:end-before:` options can to be specified, that parse and snip the code automatically.

## Other notes

Current example usage in `docs/guide.rst` and `test_using_existing_contracts.py`


docs/_ext/codesnippet.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    Directive class that allows multiple uses of :start-after: and :end-before: options
+    """
+    default_start = "docs: start"
+    default_end = "docs: end"
+
+    has_content = False
+    required_arguments = 1
+    optional_arguments = 0
+    final_argument_whitespace = True
```
war-in, line 45: What are these properties for? maybe docstring again? 😅

docs/_ext/codesnippet.py
```diff
@@ -0,0 +1,107 @@
...
+        filters = [reader.start_filter, reader.end_filter, reader.dedent_filter]
+        result = []
+
+        while True:
+            code_snippet = get_code_snippet(lines, filters, location)
+            if not valid_code_snippet(code_snippet):
+                break
+
+            result.extend(code_snippet)
+            lines = remove_code_snippet(lines, end_marker)
```
war-in, line 74: Is it common practice in the sphinx directives? To remove lines after extracting them?

docs/guide.rst
```diff
@@ -11,14 +11,14 @@ Let's say we have a contract with this interface:
 
 .. literalinclude:: ../starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```
war-in, line 12: Shouldn't it be changed to `.. codesnippet`?

war-in (changes requested):

docs/guide.rst
drknzz: `literalinclude` can be used the same way as `codesnippet` with the exception that it will fetch only the first part of a file between `:start-after:` and `:end-before:` options

war-in (approved):

docs/_ext/codesnippet.py
```diff
@@ -0,0 +1,109 @@
...
+        return node
+
+
+def setup(app) -> Dict[str, Any]:
+    app.add_directive("codesnippet", CodeSnippet)
+
+    return {
+        "version": "0.1",
+        "parallel_read_safe": True,
+        "parallel_write_safe": True,
```
cptartur, line 108: I'm not sure about this, is it really parallel safe?

cptartur (commented):
Just a single comment. I'm not really competent in sphinx enough to say if it's a good directive or bad, but if it works, it seems fine to me. 

I'd wait for @Solpatium review though.

docs/_ext/codesnippet.py
drknzz: The docs say:

```
'parallel_write_safe': a boolean that specifies if parallel writing of output files can be used when the extension is loaded. Since extensions usually don’t negatively influence the process, this defaults to True.

Note
  The parallel-write-safe extension must satisfy the following conditions:
  The core logic of the extension is parallelly executable during the writing phase.
```

This extension behaves the same as e.x. literalinclude would and since it is True there, it's default value is True and it works here when set to True I think it should be okay.

war-in:
Could you change all the literalincludes (or almost all) from the project?

war-in (approved):

PR #482, opened 2022-10-26
Move fixtures to separate files

war-in:
### **Please check if the PR fulfills these requirements**
- [ ] The formatter, linter, and tests all run without an error
- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

## **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Closes #441 

## **What is the current behavior?** (You can also link to an open issue here)

Fixtures are in the two big files which are difficult to read

## **What is the new behavior (if this is a feature change)?**

Fixtures are grouped in the separate files

## **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

No

PR #486, opened 2022-10-27
Fix details_of_account_to_be_deployed

### LINKED ISSUE #465: Make `details_of_account_to_be_deployed` always return new details
There's a problem with `details_of_account_to_be_deployed` fixture: it has a scope `module` and even though it generates a random salt and key, it's not rerun for every tests - if one attempts to do multiple deploy_account transactions in one test module, all but first will fail. 

This is problematic and will need to be addressed in separate PR.

_Originally posted by @cptartur in https://github.com/software-mansion/starknet.py/pull/459#discussion_r995819808_

drknzz:
Closes #465


cptartur (approved):
Always good to have a new factory 😎

I'd wait before #482 is merged though.

PR #499, opened 2022-11-10
Account refactor

### LINKED ISSUE #376: Separate Account from Client

Account should be a separate module from Client and not implement Client interface.

Create `Account` module that only implements account specific methods, deprecate `AccountClient` module and, for the time being, use `Account` inside of `AccountClient`.

none

**Additional context**

`AccountClient` interface is too bloated and this will allow to simplify it.

cptartur:

Closes #376 


starknet_py/net/account/account_proxy.py
```diff
@@ -0,0 +1,125 @@
...
+    EstimatedFee,
+    Hash,
+    Tag,
+)
+from starknet_py.net.models import AddressRepresentation
+from starknet_py.net.models.transaction import DeployAccount
+from starknet_py.net.models.typed_data import TypedData
+
+
+class _AccountProxy(BaseAccount):
```
cptartur, line 20: This is a temporary solution for using old `AccountClient` in code that expects new Account (to avoid multiple implementation in the transition period).

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @abstractmethod
+    async def verify_message(self, typed_data: TypedData, signature: List[int]) -> bool:
+        """
+        Verify a signature of a TypedData TypedDict
+
+        :param typed_data: TypedData TypedDict to be verified
+        :param signature: signature of the TypedData TypedDict
+        :return: true if the signature is valid, false otherwise
+        """
```
Solpatium, line 173: Does all that need to be in the account? 🤔

Solpatium: (use cases for these are quite rare)

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,173 @@
...
+        # TODO docstring
+
+    @property
+    @abstractmethod
+    def supported_tx_version(self) -> int:
+        """"""
+        # TODO docstring
+
+    @abstractmethod
+    async def signed_estimate_fee(
```
Solpatium, line 33: I don't like this method.

cptartur: What alternative do you suggest? We very much need to be able to estimate fee for signed transactions, previously `AccountClient` had an override for `Client.estimate_fee` adding a signature but of course it's not possible now.

Solpatium: User can use for instance `sign_invoke_transaction` and provide its output to Client.estimate_fee, right?

starknet_py/net/account/base_account.py
Solpatium: A TypedData module or something else. I just don't think it is useful here.

starknet_py/net/account/base_account.py
cptartur: Yes, they can do that. I think signed estimate_fee was used in some places in `Contract` but that can be replaced with this kind of flow.

starknet_py/net/account/base_account.py
cptartur: I think it can be done

Solpatium: I think only hash_message can be removed, it is not related to the account.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,401 @@
...
+            nonce=await self.get_nonce(),
+            version=self._version,
+        )
+
+        max_fee = await self._get_max_fee(
+            transaction=declare_tx, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+        declare_tx = dataclasses.replace(declare_tx, max_fee=max_fee)
+        signature = self.signer.sign_transaction(declare_tx)
+        # return dataclasses.replace(declare_tx, signature=signature, max_fee=max_fee)
```
war-in, line 279: 👀

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,401 @@
...
+            signature=[],
+            nonce=0,
+        )
+
+        max_fee = await self._get_max_fee(
+            transaction=deploy_account_tx, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+        deploy_account_tx = dataclasses.replace(deploy_account_tx, max_fee=max_fee)
+        signature = self.signer.sign_transaction(deploy_account_tx)
+        # return dataclasses.replace(deploy_account_tx, signature=signature)
```
war-in, line 320: 👀

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,401 @@
...
+        )
+        try:
+            await self._client.call_contract(call=call, block_hash="pending")
+            return True
+        except ClientError as ex:
+            if re.search(r"Signature\s.+,\sis\sinvalid", ex.message):
+                return False
+            raise ex
+
+    async def _signed_estimate_fee(
```
war-in, line 168: I'm not happy with this name :/

starknet_py/net/account/account_proxy.py
```diff
@@ -0,0 +1,110 @@
...
+    def sign_message(self, typed_data: TypedData) -> List[int]:
+        return self._account_client.sign_message(typed_data=typed_data)
+
+    async def verify_message(self, typed_data: TypedData, signature: List[int]) -> bool:
+        return await self._account_client.verify_message(
+            typed_data=typed_data, signature=signature
+        )
+
+    @staticmethod
+    def account_or_proxy(account: Union[AccountClient, BaseAccount]) -> BaseAccount:
```
war-in, line 105: I can't find the usage of this method. What is it for?

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,150 @@
...
+from starknet_py.net.models.transaction import DeployAccount
+from starknet_py.net.models.typed_data import TypedData
+
+
+class BaseAccount(ABC):
+    @property
+    @abstractmethod
+    def client(self) -> Client:
+        """"""
+        # TODO docstring
```
war-in, line 21: 👀

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @property
+    @abstractmethod
+    def supported_tx_version(self) -> int:
+        """"""
+        # TODO docstring
```
war-in, line 27: 👀

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+
+    @abstractmethod
+    async def sign_invoke_transaction(
+        self,
+        calls: Calls,
+        *,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+        # TODO remove this
```
war-in, line 56: 👀

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @abstractmethod
+    async def sign_declare_transaction(
+        self,
+        compiled_contract: str,
+        *,
+        cairo_path: Optional[List[str]] = None,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+        # TODO remove this
```
war-in, line 76: 👀

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+
+    @abstractmethod
+    async def execute(
+        self,
+        calls: Calls,
+        *,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+        # TODO remove this
```
war-in, line 120: 👀

starknet_py/contract.py
```diff
@@ -220,35 +235,32 @@ async def estimate_fee(
...
             calls=self, max_fee=0, version=self.version
         )
 
-        return await self._account_client.estimate_fee(
+        return await self._client.estimate_fee(
             tx=tx, block_hash=block_hash, block_number=block_number
         )
 
-    @property
-    def _account_client(self) -> AccountClient:
```
war-in, line 232: This is breaking. We should remember about it in the docs

starknet_py/contract.py
```diff
@@ -270,13 +282,18 @@ def prepare(
         :return: PreparedFunctionCall
         """
         if version is None:
+            # version = (
+            #     self._client.supported_tx_version
+            #     if isinstance(self._client, AccountClient)
+            #     else 0
+            # )
+            # TODO add better logic without using internal account
```
war-in, line 290: Is it WIP or something else?

starknet_py/contract.py
```diff
@@ -651,3 +698,24 @@ async def _make_contract_recursively(
...
+    if client is not None:
+        if isinstance(client, AccountClient):
+            return client.client, _AccountProxy(client)
+
+        return client, None
+
+    if account is not None:
+        return account.client, account
+
+    raise ValueError()  # TODO fix typechecker
```
war-in, line 721: Please add docstring here 🙏

war-in (changes requested):

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,401 @@
...
+        )
+
+    async def get_balance(
+        self, token_address: Optional[AddressRepresentation] = None
+    ) -> int:
+        """
+        Checks account's balance of specified token.
+
+        :param token_address: Address of the ERC20 contract.
+            If not specified it will be payment token (wrapped ETH) address.
```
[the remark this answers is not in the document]
cptartur: It was copied from `AccountClient` but we can do that.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,401 @@
...
+
+        return (high << 128) + low
+
+    async def sign_invoke_transaction(
+        self,
+        calls: Calls,
+        *,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+        **kwargs,
```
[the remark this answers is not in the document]
cptartur: This is a workaround I'm not 100% happy about: I want to remove `version` parameter from these methods that was present in `AccountClient` because it doesn't make sense to have it if the account has a version anyway. 

So I've added `**kwargs` so that `version` can still be passed and handled for the time being (only used in `_AccountClient`.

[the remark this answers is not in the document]
cptartur: But perhaps the more future proof approach would be to not have `**kwargs` here and just ignore typechecker in few places it's complaining?

starknet_py/net/account/account.py
cptartur: It should be removed entirely

starknet_py/net/account/account.py
cptartur: It's not great, do you have an idea for an alternative?

starknet_py/net/account/account_proxy.py
cptartur: Leftover of different approach I think. I'll remove it.

starknet_py/net/account/account.py
war-in: Not really, will think about it

starknet_py/contract.py
cptartur: Agreed

There's couple of different breaking changes like the change in the constructors of `ContractFunction` but these aren't probably created by the user.

starknet_py/contract.py
cptartur: As you can see by `TODO` 😅. I forgot about this one, will be changed.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,401 @@
...
+                raise ValueError(
+                    "Either a signer or a key_pair must be provided in AccountClient constructor"
+                )
+
+            chain = chain_from_network(net=client.net, chain=chain)
+            signer = StarkCurveSigner(
+                account_address=self.address, key_pair=key_pair, chain_id=chain
+            )
+        self.signer = signer
+        self._version = 1
```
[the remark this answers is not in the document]
cptartur: In the new `Account` yes, `AccountClient` remains unchanged though.

starknet_py/net/account/account.py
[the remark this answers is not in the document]
war-in: Totally agree. We should design it for the new `Account` not the old `AccountClient`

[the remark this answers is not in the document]
cptartur: Let's do that then,

starknet_py/contract.py
cptartur: Actually, this is final. I've removed commented out code and TODOs.

starknet_py/contract.py
Solpatium: This is a private property.

cptartur: Oh, you're right. So we don't have to worry about it at all 😅

starknet_py/net/account/account.py
Solpatium: Why not just `_estimate_fee`? It is a protected method.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+
+        return (high << 128) + low
+
+    async def sign_invoke_transaction(
+        self,
+        calls: Calls,
+        *,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+    ) -> InvokeFunction:
```
Solpatium, line 224: Perhaps we could create an alias InvokeTransaction for InvokeFunction?

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+        :return: SentTransactionResponse
+        """
+        execute_transaction = await self.sign_invoke_transaction(
+            calls, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+        return await self._client.send_transaction(execute_transaction)
+
+    def sign_message(self, typed_data: TypedData) -> List[int]:
+        """
+        Sign an TypedData TypedDict for off-chain usage with the starknet private key and return the signature
```
Solpatium, line 338: Sometimes a dot is used at the end of a sentence and sometimes it is not.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+        This adds a message prefix, so it can't be interchanged with transactions
+
+        :param typed_data: TypedData TypedDict to be signed
+        :return: The signature of the TypedData TypedDict
+        """
+        return self.signer.sign_message(typed_data, self.address)
+
+    async def verify_message(self, typed_data: TypedData, signature: List[int]) -> bool:
+        """
+        Verify a signature of a TypedData TypedDict
```
Solpatium, line 348: TypedData TypedDict?

Solpatium, line 348: Please add info that it is verified on chain.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+def _add_signature_to_transaction(
+    tx: SignableTransaction, signature: List[int]
+) -> SignableTransaction:
+    return dataclasses.replace(tx, signature=signature)
+
+
+def _merge_calls(calls: Iterable[Call]) -> List:
+    def parse_call(
```
Solpatium, line 371: It seems that this function doesn't have to be nested here, it doesn't use any variables form the clojure.

starknet_py/net/account/account_proxy.py
```diff
@@ -0,0 +1,102 @@
...
+    SentTransactionResponse,
+    Declare,
+    InvokeFunction,
+)
+from starknet_py.net.models import AddressRepresentation
+from starknet_py.net.models.transaction import DeployAccount
+from starknet_py.net.models.typed_data import TypedData
+
+
+class _AccountProxy(BaseAccount):
```
Solpatium, line 17: _ should not be used in class name.

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -117,8 +118,20 @@ def _sign_deploy_account_transaction(self, transaction: DeployAccount) -> List[i
...
         return [r, s]
 
-    def sign_message(self, typed_data: TypedData, account_address: int) -> List[int]:
-        typed_data_dataclass = TypedDataDataclass.from_dict(data=typed_data)
+    def sign_message(
+        self, typed_data: Union[TypedData, TypedDataDataclass], account_address: int
+    ) -> List[int]:
+        if isinstance(typed_data, dict):
+            warnings.warn(
+                "TypedData as dict has been deprecated. Use starknet_py.utils.TypedData dataclass instead.",
```
Solpatium, line 126: `TypedDataDataclass`, but it is a bad name. Can we name it differently if people should use it directly?

starknet_py/contract.py
```diff
@@ -611,12 +648,15 @@ def __init__(
         client: GatewayClient,
         max_steps: int,
         proxy_checks: List[ProxyCheck],
+        account: Optional[BaseAccount] = None,
```
Solpatium, line 651: Shouldn't we use `_unpack_client_and_account` to get Account from AccountClient here and create mutable contract in _make_contract_recursively?

starknet_py/contract.py
```diff
@@ -185,17 +200,25 @@ async def invoke(
...
-            auto_estimate=auto_estimate,
-            version=self.version,
-        )
-        response = await self._account_client.send_transaction(transaction)
+        if isinstance(self._account, _AccountProxy):
+            transaction = await self._account.sign_invoke_transaction(
+                calls=self,
+                max_fee=self.max_fee,
+                auto_estimate=auto_estimate,
+                version=self.version,
```
Solpatium, line 208: What is the purpose of this? The old account had a version provided in constructor, right?

Solpatium (changes requested):

starknet_py/net/account/account.py
cptartur: I think it will just be confusing, since all other methods are excepting `InvokeFunction`.

starknet_py/net/account/account.py
cptartur: `TypedData dict`?

cptartur:
```suggestion
        Verify a signature of a TypedData dict
```
?

starknet_py/net/account/account.py
cptartur: I guess it was just nested because it's only used in this method, but we can changed it.

starknet_py/net/account/account_proxy.py
cptartur: But I want to indicate it's a private class and users should not ever use it (since it may change without any deprecations).

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+        Takes calls and creates InvokeFunction from them
+
+        :param calls: Single call or list of calls
+        :param max_fee: Max amount of Wei to be paid when executing transaction
+        :param auto_estimate: Use automatic fee estimation, not recommend as it may lead to high costs
+        :return: InvokeFunction created from the calls (without the signature)
+        """
+        nonce = await self.get_nonce()
+
+        calldata_py = _merge_calls(ensure_iterable(calls))
```
war-in, line 126: I find this a little confusing :/ `ensure_iterable` returns [calls] if calls are not a list, but maybe we could find a better name?

starknet_py/net/signer/stark_curve_signer.py
cptartur: There's no `TypedDataDataclass`, it's just used as an import alias so we can import both actual TypedData dataclass and a typeddict (used only for typing) some methods.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+        Create and sign declare transaction.
+
+        :param compiled_contract: string containing compiled contract bytecode.
+                                  Useful for reading compiled contract from a file
+        :param cairo_path: a ``list`` of paths used by starknet_compile to resolve dependencies within contracts
+        :param max_fee: Max amount of Wei to be paid when executing transaction
+        :param auto_estimate: Use automatic fee estimation, not recommend as it may lead to high costs
+        :return: Signed Declare transaction
+        """
+        # pylint: disable=too-many-arguments
```
war-in, line 256: Do we still need this?

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+            max_fee=0,
+            signature=[],
+            nonce=await self.get_nonce(),
+            version=self._version,
+        )
+
+        max_fee = await self._get_max_fee(
+            transaction=declare_tx, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+        declare_tx = dataclasses.replace(declare_tx, max_fee=max_fee)
```
war-in, line 272: Can be replaced with `_add_max_fee_to_transaction` like the signature was.
We use it in the `_prepare_invoke_transaction` too

war-in: Or maybe we could create a function which changes any field we want? :hmm:

starknet_py/net/signer/base_signer.py
```diff
@@ -29,12 +30,22 @@ def sign_transaction(self, transaction: Transaction) -> List[int]:
...
+
+             .. deprecated:: 1.0.0
+                TypedData as dict has been deprecated as possible argument.
+                Use :py:class:`starknet_py.utils.TypedData` instead
         :param account_address: account address
         :return: the signature of the JSON object
+
+
         """
+        # TODO use proper deprecated version
```
war-in, line 51: 👀

starknet_py/net/account/account.py
Solpatium: Yes, we would need to change it on other places. It's just InvokeFunction sounds like you can invoke a single function, you know what I mean, right?

starknet_py/net/account/account.py
Solpatium: Sure thing, let's use this opportunity to make sure it is all consistent <3

starknet_py/net/account/account_proxy.py
Solpatium: I know, but it breaks the naming convention. I think it shouldn't be a problem if we don't document it and a comment here.

starknet_py/net/account/account.py
Solpatium: You mean the `ensure_iterable` itself? I think the name is ok actually as it is exactly what it does, but I was probably the one who created it :)

starknet_py/net/signer/stark_curve_signer.py
Solpatium: Hmmm can't we leave it like it was? I think we can accept the dict as we did before, isn't it easier for users?
Also, what is displayed in docs for users? Will it be `Union[TypedData, TypedData]`? 🤔

starknet_py/net/account/account.py
cptartur: I think I do, but perhaps that's for another PR?

starknet_py/net/account/account_proxy.py
cptartur: But if keeping to convention is more important in your opinion, we can change it. I don't have that strong opinion on that matter.

starknet_py/net/signer/stark_curve_signer.py
cptartur: Well it is `Union[TypedData, TypedData]` 😅 so the argument about names makes more sense now. What renaming would you recommend?

About accepting dicts: for signing transactions we're already accepting a constructed tx object, so it seemed to me that'd make more sense to do the same for the typeddata: Using dicts is more for being user friendly in the higher level account interface (that abstracts using signer from you).

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,394 @@
...
+                raise ValueError(
+                    "Either a signer or a key_pair must be provided in AccountClient constructor"
+                )
+
+            chain = chain_from_network(net=client.net, chain=chain)
+            signer = StarkCurveSigner(
+                account_address=self.address, key_pair=key_pair, chain_id=chain
+            )
+        self.signer: BaseSigner = signer
+        self._version = 1
```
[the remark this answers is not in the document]
cptartur: This is the way.

starknet_py/net/account/account.py
cptartur: You mean like `dataclasses.replace` does exactly? But I guess we can abstract it to `_add_max_fee_to_transaction`, keep the steps more obvious.

starknet_py/net/signer/base_signer.py
cptartur: Cannot resolve until we decide what version this change will be (1.0.0 perhaps?)

starknet_py/net/account/account.py
war-in: Maybe it was just me :? Resolved!

starknet_py/net/signer/stark_curve_signer.py
Solpatium: If we want to switch to using typed data then we could type it as dict | TypedData 🤔

starknet_py/net/account/account_proxy.py
Solpatium: There are better ways of doing that. You should prefix the file name with `_` and set `__all__` to empty list (https://peps.python.org/pep-0008/#public-and-internal-interfaces).

starknet_py/net/signer/stark_curve_signer.py
cptartur: I guess we could, we'll be loosing better typing from typeddict we now have, but perhaps we can accept that.

starknet_py/net/account/account_proxy.py
cptartur: I guess we can do that, perhaps in other places as well. Should we create an issue for it?

starknet_py/contract.py
cptartur: Yes, originally I tried adding `**kwargs` to the BaseAccount interface so it would optionally accept `version` but it was weird to have that there. I think we can do with a few `isinstance` checks like this instead.

starknet_py/net/account/account.py
cptartur: Turns out we don't

starknet_py/net/signer/stark_curve_signer.py
Solpatium: Right, but users shouldn't be using the old way anyway

starknet_py/net/account/account_proxy.py
Solpatium: Yes, let's do it like with the keyword only arguments and start here.

starknet_py/contract.py
Solpatium: But why do we need to provide version here?

cptartur: For some reason `AccountClient` accepts optional `version` parameter in methods like `sign_invoke_transaction` even though it also has accepted tx version in `self`. 

We could omit it and rely on the default behavior but perhaps this is safer?

cptartur: Actually it seems that if you set `version=0` on new account using version 1, you'll get a `ValueError` so even more illogical.  I agree we can remove it from Contract.

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -117,8 +117,21 @@ def _sign_deploy_account_transaction(self, transaction: DeployAccount) -> List[i
         r, s = message_signature(msg_hash=tx_hash, priv_key=self.private_key)
         return [r, s]
 
-    def sign_message(self, typed_data: TypedData, account_address: int) -> List[int]:
-        typed_data_dataclass = TypedDataDataclass.from_dict(data=typed_data)
+    def sign_message(
+        self, typed_data: Union[dict, TypedData], account_address: int
```
Solpatium, line 121: I think that dict is a function

Solpatium (commented):
LGTM

war-in (approved):

war-in:
I would mention in the `Account.deploy_account` docstring that while using mainnet or testnet the balance is fetched.
Also, let's fetch the balance before signing the transaction. It will save some time.

cptartur:
> I would mention in the `Account.deploy_account` docstring that while using mainnet or testnet the balance is fetched. Also, let's fetch the balance before signing the transaction. It will save some time.

Checking before signing is only possible if using max_fee. It the user used `auto_estimate` we must first sign to get the max_fee.

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
+X.X.X migration guide
+=======================
+
+TODO(Update this with correct version or move to separate file)
+
+This version deprecates the AccountClient, which is a major change to the StarkNet.py
```
drknzz, line 6:
```suggestion
This version deprecates the :class:`AccountClient <starknet_py.net.account.AccountClient>`, which is a major change to the StarkNet.py
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+It is being replaced with new :class:`BaseAccount <starknet_py.net.account.base_account.BaseAccount>` ABC and its
+default implementation :class:`Account <starknet_py.net.account.account.Account>`.
+
+Unlike ``AccountClient``, an ``Account`` is not a Client anymore. That means that methods like
```
drknzz, line 10:
```suggestion
Unlike ``AccountClient``, an ``Account`` is not a :class:`Client <starknet_py.net.client.Client>` anymore. That means that methods like
```

drknzz, line 10:
```suggestion
Unlike ``AccountClient``, an ``Account`` is not a Client anymore. This means that methods like
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+``get_storage_at``, ``call_contract`` etc. are not available in the Account interface.
+
+However, ``Account`` now exposes a `.`client`` property, which means using an ``Account`` is
```
drknzz, line 13:
```suggestion
However, ``Account`` now exposes a ``.client`` property, which means using an ``Account`` is
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+Replacing inheritance with composition simplifies the ``Account`` interface and will make
+maintaining ``Account`` simpler.
+
+Changes in the Account interface
+--------------------------------
+
+1. Removed ``hash_message`` method. Use ``TypedData.message_hash`` directly instead.
+2. ``Account`` doesn't expose a ``net`` property.
+3. ``Account`` doesn't accept a ``supported_tx_version`` parameter. It currently always uses version 1.
+4. Some parameters liek ``max_fee`` or `auto_estimate`` are now keyword only arguments. They have to be explicitily named like ``account.sign_invoke_transaction(Call(...), max_fee=1000)``. Writing ``account.sign_invoke_transaction(Call(...), 1000)`` will not work.
```
drknzz, line 29:
```suggestion
4. Some parameters like ``max_fee`` or ``auto_estimate`` are now keyword only arguments. They have to be explicitily named like ``account.sign_invoke_transaction(Call(...), max_fee=1000)``. Writing ``account.sign_invoke_transaction(Call(...), 1000)`` will not work.
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+
+
+Deprecatations
```
drknzz, line 32:
```suggestion
Deprecations
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+--------------
```
drknzz, line 33:
```suggestion
------------
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+1. Passing a dict as ``BaseSigner.sign_message`` parameter has been deprecated in favor of :class:`starknet_py.utils.TypedData` dataclass.
+
+
+Breaking changes
+----------------
+
+1. ``version`` parameter has been removed from the most ``Contract`` methods. ``Contract`` will now use version that the ``Account`` or ``AccountClient`` is using.
+2. ``DeclareResult`` now only accepts :class:`BaseAccount <starknet_py.net.account.base_account.BaseAccount>`.
+
+
```
drknzz, line 44: I'd insert here (and on top of all the previous versions' migration guides) two lines with a single pipe sign in each, like:

```
2. ``DeclareResult`` now only accepts :class:`BaseAccount <starknet_py.net.account.base_account.BaseAccount>`.

|
|

0.12.0 Migration guide
======================
```

Currently the migration guides look like a messy pile of changes.
The pipe makes a single new line, so it's cleaner and the guides are separated.

I think we could use even 2 pipes before each guide, rather than 1.

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+    :language: python
+    :dedent: 4
+
+Replacing inheritance with composition simplifies the ``Account`` interface and will make
+maintaining ``Account`` simpler.
+
+Changes in the Account interface
+--------------------------------
+
+1. Removed ``hash_message`` method. Use ``TypedData.message_hash`` directly instead.
```
drknzz, line 26:
```suggestion
1. Removed ``hash_message`` method. Use :meth:`TypedData.message_hash <starknet_py.utils.typed_data.TypedData.message_hash>` directly instead.
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+2. ``Account`` doesn't expose a ``net`` property.
+3. ``Account`` doesn't accept a ``supported_tx_version`` parameter. It currently always uses version 1.
+4. Some parameters liek ``max_fee`` or `auto_estimate`` are now keyword only arguments. They have to be explicitily named like ``account.sign_invoke_transaction(Call(...), max_fee=1000)``. Writing ``account.sign_invoke_transaction(Call(...), 1000)`` will not work.
+
+
+Deprecatations
+--------------
+
+1. Passing a dict as ``BaseSigner.sign_message`` parameter has been deprecated in favor of :class:`starknet_py.utils.TypedData` dataclass.
```
drknzz, line 35:
```suggestion
1. Passing a dict to ``BaseSigner.sign_message`` as parameter has been deprecated in favor of :class:`TypedData <starknet_py.utils.typed_data.TypedData>` dataclass.
```

starknet_py/tests/e2e/docs/migration_guide/test_account_comparision.py
```diff
@@ -0,0 +1,39 @@
...
+    # becomes
+    tx = await account.sign_invoke_transaction(call, max_fee=max_fee)
+    # Note that max_fee is now keyword-only argument
+    await account.client.send_transaction(tx)
+
+    # Using execute method
+    await account_client.execute(call, max_fee)
+    # becomes
+    await account.execute(call, max_fee=max_fee)
+    # docs: end
```
drknzz, line 39:
```suggestion
    # docs-1: start
    # Inspecting storage

    await account_client.get_storage_at(contract_address=address, key=key)

    # becomes

    await account.client.get_storage_at(contract_address=address, key=key)
    # docs-1: end

    call = Call(to_addr=0x1, selector=0x1234, calldata=[])
    max_fee = 1000

    # docs-2: start
    # Sending transactions

    tx = await account_client.sign_invoke_transaction(call, max_fee)
    await account_client.send_transaction(tx)

    # becomes

    tx = await account.sign_invoke_transaction(call, max_fee=max_fee)
    # Note that max_fee is now keyword-only argument
    await account.client.send_transaction(tx)
    #docs-2: end

    #docs-3: start
    # Using execute method

    await account_client.execute(call, max_fee)

    # becomes

    await account.execute(call, max_fee=max_fee)
    # docs-3: end
```

starknet_py/tests/e2e/docs/migration_guide/test_account_comparision.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+from starknet_py.net.client_models import Call
+
+
+@pytest.mark.asyncio
+async def test_account_comparision(
```
drknzz, line 7:
```suggestion
async def test_account_comparison(
```

drknzz: Also, there's a typo in the file name

docs/migration_guide.rst
```diff
@@ -1,10 +1,54 @@
...
+
+Unlike ``AccountClient``, an ``Account`` is not a Client anymore. That means that methods like
+``get_storage_at``, ``call_contract`` etc. are not available in the Account interface.
+
+However, ``Account`` now exposes a `.`client`` property, which means using an ``Account`` is
+just as simple as ``AccountClient`` was. For example:
+
+.. codesnippet:: ../starknet_py/tests/e2e/docs/migration_guide/test_account_comparision.py
+    :language: python
+    :dedent: 4
```
drknzz, line 18:
```suggestion

.. literalinclude:: ../starknet_py/tests/e2e/docs/migration_guide/test_account_comparision.py
    :language: python
    :dedent: 4
    :start-after: docs-1: start
    :end-before: docs-1: end

.. literalinclude:: ../starknet_py/tests/e2e/docs/migration_guide/test_account_comparision.py
    :language: python
    :dedent: 4
    :start-after: docs-2: start
    :end-before: docs-2: end

.. literalinclude:: ../starknet_py/tests/e2e/docs/migration_guide/test_account_comparision.py
    :language: python
    :dedent: 4
    :start-after: docs-3: start
    :end-before: docs-3: end
    
```

drknzz: The file name here still has the typo.

starknet_py/tests/e2e/docs/migration_guide/test_account_comparision.py
cptartur: Didn't know you could do that. Good feature!

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,488 @@
...
+    @property
+    def client(self) -> Client:
+        return self._client
+
+    @property
+    def supported_tx_version(self) -> int:
+        return self._version
+
+    def _get_default_token_address(self) -> str:
+        if self._client.net not in [TESTNET, MAINNET]:
```
war-in, line 92:
```suggestion
        if self._client.net not in [TESTNET, TESTNET2, MAINNET]:
```

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,488 @@
...
+        )
+
+    async def get_balance(
+        self, token_address: Optional[AddressRepresentation] = None
+    ) -> int:
+        """
+        Checks account's balance of specified token.
+
+        :param token_address: Address of the ERC20 contract.
+            If not specified it will be payment token (wrapped ETH) address.
```
war-in, line 210:
```suggestion
            If not specified it will be payment token address.
```

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,488 @@
...
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+    ) -> AccountDeploymentResult:
+        """
+        Deploys an account contract with provided class_hash on StarkNet and returns
+        an AccountDeploymentResult that allows waiting for transaction acceptence.
+
+        Provided address must be first prefunded with enough tokens, otherwise the method will fail.
+
+        If using Client for either TESTNET or MAINNET, this method will verify if the address balance
```
war-in, line 386:
```suggestion
        If using Client for either TESTNET, TESTNET2 or MAINNET, this method will verify if the address balance
```

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,488 @@
...
+        is high enough, to cover deployment costs.
+
+        :param address: calculated and prefunded address of the new account.
+        :param class_hash: class_hash of the account contract to be deployed.
+        :param salt: salt used to calculate the address.
+        :param client: a Client instance used for deployment.
+        :param key_pair: KeyPair used to calculate address and sign deploy account transaction.
+        :param chain: id of the StarkNet chain used.
+        :param constructor_calldata: optional calldata to account contract constructor. If ``None`` is passed,
+            ``key_pair.public_key`` will be used as calldata.
```
war-in, line 396:
```suggestion
            ``[key_pair.public_key]`` will be used as calldata.
```
?

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,488 @@
...
+
+        deploy_account_tx = await account.sign_deploy_account_transaction(
+            class_hash=class_hash,
+            contract_address_salt=salt,
+            constructor_calldata=calldata,
+            max_fee=max_fee,
+            auto_estimate=auto_estimate,
+        )
+
+        if client.net in (TESTNET, MAINNET):
```
war-in, line 432:
```suggestion
        if client.net in (TESTNET, TESTNET2, MAINNET):
```

starknet_py/net/account/base_account.py
```diff
@@ -0,0 +1,159 @@
...
+        cairo_path: Optional[List[str]] = None,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+    ) -> Declare:
+        """
+        Create and sign declare transaction.
+
+        :param compiled_contract: string containing compiled contract bytecode.
+            Useful for reading compiled contract from a file.
+        :param cairo_path: a ``list`` of paths used by starknet_compile to resolve dependencies within contracts.
```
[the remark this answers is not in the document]
cptartur: Nope, I left it by accident

war-in (approved):

starknet_py/net/account/_account_proxy.py
```diff
@@ -0,0 +1,106 @@
...
+            auto_estimate=auto_estimate,
+        )
+
+    async def execute(
+        self,
+        calls: Calls,
+        *,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+        **kwargs,
```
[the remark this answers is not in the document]
cptartur: We shouldn't. I think I forgot to remove it.

starknet_py/net/account/account_deployment_result.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
+
+from starknet_py.contract import SentTransaction
+from starknet_py.utils.sync import add_sync_methods
+
+
+@add_sync_methods
+@dataclass(frozen=True)
+class AccountDeploymentResult(SentTransaction):
+    account: "Account" = None  # pyright: ignore
```
Solpatium, line 10: Does it behave nicely in documentation? I think it was problematic there. 
We can do this instead: https://stackoverflow.com/a/39757388 (__future__.annotations, import based on TYPE_CHECKING).

Solpatium (commented):
I've checked the code without tests

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,486 @@
...
+    ) -> int:
+        """
+        Checks account's balance of specified token.
+
+        :param token_address: Address of the ERC20 contract.
+            If not specified it will be payment token address.
+        :return: Token balance
+        """
+
+        token_address = token_address or self._get_default_token_address()
```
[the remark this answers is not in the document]
cptartur: We can move it, but where do you think it should go?

starknet_py/net/account/account_deployment_result.py
cptartur: I think it's not really documented i.e there's no section for `AccountDeploymentResult`. Should that be changed?

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+def _add_max_fee_to_transaction(
+    tx: SignableTransaction, max_fee: int
+) -> SignableTransaction:
+    return dataclasses.replace(tx, max_fee=max_fee)
+
+
+def _parse_call(
+    _call: Call, _current_data_len: int, _entire_calldata: List
```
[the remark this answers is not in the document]
cptartur: No reason I think, I'll simplify it.

starknet_py/net/account/account_deployment_result.py
cptartur: Added a section in `account.rst`. It looks like this 
image

It's basically the same as current docs of DeployResult and DeclareResult look in Contract section.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,480 @@
...
+        key_pair: KeyPair,
+        client: Client,
+        chain: StarknetChainId,
+        constructor_calldata: Optional[List[int]] = None,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+    ) -> AccountDeploymentResult:
+        """
+        Deploys an account contract with provided class_hash on StarkNet and returns
+        an AccountDeploymentResult that allows waiting for transaction acceptence.
```
drknzz, line 380:
```suggestion
        an AccountDeploymentResult that allows waiting for transaction acceptance.
```

drknzz (changes requested):
🔥

starknet_py/contract.py
```diff
@@ -452,6 +465,7 @@ async def from_address(
         :raises ProxyResolutionError: when given ProxyChecks were not sufficient to resolve proxy's implementation
         :param address: Contract's address
         :param client: Client
+        :param account: BaseAccount
```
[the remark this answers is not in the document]
cptartur: I does it thing? You mean should we use keyword-only arguments? I think we could, it's a bit breaking that's why I haven't enforced it, but perhaps it's a good idea?

starknet_py/contract.py
```diff
@@ -636,3 +660,38 @@ def _create_proxy_config(proxy_config) -> ProxyConfig:
...
+) -> Tuple[Client, Optional[BaseAccount]]:
+    """
+    Get the client and optional account to be used by Contract
+
+    If provided with AccountClient, returns underlying Client and _AccountProxy
+    If provided with Client, returns this Client and None
+    If provided with Account, returns underlying Client and the account
+    """
+    if client is not None and account is not None:
+        raise ValueError("Account and client are mutually exclusive")
```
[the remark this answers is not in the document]
cptartur: We usually used capitalized names in errors even if they were parameter names.

starknet_py/contract.py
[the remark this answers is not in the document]
cptartur: cc @Solpatium @war-in

starknet_py/contract.py
```diff
@@ -636,3 +660,38 @@ def _create_proxy_config(proxy_config) -> ProxyConfig:
...
+        if isinstance(client, AccountClient):
+            return client.client, AccountProxy(client)
+
+        return client, None
+
+    if account is not None:
+        return account.client, account
+
+    # This is needed for typechecker
+    raise ValueError()  # pragma: no cover
```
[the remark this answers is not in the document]
cptartur: I think I found a way to remove the typechecker override that I like a bit better.

starknet_py/contract.py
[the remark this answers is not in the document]
drknzz: Let's decide on a single solution then 🙈 
https://github.com/software-mansion/starknet.py/pull/576#discussion_r1043548968

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,480 @@
...
+        Provided address must be first prefunded with enough tokens, otherwise the method will fail.
+
+        If using Client for either TESTNET, TESTNET2 or MAINNET, this method will verify if the address balance
+        is high enough, to cover deployment costs.
+
+        :param address: calculated and prefunded address of the new account.
+        :param class_hash: class_hash of the account contract to be deployed.
+        :param salt: salt used to calculate the address.
+        :param client: a Client instance used for deployment.
+        :param key_pair: KeyPair used to calculate address and sign deploy account transaction.
```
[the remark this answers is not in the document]
cptartur: I don't think it does but I'll fix it.

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,480 @@
...
+            signature=[],
+            nonce=0,
+        )
+
+        max_fee = await self._get_max_fee(
+            transaction=deploy_account_tx, max_fee=max_fee, auto_estimate=auto_estimate
+        )
+        deploy_account_tx = _add_max_fee_to_transaction(deploy_account_tx, max_fee)
+        signature = self.signer.sign_transaction(deploy_account_tx)
+        return _add_signature_to_transaction(deploy_account_tx, signature)
```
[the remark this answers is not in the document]
cptartur: The duplications are just 4 calls to functions, we can extract it to a separate method, but I'm not sure if it makes sense at this point.

starknet_py/contract.py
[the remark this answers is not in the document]
cptartur: Perhaps let's create an issue for that?

starknet_py/net/account/account.py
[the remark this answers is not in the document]
war-in: Maybe to the `starknet_py/net/networks.py`?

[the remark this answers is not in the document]
cptartur: I think it can be moved there. Perhaps with renaming like `default_token_address_for_network`?

starknet_py/contract.py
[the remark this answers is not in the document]
cptartur: Changed to lowercase

drknzz (approved):
guitar bull 👍

starknet_py/contract.py
```diff
@@ -248,24 +264,24 @@ async def invoke(
...
+
+        response = await self._client.send_transaction(transaction)
 
         invoke_result = InvokeResult(
             hash=response.transaction_hash,  # noinspection PyTypeChecker
-            _client=self._account_client,
+            _client=self._client,
             contract=self._contract_data,
             invoke_transaction=transaction,
         )
```
cptartur, line 287: InvokeResult will now contain a `client` i.e. Gateway or FullNode even if `AccountClient` was used to create a Contract instance. Should this be included in the migration guide?

cptartur (commented):
I've changed `supported_tx_version` to `supported_transaction_version` in BaseAccount and inheritors. 

Also I left a question for reviewers.

starknet_py/contract.py
war-in: It's private property so I don't think so

Solpatium: It is a private property, right? If anyone is using it they are to blame.

starknet_py/net/account/account_deployment_result.py
Solpatium: Hmm I think we should include properties that we want to expose.

starknet_py/contract.py
cptartur: Agreed, not adding it then.

starknet_py/contract.py
```diff
@@ -210,20 +216,30 @@ def __init__(
...
         self.max_fee = max_fee
         self.version = version
 
+    @property
+    def _account(self) -> BaseAccount:
+        if self._internal_account is not None:
+            return self._internal_account
+
+        raise ValueError(
+            "Contract was created without Account provided or with Client that is not an account."
```
Solpatium, line 231:
```suggestion
            "Contract was created without Account or with Client that is not an account."
```

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,418 @@
...
+        auto_estimate: bool = False,
+    ) -> AccountDeploymentResult:
+        """
+        Deploys an account contract with provided class_hash on StarkNet and returns
+        an AccountDeploymentResult that allows waiting for transaction acceptance.
+
+        Provided address must be first prefunded with enough tokens, otherwise the method will fail.
+
+        If using Client for either TESTNET, TESTNET2 or MAINNET, this method will verify if the address balance
+        is high enough, to cover deployment costs.
```
Solpatium, line 325: I think the comma here is unnecessary 🤔

starknet_py/tests/e2e/contract_interaction/interaction_test.py
```diff
@@ -54,40 +53,36 @@ async def test_throws_on_both_max_fee_and_auto_estimate(map_contract):
     value = 3
 
     invocation = map_contract.functions["put"].prepare(key, value)
-    with pytest.raises(ValueError) as exinfo:
+    with pytest.raises(
+        ValueError,
+        match="Max_fee and auto_estimate are exclusive and cannot be provided at the same time.",
```
Solpatium, line 58: I think we wanted to start exceptions with "Arguments x and y...", right?

Solpatium (approved):

starknet_py/tests/e2e/contract_interaction/interaction_test.py
cptartur: Agreed. I can change it, please be aware that this is a another breaking change in Contract.

cptartur: I'll change it in new Account. Errors in Contract are already being addressed in #616 so it doesn't make sense to duplicate this work.

starknet_py/tests/e2e/contract_interaction/interaction_test.py
```diff
@@ -92,7 +92,7 @@ async def test_throws_on_prepared_call_without_max_fee(map_contract):
     prepared_call = map_contract.functions["put"].prepare(key, value)
 
     with pytest.raises(
-        ValueError, match="Max_fee must be specified when invoking a transaction"
+        ValueError, match=r".*[mM]ax_fee must be specified when invoking a transaction"
```
cptartur, line 95: If #616 get's merged before this PR, these can be replaced. This is a temporary fix.

war-in (approved):

starknet_py/net/account/account.py
```diff
@@ -0,0 +1,420 @@
...
+        )
+        declare_tx = _add_max_fee_to_transaction(declare_tx, max_fee)
+        signature = self.signer.sign_transaction(declare_tx)
+        return _add_signature_to_transaction(declare_tx, signature)
+
+    async def sign_deploy_account_transaction(
+        self,
+        class_hash: int,
+        contract_address_salt: int,
+        constructor_calldata: Optional[List[int]] = None,
```
Solpatium, line 262: I think all arguments could be keyword ones

cptartur: I think it's fine to have arguments that will always be present non-keyword. Perhaps `constructor_calldata` could be changed to keyword-only, keeping the rest?

cc @war-in @drknzz

war-in: It's completely up to you. I do not have a strong opinion on that

starknet_py/tests/e2e/contract_interaction/interaction_test.py
war-in: It is merged 👀

starknet_py/contract.py
```diff
@@ -286,68 +327,60 @@ async def estimate_fee(
...
+        self.client = client
+        self.account = account
         self._payload_transformer = FunctionCallSerializer(
             abi=self.abi, identifier_manager=self.contract_data.identifier_manager
         )
 
     def prepare(
         self,
         *args,
-        version: Optional[int] = None,
```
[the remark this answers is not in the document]
Solpatium: I mean usage of keyword argument, which will go to kwargs

docs/migration_guide.rst
```diff
@@ -3,8 +3,8 @@ X.X.X migration guide
 
 TODO(Update this with correct version or move to separate file)
```
war-in, line 4: We can't forget about those todos. I've noticed that the old comment is resolved. Just reminding 😅

starknet_py/contract.py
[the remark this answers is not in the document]
cptartur: Hmm, I guess it would be consumed by args or kwargs and fail in unexpected way. Should we add some sort of handling for that?

starknet_py/tests/e2e/contract_interaction/interaction_test.py
cptartur: Merged development, these should be replaced now.

starknet_py/contract.py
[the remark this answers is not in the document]
Solpatium: There is for sure _some_ handling, data transformer would point that there is an excessive argument. I just wonder if it is descriptive enough for user to know what's happening.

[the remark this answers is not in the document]
cptartur: It's currently if version is provided as positional 

```shell
starknet_py.utils.data_transformer.errors.InvalidTypeException: Provided 2 positional arguments, 1 possible.
````

and for keyword
```shell
starknet_py.utils.data_transformer.errors.InvalidTypeException: Unnecessary named arguments provided: {'version'}.
```

I'd say it's descriptive enough.

starknet_py/net/account/account.py
cptartur: I though about this and I think making these to be keyword only is not very ergonomic for the end user.

PR #504, opened 2022-11-15
Add `abi` to DeclaredContract

### LINKED ISSUE #438: Add `abi` to result of `gateway_client.get_class_by_hash` when RPC v0.2.0 is supported

drknzz:
Closes #438 

- Add structured Abi to RPC Schemas
- Add `abi` member to `DeclaredContract` dataclass
- Adapt `Contract.from_address` to support `FullNodeClient`


starknet_py/net/schemas/rpc.py
```diff
@@ -268,6 +275,79 @@ def make_dataclass(self, data, **kwargs) -> BlockStateUpdate:
...
 
 
+class StructMemberSchema(Schema):
+    name = fields.String(data_key="name", required=True)
+    type = fields.String(data_key="type", required=True)
+    offset = fields.Integer(data_key="offset", required=True)
+
+    @post_load
+    def make_dataclass(self, data, **kwargs) -> StructMember:
+        return StructMember(**data)
```
war-in, line 285: Are we able to test it?

starknet_py/net/schemas/rpc.py
```diff
@@ -268,6 +275,79 @@ def make_dataclass(self, data, **kwargs) -> BlockStateUpdate:
...
+    type = fields.String(data_key="type", required=True)
+    name = fields.String(data_key="name", required=True)
+    size = fields.Integer(data_key="size", required=True)
+    members = fields.List(
+        fields.Nested(StructMemberSchema()), data_key="members", required=True
+    )
+
+    @post_load
+    def make_dataclass(self, data, **kwargs) -> StructAbiEntry:
+        return StructAbiEntry(**data)
```
war-in, line 337: And this?

war-in (approved):

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -134,7 +134,7 @@ async def _get_class_by_address(self, address: Address) -> DeclaredContract:
                 class_hash=contract_class_hash
             )
         except ClientError as err:
-            if "is not deployed" in err.message:
+            if "is not deployed" in err.message or err.code == 28:
```
cptartur, line 137: I think we should extract this code to some constant and declare it at the top of the file (or in other place).

starknet_py/proxy/proxy_check.py
```diff
@@ -64,7 +64,7 @@ async def get_implementation(
             (implementation,) = await client.call_contract(call=call)
             await get_class_func(implementation)
         except ClientError as err:
-            if re.search(err_msg, err.message, re.IGNORECASE):
+            if re.search(err_msg, err.message, re.IGNORECASE) or err.code == 21:
```
cptartur, line 67: Same here

cptartur (commented):
Seems good to me, two questions:

1. Can the dataclasses for the abi also be used for the gateway abi?
2. Is abi returned from the Gateway compatible with things like data transformer?

drknzz:
> 1. Can the dataclasses for the abi also be used for the gateway abi?

Probably yes.
One thing is that abi in RPC is specified in the specs.
However the same cannot be said for gateway abi. Although when looking into cairo-lang I didn't spot any fields that would be possible other than these already included in the dataclasses.

> 2. Is abi returned from the Gateway compatible with things like data transformer?

I don't quite get the question.
It would take a bigger refactor to replace all abi usages with the dataclasses, as modules like data transformer use abi typed vaguely as a list of dicts.

cptartur:
1. We can consider this as possible improvement in the future. We should create a question issue with that.
2. I meant, if I query the Rpc provider for the ABI, can it be then used in our modules like datatransformer? Is the format compatible?

cptartur (approved):

drknzz:
> 2\. I meant, if I query the Rpc provider for the ABI, can it be then used in our modules like datatransformer? Is the format compatible?

If you mean, whether the RPC abi specification is compatible with Gateway abi - yes.
If we cast the RPC abi from dataclass to a list of dicts, it should ordinarily work with modules like DataTransformer.

PR #506, opened 2022-11-16
Simplified deployment flow

war-in:

It adds the possibility to declare and then deploy a contract through the `Contract` class

Users have to use the `Deployer`

Declare and deploy
```
declare_result = await Contract.declare(...)
await declare_result.wait_for_acceptance()
deployment = await declare_result.deploy(...)
await deployment.wait_for_acceptance()
contract = deployment.deployed_contract
```

Deploy
```
deploy_result = await Contract.deploy_contract(...)
await deploy_result.wait_for_acceptance()
contract = deploy_result.deployed_contract
```

No


starknet_py/contract.py
```diff
@@ -112,6 +114,69 @@ def __post_init__(self):
...
+@dataclass(frozen=True)
+class DeclareResult(SentTransaction):
+    class_hash: int = None  # pyright: ignore
+    account: AccountClient = None  # pyright: ignore
+    compiled_contract: str = None  # pyright: ignore
+
+    def __post_init__(self):
+        assert self.class_hash is not None
+        assert self.account is not None
+        assert self.compiled_contract is not None
```
cptartur, line 127: Should we be using asserts in production code?

starknet_py/contract.py
```diff
@@ -415,6 +480,28 @@ async def from_address(
 
         return Contract(address=address, abi=abi, client=client)
 
+    @staticmethod
+    async def declare(
+        account: AccountClient,
+        compiled_contract: str,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+    ) -> DeclareResult:
```
cptartur, line 489:
```suggestion
    async def declare(
        account: AccountClient,
        compiled_contract: str,
        *,
        max_fee: Optional[int] = None,
        auto_estimate: bool = False,
    ) -> DeclareResult:
```
What about this? I like the idea of enforcing optional parameters to be passed as keyword arguments, but it's perhaps inconsistent with the rest of the `Contract`.

starknet_py/contract.py
war-in: Don't know, but `DeployResult` has it 
image

cptartur: Perhaps we should change it here anyway, or maybe even in both places and explicitly do `if ... is not None: raise ValueError`? Seems like a cheap fix.

Arcticae: I guess marshmallow dataclasses would be OK here as well, to validate those fields.

starknet_py/contract.py
```diff
@@ -112,6 +114,69 @@ def __post_init__(self):
...
+        unique: bool = True,
+        constructor_args: Optional[Union[List, Dict]] = None,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
+    ) -> "DeployResult":
+        """
+        Deploys a contract
+
+        :param deployer_address: Address of the UDC. Is set to the address of
+            the default UDC (same address on real nets and devnet) by default.
```
Arcticae, line 143:
```suggestion
            the default UDC (same address on testnet and devnet) by default.
```

Arcticae, line 143:
```suggestion
            the default UDC (same address on mainnet/testnet/devnet) by default.
```

starknet_py/contract.py
```diff
@@ -112,6 +114,69 @@ def __post_init__(self):
...
+            Must be set when using custom network other than devnet.
```
Arcticae, line 144:
```suggestion
            Must be set when using custom network other than ones listed above.
```

starknet_py/contract.py
```diff
@@ -112,6 +114,69 @@ def __post_init__(self):
...
+        :param salt: Optional salt. Random value is selected if it is not provided.
+        :param unique: Determines if the contract should be salted with the account address.
+        :param constructor_args: a ``list`` or ``dict`` of arguments for the constructor.
+        :param max_fee: Max amount of Wei to be paid when executing transaction.
+        :param auto_estimate: Use automatic fee estimation, not recommend as it may lead to high costs.
```
Arcticae, line 149:
```suggestion
        :param auto_estimate: Use automatic fee estimation (not recommended, as it may lead to high costs).
```

starknet_py/contract.py
Solpatium: Why can't we make those fields required, without default values = None?

cptartur: We've kind of decided agains using marshmallow dataclasses as the package is not very well maintained. We could use marshmallow schemas here, but it seems kind of overkill to me.

starknet_py/contract.py
```diff
@@ -112,14 +114,80 @@ def __post_init__(self):
 InvocationResult = InvokeResult
 
 
+@add_sync_methods
+@dataclass(frozen=True)
+class DeclareResult(SentTransaction):
+    class_hash: int = None  # pyright: ignore
+    account: AccountClient = None  # pyright: ignore
```
Solpatium, line 121: account should be private here

starknet_py/contract.py
```diff
@@ -415,6 +483,81 @@ async def from_address(
...
+        return DeclareResult(
+            hash=res.transaction_hash,
+            _client=account.client,
+            class_hash=res.class_hash,
+            account=account,
+            compiled_contract=compiled_contract,
+        )
+
+    @staticmethod
+    async def deploy_contract(
```
Solpatium, line 519: We should mention it in `.deploy`below.

Solpatium, line 519: `declare_class` and `deploy_from_class`? I think this suggests logical flow of operations.

starknet_py/contract.py
```diff
@@ -415,6 +483,81 @@ async def from_address(
...
+        class_hash: Hash,
+        account: AccountClient,
```
Solpatium, line 521: We should take account as the first parametr, like above.

starknet_py/contract.py
```diff
@@ -415,6 +483,81 @@ async def from_address(
...
+        abi: List,
+        constructor_args: Optional[Union[List, Dict]] = None,
+        *,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
```
Solpatium, line 526: We should also take optional parameter for UDC address.

Solpatium (changes requested):
Good job <3

starknet_py/contract.py
Solpatium: @cptartur @Arcticae what do you think?

starknet_py/contract.py
```diff
@@ -415,6 +483,81 @@ async def from_address(
...
+    ) -> "DeployResult":
```
Solpatium, line 527: What if we offered `prefunded_deploy` flag here to deploy using deploy account? 🤔

Solpatium: Or a seperate method, `deploy_prefunded` for both Contract & DeclareResult?

war-in: `deploy_prefunded` would be better. However, I would make it in a separate PR

cptartur: Let's have it as a separate method, it goes agains "simplified" to have many possible flows in single method.

starknet_py/contract.py
cptartur: I think `declare_class` is unecessary, since we have `declare` for a while now and it's kind of obvious what it does. 

`deploy_from_class` seems kind of redundant - all kinds of deploys will be based on class_hash soon so it doesn't make sense to specify that.

starknet_py/contract.py
```diff
@@ -415,6 +485,87 @@ async def from_address(
...
+            compiled_contract=compiled_contract,
+        )
+
+    @staticmethod
+    async def deploy_contract(
+        account: AccountClient,
+        class_hash: Hash,
+        abi: List,
+        constructor_args: Optional[Union[List, Dict]] = None,
+        deployer_address: AddressRepresentation = DEFAULT_DEPLOYER_ADDRESS,
```
cptartur, line 526: deployer_address should be keyword only

cptartur: So below the `*`

starknet_py/contract.py
Solpatium: But do we plan to remove .deploy before this version is deployed?

cptartur: I think so, yes. Deploy will most likely removed soon anyways and it's been deprecated for like good 1.5 month now.

starknet_py/contract.py
war-in: https://github.com/software-mansion/starknet.py/issues/515

cptartur (approved):

Solpatium (approved):

PR #511, opened 2022-11-17
Improve ContractNotFoundError message

### LINKED ISSUE #505: Improve error message for when contract is not found

I am currently facing a problem that results in this error:

```
starknet_py.net.client_errors.ContractNotFoundError: Client failed: No contract found for identifier: pending
```

It would be helpful for me to know something about the contract, like perhaps its class hash? For debugging purposes.
Also the word before "identifier" should maybe say "block" because it sounds like "pending" is referring to a contract identifier but I am pretty sure it is a block, but it might not be obvious enough.

```
starknet_py.net.client_errors.ContractNotFoundError: Client failed: No contract with hash 1234123125 found for block identifier: pending
```

I thought about this for 15 seconds and I am not sure exactly but will keep it in mind, thanks.

**Additional context**

drknzz:
Closes #505


war-in (approved):

starknet_py/net/client_errors.py
```diff
@@ -19,17 +22,16 @@ class ContractNotFoundError(ClientError):
...
+        is_identifier = block_hash is not None or block_number is not None
         identifier = block_hash or block_number
-        self.identifier = str(identifier) if isinstance(identifier, int) else identifier
-
-        super().__init__(message=f"No contract found for identifier: {self.identifier}")
+        identifier_name = "block_hash" if block_hash else "block_number"
 
+        message = f"No contract with address {address} found"
+        block_info = f" for block with {identifier_name}: {identifier}"
+        full_message = f"{message}{block_info if is_identifier else ''}"
```
cptartur, line 35:
```suggestion
        block_info = f" for block with {identifier_name}: {identifier}" if is_identifier else ""
        full_message = message + block_info
```

starknet_py/net/client_errors.py
```diff
@@ -19,17 +22,16 @@ class ContractNotFoundError(ClientError):
...
+        block_number: Optional[Union[int, Tag]] = None,
     ):
-        require_block_identifier(block_hash, block_number)
-
+        is_identifier = block_hash is not None or block_number is not None
         identifier = block_hash or block_number
-        self.identifier = str(identifier) if isinstance(identifier, int) else identifier
-
-        super().__init__(message=f"No contract found for identifier: {self.identifier}")
+        identifier_name = "block_hash" if block_hash else "block_number"
```
cptartur, line 31: I wonder if we should also handle "latest" and "pending, since we're doing a bit of extra logic with these: In the GatewayClient if someone uses parameter `block_hash="latest"`, we're actually making a request with `{"block_number": "latest}`.

cptartur (commented):
I think we could extend the usage of ContractNotFoundErrror to the FullNodeClient as well, and perhaps document it in the `Client` interface itself.

starknet_py/net/client_errors.py
drknzz: Well, it's a bug on our side. `block_hash` hash to be an actual hash and not any literal, as we allow in many places in code.
But I guess it'd be a small breaking change and it's not that harmful to leave it.

drknzz:
> I think we could extend the usage of ContractNotFoundErrror to the FullNodeClient as well, and perhaps document it in the `Client` interface itself.

Well, currently only one **gateway** endpoint is wrapped in the exception.
The list of client methods where it could be thrown:
- get_class_hash_at
- call_contract
- get_nonce - can't really know, because it returns 0 as default

And gateway specific:
- get_code - returns empty bytecode as default

If we want to wrap the responses with `"code": StarknetErrorCode.UNINITIALIZED_CONTRACT` into `ContractNotFoundError`, we should do it for all of the methods probably.

But also, I feel like if we would want to use our exceptions in these methods, we should also consider using other exceptions corresponding to the errors that can occur, some of them specified in the [RPC specs](https://github.com/starkware-libs/starknet-specs/blob/v0.2.1/api/starknet_api_openrpc.json#L1910).

cptartur:
We could definitely include custom exceptions in other methods as well. My point is it's little bit inconsistent to have ContractNotFoundErrror only in one implementation of the interface so addresing this should be higher priority (could be a separate PR) than extending it to other methods.

PR #522, opened 2022-11-21
Remove `AccountClient.create_account`

war-in:


starknet_py/tests/e2e/docs/guide/test_using_cairo_serializer.py
```diff
@@ -64,7 +47,7 @@ async def test_using_cairo_serializer(network, gateway_account_client):
     await invoke_result.wait_for_acceptance()
 
     transaction_hash = invoke_result.hash
-    transaction_receipt = await client.get_transaction_receipt(transaction_hash)
+    transaction_receipt = await account_client.get_transaction_receipt(transaction_hash)
```
cptartur, line 50: Can we use `client` like it was previously?

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```diff
@@ -26,40 +25,24 @@
...
-    from starknet_py.net.models import StarknetChainId
 
     address = "0x00178130dd6286a9a0e031e4c73b2bd04ffa92804264a25c1c08c1612559f458"
-    client = GatewayClient(TESTNET)
-    # docs: end
-    client = gateway_client
-    # docs: start
 
-    contract = Contract(address=address, abi=abi, client=gateway_client)
+    contract = Contract(address=address, abi=abi, client=account_client)
```
cptartur, line 35: Lets add a comment that this is a preferred method.

starknet_py/tests/e2e/fixtures/contracts.py
```diff
@@ -155,30 +158,42 @@ def fixture_balance_contract() -> str:
...
+    await account.wait_for_tx(resp.transaction_hash)
 
     return resp.class_hash
 
 
+@pytest_asyncio.fixture(scope="module")
+async def account_with_validate_deploy_class_hash(
+    pre_deployed_account_with_validate_deploy: AccountClient,
+) -> int:
+    compiled_contract = read_contract("account_with_validate_deploy_compiled.json")
```
cptartur, line 182: Why is it not a constant like in the method below?

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
Solpatium: And let's include info that this way we save ourselves additional request to fetch the ABI.

cptartur: Isn't it a bit too technical for a (rather simple) guide? I mean, we can include it but I wonder if we should.

starknet_py/tests/e2e/docs/guide/test_using_cairo_serializer.py
war-in: That's a tricky one because previously client was an AccountClient 😅 (check out a few previous lines)

starknet_py/tests/e2e/fixtures/contracts.py
war-in: Because we have a contract for the new account but do not have one for the old one. The old account is compiled and we keep it as a constant

starknet_py/tests/e2e/fixtures/account_clients.py
```diff
@@ -156,12 +170,33 @@ async def new_devnet_account_details(
...
+        supported_tx_version=1,
+    )
+    res = await account.deploy_account(deploy_account_tx)
+    await account.wait_for_tx(res.transaction_hash)
+
     return hex(address), hex(key_pair.private_key)
 
 
 @pytest_asyncio.fixture(scope="module")
 async def new_address_and_private_key(
```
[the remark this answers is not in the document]
war-in: I'm not sure about this one. Could you explain?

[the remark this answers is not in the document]
drknzz: I meant to maybe somehow use the fixture you added below in this fixture, similarly to how you used it in the other fixture, without `new_` in name

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
Solpatium: This gives user a reason for doing it **_properly_**

PR #546, opened 2022-11-29
Add  `Account.deploy_account` simplified flow

### LINKED ISSUE #515: Add the `?.deploy_prefunded` method to ease account deployment
I've added `?` in the title because I'm unsure if it should be in the `Contract` or in the `Account`.

cptartur:

Closes #515 


starknet_py/net/account/account.py
```diff
@@ -359,6 +360,57 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+            computed := compute_address(
+                salt=salt,
+                class_hash=class_hash,  # class_hash of the Account declared on the StarkNet
+                constructor_calldata=[key_pair.public_key],
+                deployer_address=0,
+            )
+        ):
+            raise ValueError(
+                f"Provided address {hex(address)} is different than computed address {hex(computed)} "
+                f"for the given class_hash and salt"
```
Solpatium, line 398: Why do we accept address when we can calculate it ourselves?

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,57 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+            address=address, client=client, key_pair=key_pair, chain=chain
+        )
+        deploy_account_tx = await account.sign_deploy_account_transaction(
+            class_hash=class_hash,
+            contract_address_salt=salt,
+            constructor_calldata=[key_pair.public_key],
+            max_fee=max_fee,
+        )
+        result = await client.deploy_account(deploy_account_tx)
+        await client.wait_for_tx(result.transaction_hash)
```
Solpatium, line 411: I don't like waiting for txs at all. In case of mainnet it can take minutes. I suggest doing similar flow as with deploying regular contracts.

Solpatium (changes requested):

starknet_py/net/account/account.py
cptartur: I wanted to have an "extra check" here, that the user in fact prefunded the correct address for the given class hash and salt. If they accidentally prefunded different than the one we want to deploy in this method, this error would tell them.

starknet_py/net/account/account.py
cptartur: This is a helper method to ease onboarding new accounts i.e. user creating their first account. I think they'd want to wait for their account to be usable anyway. If someone wants more flexibility they should use UDC, Contract or create deploy_account transactions themselves. This is not the point of this method.

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,57 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+        client: Client,
+    ) -> "Account":
+        """
+        Deploys an account contract with provided class_hash on starknet and returns
+        a Account instance to be used.
+
+        Provided address must be first prefunded with enough tokens, otherwise the method will fail.
+
+        :param address: calculated and prefunded address of the new account.
+        :param class_hash: class_hash of the account contracat to be deployed.
```
drknzz, line 381:
```suggestion
        :param class_hash: class_hash of the account contract to be deployed.
```

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,57 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+        address: int,
+        class_hash: int,
+        salt: int,
+        key_pair: KeyPair,
+        chain: StarknetChainId,
+        max_fee: int,
+        client: Client,
+    ) -> "Account":
+        """
+        Deploys an account contract with provided class_hash on starknet and returns
```
drknzz, line 375:
```suggestion
        Deploys an account contract with provided class_hash on StarkNet and returns
```

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,57 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+            )
+
+        account = Account(
+            address=address, client=client, key_pair=key_pair, chain=chain
+        )
+        deploy_account_tx = await account.sign_deploy_account_transaction(
+            class_hash=class_hash,
+            contract_address_salt=salt,
+            constructor_calldata=[key_pair.public_key],
+            max_fee=max_fee,
```
drknzz, line 408: `max_fee` in `AccountClient.sign_deploy_account_transaction` is `Optional`, with the possibility to use `auto_estimate`
Don't we want to maybe use it here as well?

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,57 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+        :param salt: salt used to calculate the address.
+        :param client: a Client instance used for deployment.
+        :param key_pair: KeyPair used to calculate address and sign deploy account transaction.
+        :param chain: id of the StarkNet chain used.
+        :param max_fee: max fee to be paid for deployment, must be less or equal to the amount of tokens prefunded.
+        """
+        if address != (
+            computed := compute_address(
+                salt=salt,
+                class_hash=class_hash,  # class_hash of the Account declared on the StarkNet
```
drknzz, line 391: Is this comment necessary?

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,60 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+                f"for the given class_hash and salt"
+            )
+
+        account = Account(
+            address=address, client=client, key_pair=key_pair, chain=chain
+        )
+        deploy_account_tx = await account.sign_deploy_account_transaction(
+            class_hash=class_hash,
+            contract_address_salt=salt,
+            constructor_calldata=[key_pair.public_key],
```
war-in, line 409: We do not support all accounts here. Maybe we should allow passing constructor_calldata as an argument and default to this one?

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,60 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
         message_hash = typed_data_dataclass.message_hash(account_address=self.address)
         return await self._verify_message_hash(message_hash, signature)
 
+    @staticmethod
+    async def deploy_account(
+        *,
+        address: int,
+        class_hash: int,
```
war-in, line 367: Should we allow for a string class_hash? My point is that in Starkscan or Voyager `class_hashes` are displayed as hex values. New users will take them and be confused why we require int here

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,60 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+        salt: int,
+        key_pair: KeyPair,
+        client: Client,
+        chain: StarknetChainId,
+        max_fee: Optional[int] = None,
+        auto_estimate: bool = False,
```
war-in, line 373: I saw the @drknzz comment, but I'm unsure if we should use `auto_estimate` here. It is not a good idea to use it while deploying an account. Also maybe `AccountClient.sign_deploy_account_transaction` shouldn't have an `auto_estimate` argument too?

starknet_py/net/account/account.py
Solpatium: Python converts hex literals to integers

starknet_py/net/account/account.py
Solpatium: They would be using the same compute_address function to compute it in the first place. If they provide wrong address they will get an error on chain. 
I would rather check if the address they want to use is prefunded - this way we can inform users what has to be done first. 

I am also wondering how to make address computation easier, it is not obvious for users how to use `compute_address`. Maybe a flow when you prepare the deployment and get the address before the deployment?

starknet_py/net/account/account.py
Solpatium: If users use this method in case of some network malfunction they will be left with a long wait, wondering what's going on. There will be no way of knowing the tx hash that they can check later. I suggest doing flow similar as with contracts - you get a tx that you can wait for with an additional account field.

starknet_py/net/account/account.py
Solpatium: That's true - we don't have any class hash hardcoded here. In reality you want to deploy account behind the proxy. At minimum it will take both the public key and class hash.

starknet_py/net/account/account.py
cptartur: I discussed this flow with @war-in at first, and while it seemed nice, it's not very ergonomic. It would require the users to re-run their scripts anyways: They'd have to do the first step to get address and prefund it, and then run first and second step again after they prefunded.

starknet_py/net/account/account.py
cptartur: I get your point about the tx hash and I think there should be some way to retrieve it. However, this method is meant as kind of replacement for `AccountClient.create_account` for the less advanced users - I think that adding extra steps will increase confusion.

starknet_py/net/account/account.py
```diff
@@ -359,6 +360,60 @@ async def verify_message(self, typed_data: TypedData, signature: List[int]) -> b
...
+        if address != (
+            computed := compute_address(
+                salt=salt,
+                class_hash=class_hash,
+                constructor_calldata=[key_pair.public_key],
+                deployer_address=0,
+            )
+        ):
+            raise ValueError(
+                f"Provided address {hex(address)} is different than computed address {hex(computed)} "
```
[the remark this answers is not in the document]
cptartur: Yeah, I initially used int, but it kind of seemed weird printing addresses in that. I can change it back however if you think it makes more sense.

starknet_py/net/account/account.py
cptartur: Okay, we can accept any calldata.

starknet_py/net/account/account.py
Solpatium: So maybe a method that allows you to calculate the address using same arguments as in here?

starknet_py/net/account/account.py
Solpatium: I think that if documentation uses `wait_for_tx` there shouldn't be any confusion (people will just copy it to their codebases) and we'll avoid questions like "Why does it take so long to deploy account using starknet.py?".

starknet_py/net/account/account.py
cptartur: We already have that in `net.models.addres.compute_address`. It is also outlined in the guide about deploying accounts.

Solpatium: And wheat about checking if address is prefunded? This way we can have a very nice error message.

cptartur: This I think can be done and I agree it would produce a nicer flow.

starknet_py/net/account/account.py
cptartur: Okay, I'll try changing it up and we'll see if it's simple enough.

starknet_py/net/account/account.py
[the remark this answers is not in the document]
war-in: If the `address` will be of type `Union[int, str]` it is ok to print it as a hex (hex is more readable)

starknet_py/net/account/account.py
cptartur: I've added a simple balance check, you can take a look.

Potential problem with this check: if someone is using a different network than testnet/mainnet, we're unable to check the balance, because the token address is unknown to us.

cptartur:
Thing not addressed by this PR: tests of `Account.deploy_account`. Will be done in the separate PR.

starknet_py/net/account/account.py
Solpatium: Why isn't this a good idea to use it for deploying your account? I would say sending invokes with this flag is more dangerous.
At the end of the day you can't lose more than what you sent to your account.

Solpatium (approved):
Great 👍

war-in (approved):

PR #548, opened 2022-11-29
Add Invoke alias for InvokeFunction

### LINKED ISSUE #525: Alias `InvokeFunction` to `Invoke`
Perhaps we could create an alias InvokeTransaction for InvokeFunction?

_Originally posted by @Solpatium in https://github.com/software-mansion/starknet.py/pull/499#discussion_r1026413147_

drknzz:
Closes #525 

- Add `Invoke` alias
- Ignore pylint's `unused-import` for backward compatibility

- Fix error `"stubPath /.../starknet.py/typings is not a valid directory."` when running typechecker


pyproject.toml
```diff
@@ -101,3 +101,4 @@ exclude = [
     "**/__pycache__",
     "starknet_py/tests/e2e/docs",
 ]
+stubPath = ""
```
cptartur, line 104: Why is that needed?

starknet_py/net/account/account_client.py
```diff
@@ -636,7 +633,7 @@ def parse_call(
     return [calldata, entire_calldata]
 
 
-def make_invoke_function_by_version(
+def make_invoke_by_version(
```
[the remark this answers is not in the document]
cptartur: I don't think so this should be consider a private function anyways.

cptartur (commented):
Looks good.

I wonder if there's a way to add some kind of deprecation warning if someone uses `InvokeFunction` but that probably would be complicated.

pyproject.toml
drknzz: The 3rd point in PR description:

- Fix error `"stubPath /.../starknet.py/typings is not a valid directory."` when running typechecker

The error does nothing bad, just annoying to see.

cptartur: Interesting, I never noticed this error but if it fixes it, I guess we can keep it.

cptartur: cc @Solpatium @war-in

drknzz: https://github.com/microsoft/pyright/issues/777

drknzz:
> I wonder if there's a way to add some kind of deprecation warning if someone uses `InvokeFunction` but that probably would be complicated.

This exists: https://peps.python.org/pep-0562/#rationale

pyproject.toml
war-in: It's OK for me but please add a comment about why we need that

war-in (approved):
All good, just one question. Should we create an issue to remove the `InvokeFunction,  # backward compatibility` one day?

starknet_py/net/account/account_client.py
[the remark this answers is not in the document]
Solpatium: Please prefix it with `_` to mark that it is private.

starknet_py/net/full_node_client.py
```diff
@@ -7,12 +7,13 @@
...
 from starknet_py.net.client_errors import ClientError
-from starknet_py.net.client_models import (
+from starknet_py.net.client_models import (  # pylint: disable=unused-import
     SentTransactionResponse,
     TransactionReceipt,
     BlockStateUpdate,
     StarknetBlock,
-    InvokeFunction,
+    InvokeFunction,  # backward compatibility
+    Invoke,
```
Solpatium, line 16: I am not sure if we need to leave `InvokeFunction` imports in every place. If someone imported it from a wrong place they should adjust their import paths. What do you guys think?

Solpatium: @cptartur @war-in

cptartur: I think it makes sense to do remove them.

Solpatium (approved):
🔥

PR #559, opened 2022-12-02
Fix estimating fee on prepared calls

### LINKED ISSUE #556: [BUG] Docs about fee estimation
### What happened

https://starknetpy.readthedocs.io/en/latest/guide.html#fee-estimation

The code example is wrong. We can't pass `max_fee` to the `estimate_fee`.

**We should write better docs about fee estimation**

### Stack trace

not applicable

### Steps to reproduce

1. Try to do `estimate_fee` on the `PreparedFuctionCall` with `max_fee != 0`
2. Get an error: `Cannot estimate fee of PreparedFunctionCall with max_fee not None or 0.`

### SDK Version

The newest one

### Python version

3.9.13

Mac

drknzz:
Closes #556 

## Introduced changes

Instead of changing the documentation as suggested in the [issue](https://github.com/software-mansion/starknet.py/issues/556), removed the check for `max_fee`. 
- `ValueError` is not raised when `max_fee` is given
- Raising the error there doesn't make much sense since it's a fee estimation of an isolated prepared function call object (so the `max_fee` in the call is not tied to the method in any way)

##

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -284,13 +284,7 @@ class PreparedFunctionCall(Call):
         :param block_number: Estimate fee at given block number
             (or "latest" / "pending" for the latest / pending block), default is "pending"
         :return: Estimated amount of Wei executing specified transaction will cost
-        :raises ValueError: when max_fee of PreparedFunctionCall is not None or 0.
         """
-        if self.max_fee is not None and self.max_fee != 0:
-            raise ValueError(
-                "Cannot estimate fee of PreparedFunctionCall with max_fee not None or 0."
-            )
-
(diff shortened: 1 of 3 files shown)


cptartur (approved):
I think we can do that. There shouldn't be any technical reason for disallowing estimating txs with fee > 0.

war-in:
There is no technical reason, just UX. In my opinion, it is not OK to receive an `EstimatedFee.overall_fee` higher than the `max_fee` set during preparation.

war-in:
Ok then. I do not have a strong opinion on that. We can leave it as it is.

war-in (approved):

PR #560, opened 2022-12-02
Add `Account.deploy_account` tests

cptartur:
## Introduced changes

- Adds tests for `Account.deploy_account` method

##

### DIFF
--- a/starknet_py/tests/e2e/account/account_test.py
+++ b/starknet_py/tests/e2e/account/account_test.py
@@ -1,9 +1,18 @@
+from unittest.mock import patch, AsyncMock
+
 import pytest
 from starkware.starknet.public.abi import get_selector_from_name
 
 from starknet_py.contract import Contract
-from starknet_py.net.client_models import Call
+from starknet_py.net.account.account import Account
+from starknet_py.net.account.base_account import BaseAccount
+from starknet_py.net.client_models import (
+    Call,
+    TransactionStatus,
+    DeployAccountTransaction,
+)
 from starknet_py.net.gateway_client import GatewayClient
+from starknet_py.net.models import StarknetChainId, compute_address
 from starknet_py.tests.e2e.fixtures.constants import MAX_FEE
 from starknet_py.transaction_exceptions import TransactionRejectedError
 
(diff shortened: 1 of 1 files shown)


starknet_py/tests/e2e/account/account_test.py
```diff
@@ -178,3 +187,138 @@ async def test_sign_deploy_account_transaction(gateway_account):
...
+        "starknet_py.net.gateway_client.GatewayClient.call_contract", AsyncMock()
+    ) as mocked_balance:
+        mocked_balance.return_value = (0, 100)
+
+        await Account.deploy_account(
+            address=address,
+            class_hash=class_hash,
+            salt=salt,
+            key_pair=key_pair,
+            client=GatewayClient(net="testnet"),
```
war-in, line 292: same as above

war-in (approved):

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -178,3 +187,138 @@ async def test_sign_deploy_account_transaction(gateway_account):
...
+        with pytest.raises(
+            ValueError,
+            match="Not enough tokens at the specified address to cover deployment costs",
+        ):
+            await Account.deploy_account(
+                address=address,
+                class_hash=class_hash,
+                salt=salt,
+                key_pair=key_pair,
+                client=GatewayClient(net="testnet"),
```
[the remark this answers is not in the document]
cptartur: Yes, but if I used `network` fixture, the test wouldn't work as expected. `Account.deploy_account` only verifies the balance if the net is either `TESTNET` or `MAINNET`: Token address is unknown on a custom networks so we cannot verify the balance.

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -178,3 +187,138 @@ async def test_sign_deploy_account_transaction(gateway_account):
...
+        max_fee=MAX_FEE,
+    )
+    _, status = await account.client.wait_for_tx(
+        res.transaction_hash, wait_for_accept=True
+    )
+
+    assert status in (
+        TransactionStatus.ACCEPTED_ON_L1,
+        TransactionStatus.ACCEPTED_ON_L2,
+    )
```
Solpatium, line 230: Do we have more code like this in the code? I think we could use a custom `assert_tx_passed(tx_hash)` for something like this.

Solpatium (approved):
LGTM

starknet_py/tests/e2e/account/account_test.py
cptartur: `wait_for_accept` especially when you add a `wait_for_accept=True` parameter basically asserts that so in most places we don't check the status of transaction (tests would fail if it were rejected). I added status assert here just to be extra sure.

PR #566, opened 2022-12-05
Add cairo type parser

Solpatium:
Part of the effort to refactor data transformer.

### DIFF
--- a/starknet_py/cairo/type_parser.py
+++ b/starknet_py/cairo/type_parser.py
@@ -50,7 +50,7 @@ class TypeParser:
     def parse_inline_type(self, type_string: str) -> CairoType:
         """
         Inline type it one that can be used inline. For instance (a: Uint256, b: felt*, c: (felt, felt)).
-        
+
         :param type_string: type to parse
         """
         parsed = parse_type(type_string)
--- a/starknet_py/cairo/type_parser.py
+++ b/starknet_py/cairo/type_parser.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from collections import OrderedDict
-from typing import Dict, Optional
+from typing import Dict
 
 from starkware.cairo.lang.compiler.parser import parse_type
 import starkware.cairo.lang.compiler.ast.cairo_types as cairo_lang_types
--- a/starknet_py/cairo/type_parser.py
+++ b/starknet_py/cairo/type_parser.py
@@ -62,8 +62,8 @@ class TypeParser:
         self, cairo_type: cairo_lang_types.CairoType
     ) -> CairoType:
         """
-        For now, we use parse function from cairo-lang pacakge. It will be replaced in the future, but we need to hide it
-        from the users.
+        For now, we use parse function from cairo-lang pacakge. It will be replaced in the future, but we need to hide
+        it from the users.
         This function takes types returned by cairo-lang package and maps them to our type classes.
 
         :param cairo_type: type returned from parse_type function.
--- a/starknet_py/cairo/type_parser.py
+++ b/starknet_py/cairo/type_parser.py
@@ -1,0 +1,100 @@
+from __future__ import annotations
+from collections import OrderedDict
+from typing import Dict, Optional
+
+from starkware.cairo.lang.compiler.parser import parse_type
+import starkware.cairo.lang.compiler.ast.cairo_types as cairo_lang_types
+
+from starknet_py.cairo.data_types import (
+    CairoType,
+    FeltType,
+    TupleType,
+    NamedTupleType,
+    ArrayType,
+    StructType,
+)
+
+
+class UnknownTypeError(ValueError):
+    """
+    Error thrown when TypeParser finds type that was not declared prior to parsing.
+    """
+
+    type_name: str
+
+    def __init__(self, type_name: str):
+        super().__init__(f"Type '{type_name}' is not defined")
+        self.type_name = type_name
+
+
+class TypeParser:
+    """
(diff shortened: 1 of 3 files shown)


starknet_py/cairo/type_parser.py
```diff
@@ -0,0 +1,100 @@
...
+        for name, struct in defined_types.items():
+            if name != struct.name:
+                raise ValueError(
+                    f"Keys must match name of type, [{name}] != [{struct.name}]"
+                )
+
+    def parse_inline_type(self, type_string: str) -> CairoType:
+        """
+        Inline type it one that can be used inline. For instance (a: Uint256, b: felt*, c: (felt, felt)).
+        :param type_string: type to parse
```
war-in, line 52:
```suggestion
        Inline type it one that can be used inline. For instance (a: Uint256, b: felt*, c: (felt, felt)).
        
        :param type_string: type to parse
```

war-in (approved):

starknet_py/cairo/type_parser.py
```diff
@@ -0,0 +1,102 @@
...
+            if name != struct.name:
+                raise ValueError(
+                    f"Keys must match name of type, [{name}] != [{struct.name}]"
+                )
+
+    def parse_inline_type(self, type_string: str) -> CairoType:
+        """
+        Inline type it one that can be used inline. For instance (a: Uint256, b: felt*, c: (felt, felt)).
+
+        :param type_string: type to parse
```
drknzz, line 54:
```suggestion
        :param type_string: type to parse.
```

starknet_py/cairo/type_parser.py
```diff
@@ -0,0 +1,102 @@
...
+    CairoType,
+    FeltType,
+    TupleType,
+    NamedTupleType,
+    ArrayType,
+    StructType,
+)
+
+
+class UnknownTypeError(ValueError):
```
cptartur, line 18: This name is too familiar to python's `TypeError`. Can we rename it?

starknet_py/cairo/type_parser.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        TypeParser constructor.
+
+        :param defined_types: dictionary containing all defined types. For now they can only be structures.
+        """
+        self.defined_types = defined_types
+        for name, struct in defined_types.items():
+            if name != struct.name:
+                raise ValueError(
+                    f"Keys must match name of type, [{name}] != [{struct.name}]"
```
drknzz, line 47: Should this error message wrap the types in square brackets `[]`?
It might seem like a `list` at first glance.
`UnknownTypeError` has the type wrapped in single quotes - `f"Type '{type_name}' is not defined"`.

starknet_py/cairo/type_parser.py
```diff
@@ -0,0 +1,102 @@
...
+                )
+
+    def parse_inline_type(self, type_string: str) -> CairoType:
+        """
+        Inline type it one that can be used inline. For instance (a: Uint256, b: felt*, c: (felt, felt)).
```
cptartur, line 52: The first sentence of this comment doesn't make much sense to me: I think you left some part of an old comment or something.

starknet_py/cairo/type_parser.py
```diff
@@ -0,0 +1,102 @@
...
+        # Contracts don't support codeoffset as input/output type, user can only use it if it was defined in types
+        if isinstance(cairo_type, cairo_lang_types.TypeCodeoffset):
+            return self._get_struct("codeoffset")
+
+        # Other options are: TypeFunction, TypeStruct
+        # Neither of them are possible. In particular TypeStruct is not possible because we parse structs without
+        # info about other structs, so they will be just TypeIdentifier (structure that was not parsed).
+
+        # This is an error of our logic, so we throw a RuntimeError.
+        raise RuntimeError(f"Received unknown type [{cairo_type}] from parser")
```
drknzz, line 97: Same question here

starknet_py/cairo/type_parser.py
```diff
@@ -0,0 +1,102 @@
...
+        Inline type it one that can be used inline. For instance (a: Uint256, b: felt*, c: (felt, felt)).
+
+        :param type_string: type to parse
+        """
+        parsed = parse_type(type_string)
+        return self._transform_cairo_lang_type(parsed)
+
+    def _transform_cairo_lang_type(
+        self, cairo_type: cairo_lang_types.CairoType
+    ) -> CairoType:
```
cptartur, line 61: I think in the case of this module, private functions could use some docstrings.

starknet_py/cairo/type_parser_test.py
```diff
@@ -0,0 +1,104 @@
...
+    }
+    parsed = TypeParser(types).parse_inline_type(type_string)
+    assert parsed == expected
+
+
+def test_code_offset():
+    with pytest.raises(UnknownTypeError) as err_info:
+        TypeParser({}).parse_inline_type("codeoffset")
+
+    assert err_info.value.type_name == "codeoffset"
```
cptartur, line 96: You raise an error here but have
```py
 if isinstance(cairo_type, cairo_lang_types.TypeCodeoffset):
            return self._get_struct("codeoffset")
```
in your code. I don't fully understand why is this a desired behavior.

Solpatium: Parser will treat codeoffset specially, but we don't want that, we want to treat codeoffset as any user defined type. This means it should throw the same exception as when user type is not provided.

drknzz (approved):

PR #570, opened 2022-12-06
Add migration guide after Account refactor

cptartur:
Closes 

## Introduced changes

- This PR adds a migration guide with changes introduced by the Account refactor

##


docs/migration_guide.rst
```diff
@@ -1,10 +1,55 @@
...
+``get_storage_at``, ``call_contract`` etc. are not available in the Account interface.
+
+However ``Account`` now exposes a `.`client`` property, which means using an ``Account`` is
+just as simple as ``AccountClient`` was. For example:
+
+.. code-block:: python
+
+    account_client.get_storage_at(contract_address=0x1234)
+    # becomes
+    account.client.get_storage_at(contract_address=0x1234)
```
Solpatium, line 20: This is quite incredible that even length it the same :D

docs/migration_guide.rst
```diff
@@ -1,10 +1,55 @@
...
+
+Replacing inheritance with composition simplifies the ``Account`` interface and will make
+maintaining ``Account`` simpler.
+
+Changes in the Account interface
+--------------------------------
+
+1. Removed ``hash_message`` method. Use ``TypedData.message_hash`` directly instead.
+2. ``Account`` doesn't expose a ``net`` property
+3. ``Account`` doesn't accept a ``supported_tx_version`` parameter. It currently always uses version 1.
```
Solpatium, line 30: Any ideas how we want to handle new tx versions?

docs/migration_guide.rst
```diff
@@ -1,10 +1,55 @@
...
+    account.client.get_storage_at(contract_address=0x1234)
+
+Replacing inheritance with composition simplifies the ``Account`` interface and will make
+maintaining ``Account`` simpler.
+
+Changes in the Account interface
+--------------------------------
+
+1. Removed ``hash_message`` method. Use ``TypedData.message_hash`` directly instead.
+2. ``Account`` doesn't expose a ``net`` property
```
Solpatium, line 29: Missing `.`.

docs/migration_guide.rst
```diff
@@ -1,10 +1,55 @@
...
+TODO(Update this with correct version or move to separate file)
+
+This version deprecates the AccountClient, which is a major change to the StarkNet.py
+It is being replaced with new :class:`BaseAccount <starknet_py.net.account.base_account.BaseAccount>` ABC and its
+default implementation :class:`Account <starknet_py.net.account.account.Account>`.
+
+Unlike ``AccountClient``, an ``Account`` is not a Client anymore. That means that methods like
+``get_storage_at``, ``call_contract`` etc. are not available in the Account interface.
+
+However ``Account`` now exposes a `.`client`` property, which means using an ``Account`` is
```
Solpatium, line 13: `However,`

docs/migration_guide.rst
```diff
@@ -1,10 +1,55 @@
...
+just as simple as ``AccountClient`` was. For example:
+
+.. code-block:: python
+
+    account_client.get_storage_at(contract_address=0x1234)
+    # becomes
+    account.client.get_storage_at(contract_address=0x1234)
+
```
Solpatium, line 21: I would add a few more examples of things that people did before (calling contracts, sending deploy_account, executing).

Solpatium (changes requested):

war-in (approved):
Can be merged after @Solpatium's comments will be addressed

docs/migration_guide.rst
cptartur: I though that if we have a account interface anyways, we can have separate implementations (or subclasses of Account) for different versions if the need arises. That way we wouldn't have to deal with spaghetti version handling like we do in AccountClient, and could deprecated older versions easily.

docs/migration_guide.rst
```diff
@@ -1,10 +1,81 @@
...
+    await account.execute(call, max_fee=1000)
+
+
+    # Estimating fee for already created transactions
+    declare_tx = Declare(...)
+    estimate = await account_client.estimate_fee(declare_tx)  # Transaction is implicitly signed
+    # Becomes
+    signature = account.signer.sign_transaction(declare_tx)
+    declare_tx = dataclasses.replace(signature=signature)
+    esitmate = await account.client.estimate_fee(declare_tx)
```
cptartur, line 45: This example is quite unfortunate as it is a bit more complicated than the other ones. 

@war-in @Solpatium

docs/migration_guide.rst
```diff
@@ -1,10 +1,81 @@
...
+    await account.client.get_storage_at(contract_address=0x1234)
+
+
+    call = Call(to_addr=0x1, selector=0x1234, calldata=[])
+
+    # Sending transactions
+    tx = await account_client.sign_invoke_transaction(call, 10000)
+    await account_client.send_transaction(tx)
+    # becomes
+    tx = await account.sign_invoke_transaction(call, max_fee=1000)  # Not that max_fee is now keyword-only argument
```
war-in, line 30:
```suggestion
    tx = await account.sign_invoke_transaction(call, max_fee=1000)  # Note that max_fee is now keyword-only argument
```

docs/migration_guide.rst
```diff
@@ -1,10 +1,81 @@
...
+    # becomes
+    await account.execute(call, max_fee=1000)
+
+
+    # Estimating fee for already created transactions
+    declare_tx = Declare(...)
+    estimate = await account_client.estimate_fee(declare_tx)  # Transaction is implicitly signed
+    # Becomes
+    signature = account.signer.sign_transaction(declare_tx)
+    declare_tx = dataclasses.replace(signature=signature)
```
war-in, line 44: That example won't work :/ I'm wondering if we should create a test case for this entire code and import it as always
@cptartur @Solpatium

war-in (changes requested):

docs/migration_guide.rst
cptartur: Hmm, we can do that, though this one example will get removed anyways.

docs/migration_guide.rst
cptartur: I decided to not include this example at all.

Solpatium: Hmm, but why do you remove signature here? I think this is unnecessary

cptartur: `signer` is not a part of a `BaseAccount` interface currently, so there's no good reason to show signing directly using signer in these examples.

docs/migration_guide.rst
Solpatium: Ok, alias will work

Solpatium (approved):

Issue #512, opened 2022-11-18
Use typed dicts for Abi


drknzz:
Originated in #504

## Is your feature request related to a problem? Please describe.

Currently Gateway uses Abi defined vaguely as a list of dicts.

RPC abi is specified in the [specs](https://github.com/starkware-libs/starknet-specs/blob/refs/tags/v0.2.1/api/starknet_api_openrpc.json#L1741).
There is no official specification for gateway abi.
Although when looking into cairo-lang I didn't spot any fields that would be possible other than these already included in the dataclasses.

## Describe the solution you'd like
Find out whether there are any differences between `abi` in RPC and Gateway.
Use Abi dataclasses in other places in code.

## Describe alternatives you've considered
Keep using dataclasses for abi only in FullNodeClient.

cptartur:
We've decided to continue using dict/typeddict in interfaces, and only use dataclasses internally.

Solpatium:
We decided to use TypedDicts instead of dataclasses as there is no benefit from using dataclasses.

cptartur:
Blocked by #547

cptartur:
We're using dataclass instead

PR #571, opened 2022-12-06
Add ABI module

Solpatium:
Part of the effort for data transfomer.

During my implementation I found that crypto facade still had a method for hashing call. It was used long time ago for other signing method of multicall, so it is no longer needed. For this reason crypto imported call and was later used by high level modules, causing circular import. Also I found that many places import Call from crypto facade.


starknet_py/net/models/abi/definition.py
```diff
@@ -0,0 +1,173 @@
...
+
+class AbiDecodingError(ValueError):
+    pass
+
+
+T = TypeVar("T", bound=Dict)
+
+
+@dataclass
+class AbiDefinition:
```
Solpatium, line 33: I don't like this name, any suggestions?

cptartur: Why not just `Abi` or `ABI`?

starknet_py/net/models/abi/definition_test.py
```diff
@@ -0,0 +1,334 @@
...
+from starknet_py.net.models.abi.definition import AbiDefinition, AbiDecodingError
+from starknet_py.cairo.data_types import StructType, FeltType, ArrayType
+from starknet_py.tests.e2e.fixtures.misc import read_contract
+
+uint256_dict = {
+    "type": "struct",
+    "name": "Uint256",
+    "size": 2,
+    "members": [
+        # low and high were switched on purpose
```
cptartur, line 16: What's the purpose?

starknet_py/net/models/abi/definition_test.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+def test_missing_type_used_by_event():
+    with pytest.raises(UnknownCairoTypeError, match="Type 'User' is not defined"):
+        AbiDefinition.from_list([user_added_dict])
+
+
+def test_deserialize_proxy_abi():
+    # Contains all types of ABI apart from structures
+    abi = json.loads(read_contract("oz_proxy_abi.json"))
```
cptartur, line 200: We should already have a fixture for this contract that could be used here.

starknet_py/net/models/abi/definition_test.py
```diff
@@ -0,0 +1,334 @@
...
+        },
+    )
+
+
+def test_deserialize_balance_struct_event_abi():
+    # Contains all types of ABI apart from structures
+    abi = json.loads(read_contract("balance_struct_event_abi.json"))
+    deserialized = AbiDefinition.from_list(abi)
+
+    print(deserialized)
```
cptartur, line 265: 👀

starknet_py/net/schemas/rpc.py
```diff
@@ -288,79 +283,6 @@ def make_dataclass(self, data, **kwargs) -> BlockStateUpdate:
         )
 
 
-class StructMemberSchema(Schema):
```
cptartur, line 291: Old abi schemas has been removed, but I don't see them being replaced. Shouldn't you also change the code in FullNodeClient or some existing rpc schemas to use newly defined ABIs?

starknet_py/net/models/abi/definition.py
```diff
@@ -0,0 +1,180 @@
...
+    L1_HANDLER_ENTRY,
+    StructMemberDict,
+)
+
+
+class AbiDecodingError(ValueError):
+    pass
+
+
+T = TypeVar("T", bound=Dict)
```
cptartur, line 30: Can we use more descriptive name to suggest it is in fact something like a dict?

starknet_py/net/models/abi/definition.py
```diff
@@ -0,0 +1,180 @@
...
+        # Without cast it complains that
+        # 'Type "TypedMemberDict" cannot be assigned to type "T@_group_by_name"'
+        members = AbiDefinition._group_by_name(entity_name, cast(List[Dict], params))
+        return OrderedDict(
+            (name, parser.parse_inline_type(param["type"]))
+            for name, param in members.items()
+        )
+
+    @staticmethod
+    def _group_by_name(entity_name: str, dicts: List[T]) -> OrderedDict[str, T]:
```
cptartur, line 171: I'd have the parameter `entity_name` follow the `dicts` parameter. Having it as a primary one suggest that it is the name that is somehow used in the grouping process, at least that's what I though at first.

And in fact it's pretty much only used for having better error messages.

cptartur: Also may it could be renamed? `_group_by_entry_name` or something like that?

starknet_py/net/models/abi/definition.py
```diff
@@ -0,0 +1,180 @@
...
+                type_parser,
+            )
+            struct.types.update(members)
+
+        # All types have their members assigned now
+        return structs
+
+    @staticmethod
+    def _parse_abi_function(
+        function: FunctionDict, type_parser: TypeParser
```
cptartur, line 137: `TypeParser` is passed around a lot by all this methods, couldn't the structure be changed somehow so it doesn't have to be passed by methods? 

Perhaps having a separate AbiParser class with non static methods and having it as a member. I'm not a fan of chains of staticmethods if it can be avoided.

cptartur (changes requested):

starknet_py/net/models/abi/definition_test.py
Solpatium: Abi parser has to ensure the order as members have the offset field.

starknet_py/net/models/abi/definition_test.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+def test_missing_type_used_by_function():
+    with pytest.raises(UnknownCairoTypeError, match="Type 'Uint256' is not defined"):
+        AbiDefinition.from_list([get_user_dict])
+
+
+def test_missing_type_used_by_event():
+    with pytest.raises(UnknownCairoTypeError, match="Type 'User' is not defined"):
+        AbiDefinition.from_list([user_added_dict])
```
war-in, line 195: Could we parametrize it?

starknet_py/net/models/abi/definition.py
Solpatium: Ok I thought about it for a while and I have an idea

starknet_py/net/models/abi/definition_test.py
war-in, line 265: 👀

starknet_py/net/models/abi/definition_test.py
cptartur: Okay, I get it. Could this comment be improved? The test makes sense but it's not 100% obvious why it was switched up imo.

starknet_py/net/models/abi/definition_test.py
```diff
@@ -0,0 +1,334 @@
+import json
+from collections import OrderedDict
+
+import pytest
+
+from starknet_py.cairo.type_parser import UnknownCairoTypeError
+from starknet_py.net.models.abi.definition import AbiDefinition, AbiDecodingError
+from starknet_py.cairo.data_types import StructType, FeltType, ArrayType
+from starknet_py.tests.e2e.fixtures.misc import read_contract
+
```
war-in, line 10: It's quite a big file 😓 Let's move the setup (creating objects) to the fixtures directory

starknet_py/net/models/abi/definition.py
```diff
@@ -0,0 +1,180 @@
...
+        # Example problem (with a simplified json structure):
+        # [{name: User, fields: {id: Uint256}}, {name: "Uint256", ...}]
+        # User refers to Uint256 even though it is not known yet (will be parsed next).
+        # This is why it is important to create the structure types first. This way other types can already refer to
+        # them when parsing types, even thought their fields are not filled yet.
+        # At the end we will mutate those structures to contain the right fields. An alternative would be to use
+        # topological sorting with an additional "unresolved type", so this flow is much easier.
+        for name, struct in structs_dict.items():
+            structs[name] = StructType(name, OrderedDict())
+            struct_members[name] = sorted(struct["members"], key=lambda m: m["offset"])
```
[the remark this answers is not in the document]
war-in: Maybe to `member`? It would be more descriptive

starknet_py/net/models/abi/definition.py
war-in: What about an `Abi`?

starknet_py/net/models/abi/definition_test.py
Solpatium: We only use its abi, not contract itself.

starknet_py/net/models/abi/definition.py
Solpatium: I prefer functional approach to this, but added AbiParser. Unfortunately parsing functions and events is dependant on parsing structures. I don't like late init variables in cases like this, but this was the simplest solution.

starknet_py/net/schemas/rpc.py
Solpatium: https://github.com/software-mansion/starknet.py/issues/512

starknet_py/net/models/abi/parser.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+    def _parse_event(self, event: EventDict) -> Abi.Event:
+        return Abi.Event(
+            name=event["name"],
+            data=self._parse_members(event["name"], event["data"]),
+        )
+
+    def _parse_members(
+        self, entity_name: str, params: List[TypedMemberDict]
```
cptartur, line 140: Let's move the entity_name to the end here as well.

starknet_py/net/models/abi/definition.py
cptartur: Late init variable is not ideal, but I like the new approach a bit more than the previous one. 👍

Solpatium: Happy to hear you like it <3

cptartur (approved):

Solpatium:
I will merge this for now to avoid PR chains. Please reach out to me if you still want to change something @war-in, I'll address them in a separate PR.

PR #586, opened 2022-12-12
Replace `get_selector_from_name` and `get_storage_var_address` implementations

### LINKED ISSUE #580: Replace `get_selector_from_name` and `get_storage_var_address` implementations
### Feature Request

Write our own implementations of these.

drknzz:
Closes #580 

## Introduced changes

Remove uses of `starkware.starknet.public.api`
- replace `get_selector_from_name`
- replace `get_storage_var_address`

##


starknet_py/common.py
```diff
@@ -40,3 +66,17 @@ def int_from_bytes(
     Converts the given bytes object (parsed according to the given byte order) to an integer.
     """
     return int.from_bytes(value, byteorder=byte_order, signed=signed)
+
+
+def get_storage_var_address(var_name: str, *args) -> int:
```
cptartur, line 71: Do we actually have any usecases of `args` in starknet.py? I know that cairo-lang's implementation had it but from my experience, I've never used it.

cptartur (commented):
Two questions:

1. Should these method still be part of the public api? Perhaps it'd make sense to have underscores in their names.
2. Is eth-hash a good choice of a library? I know that cairo-lang uses it but [its readme](https://github.com/ethereum/eth-hash) recommends using higher level `eth_utils.keccak()`. Perhaps we should consider it?

starknet_py/common.py
drknzz: As do we use `*args` in the code? We've got it in some places.

drknzz:
> 1. Should these method still be part of the public api? Perhaps it'd make sense to have underscores in their names.

Well, for instance, the `Call` class is in public api. It takes `selector` argument, which is easier to pass as `get_selector_from_name("some_function")` and one can make sure it is the actual function they want to call by just looking at their code (instead of having an int as selector there).

> 2. Is eth-hash a good choice of a library? I know that cairo-lang uses it but [its readme](https://github.com/ethereum/eth-hash) recommends using higher level `eth_utils.keccak()`. Perhaps we should consider it?

You're right. Added the higher level `eth-utils` as a dependency.

starknet_py/common.py
```diff
@@ -31,6 +40,23 @@ def int_from_hex(number: Union[str, int]) -> int:
...
+        return DEFAULT_ENTRY_POINT_SELECTOR
+
+    return starknet_keccak(data=func_name.encode("ascii"))
+
+
+def starknet_keccak(data: bytes) -> int:
+    """
+    A variant of eth-keccak that computes a value that fits in a StarkNet field element.
+    """
+    return int_from_bytes(keccak(data)) & MASK_250
```
cptartur, line 57: Perhaps at least this one should be kept private?

cptartur (commented):
1. I though about this and I think this methods are core enough, that they should be kept public. But I added a comment to one I think shouldn't.
2. Good. I started wondering though, should we be concerned with the increased package size because of that? I think it's negligible, but perhaps it's not.

cc @Solpatium @war-in

starknet_py/common.py
cptartur: I though we didn't. Let's keep it as is then.

starknet_py/common.py
drknzz: Definitely could.

starknet_py/common.py
```diff
@@ -31,6 +40,23 @@ def int_from_hex(number: Union[str, int]) -> int:
     return number if isinstance(number, int) else int(number, 16)
 
 
+def get_selector_from_name(func_name: str) -> int:
```
Solpatium, line 43: Can we move all those functions to a better place?

starknet_py/common.py
```diff
@@ -40,3 +66,17 @@ def int_from_bytes(
...
+
+def get_storage_var_address(var_name: str, *args) -> int:
+    """
+    Returns the storage address of a StarkNet storage variable given its name and arguments.
+    """
+    res = starknet_keccak(var_name.encode("ascii"))
+
+    for arg in args:
+        if not isinstance(arg, int):
+            raise TypeError(f"Expected arguments to be integers. Found: {arg}.")
```
Solpatium, line 79: I think we can just type args properly.

Solpatium (changes requested):

starknet_py/common.py
drknzz: What exactly do you mean by typing args properly?

Solpatium: This: https://stackoverflow.com/a/37032111. Currently `get_storage_var_address` accepts anything, so it has to check types in the runtime.

starknet_py/common.py
drknzz: Not sure what would be the best place for this.
`starknet_py.net.client_utils` maybe?
Or a new separate file?

Solpatium: I think cairo/selector makes sense. Selector is a part of the cairo program.

drknzz: What about `get_storage_var_address`?

Solpatium: `cairo/storage.py`? I like grouping things properly. @war-in how do you plan to arrange something like this in changed structure?

war-in (approved):
Great ✊ It can be merged after resolving @Solpatium's comment

starknet_py/constants.py
```diff
@@ -16,4 +16,14 @@
...
 RPC_CLASS_HASH_NOT_FOUND_ERROR = 28
 
+DEFAULT_ENTRY_POINT_NAME = "__default__"
+DEFAULT_L1_ENTRY_POINT_NAME = "__l1_default__"
+DEFAULT_ENTRY_POINT_SELECTOR = 0
+MASK_250 = 2**250 - 1
+
+# MAX_STORAGE_ITEM_SIZE and ADDR_BOUND must be consistent with the corresponding constant in
+# starkware/starknet/common/storage/storage.cairo.
+MAX_STORAGE_ITEM_SIZE = 256
```
cptartur, line 26: How likely is that this value will get increased in the future causing bugs in starknet.py? Is there any way to test it except for manual verification (probably not)?

starknet_py/common.py
```diff
@@ -41,23 +31,6 @@ def int_from_hex(number: Union[str, int]) -> int:
     return number if isinstance(number, int) else int(number, 16)
```
Solpatium, line 33: Don't we want maintain backward compatibility with warnings?

Solpatium (changes requested):

starknet_py/common.py
drknzz: Where exactly?

starknet_py/constants.py
Solpatium: It might be changed in long future, but it will be a breaking change in starknet.

Solpatium (changes requested):
Please add unit tests for added methods. I recommend generating some values using cairo-lang and making sure our functions return the same result.

PR #588, opened 2022-12-13
Replace `FIELD_PRIME` in felt.py

### LINKED ISSUE #582: Replace `FIELD_PRIME` in felt.py
### Feature Request

Use our own definition.

drknzz:
Closes #582 

## Introduced changes

- add `FIELD_PRIME` to `constants.py`
- disable formatter in `constants.py` to allow for long lines

##


starknet_py/constants.py
```diff
@@ -7,13 +10,10 @@
...
-DEFAULT_DEPLOYER_ADDRESS = (
-    "0x041a78e741e5aF2fEc34B695679bC6891742439f7AFB8484Ecd7766661aD02BF"
-)
+FEE_CONTRACT_ADDRESS = "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
+DEFAULT_DEPLOYER_ADDRESS = "0x041a78e741e5aF2fEc34B695679bC6891742439f7AFB8484Ecd7766661aD02BF"
 
 RPC_INVALID_MESSAGE_SELECTOR_ERROR = 21
 RPC_CLASS_HASH_NOT_FOUND_ERROR = 28
+
+FIELD_PRIME = 3618502788666131213697322783095070105623107215331596699973092056135872020481
```
cptartur, line 19: So is it possible at all that this gets changed in starknet and we don't notice? I wonder if there's any way of having tests for that, but most likely not. 

cc @Solpatium

drknzz: I feel like it's such a basic constant in StarkNet that it couldn't change without breaking everything.

Solpatium: No @cptartur, it is not possible that they will just change it suddenly.

Yes @drknzz, please use hex value here. Python knows how to read it.

cptartur (approved):
After the value is changed to hex, can be merged.

Issue #589, opened 2022-12-13
Make data_transformer.result immutable


Solpatium:
### Feature Request

Currently it uses dicts. This means this object can't be used as a key or in a set. We should make it immutable and perhaps rethink the implementation.

Solpatium:
Currently blocked by #574

Issue #492, opened 2022-11-03
Create a solution for code to integrate and typecheck nicely with mypy


Arcticae:
When using contracts from static ABI, we could generate statically some types and cast contract objects, so that they could typecheck properly.

Arcticae:
bump

cptartur:
Blocked by #547

cptartur:
@Arcticae do we still want to support this feature?

Arcticae:
It's still a good idea IMHO to autogenerate types from ABI

tomek0123456789:
Too much effort for too little outcome

PR #590, opened 2022-12-13
Add TupleDataclass

Solpatium:
# About
In data_transformer module we return `Result` class and pretend it is a NamedTuple. This can lead to some unexpected behaviour.

The most important feature of NamedTuple is that it can be used both as tuple and a regular object. By using dataclasses with `__getitem__` and `__iter__` we can replicate this behaviour. 

## Differences
The differences with previous implementation:
- No `_asdict` method
- Comparison with tuple will return False
I am wondering if we should support these two 🤔 Let me know what you think.

I also added 2 utility functions: `as_tuple` and `as_dict` to make sure the most popular flows are covered out of the box.

## Cool things
In the future when doing https://github.com/software-mansion/starknet.py/issues/492 we can easily subclass `TupleDataclass` to provide strict typing 🔥 

# Other solutions
To be honest ideally I would just switch to using dicts as returned values from functions, but we would be missing spreading values like this: `x, y = result`. It would be a big breaking change and would make quality of life worse. I think using simple dataclasses is nice compromise: we rely on solid, well known implementation and offer some simplifications.

# Related issues

Closes https://github.com/software-mansion/starknet.py/issues/589.

### DIFF
--- a/starknet_py/utils/tuple_dataclass.py
+++ b/starknet_py/utils/tuple_dataclass.py
@@ -1,0 +1,41 @@
+from __future__ import annotations
+from dataclasses import dataclass, fields, make_dataclass
+from typing import Dict, Optional, Tuple
+
+
+@dataclass(frozen=True)
+class TupleDataclass:
+    """
+    Dataclass that behaves like a tuple at the same time.
+    """
+
+    # getattr is called when attribute is not found in object
+    # This way pyright will know that there might be some arguments it doesn't know about and will stop complaining
+    # about some fields that don't exist statically.
+    def __getattr__(self, item):
+        # This should always fail - only attributes that don't exist end up in here.
+        # We use __getattribute__ to get the native error.
+        return super().__getattribute__(item)
+
+    def __getitem__(self, item):
+        field = fields(self)[item]
+        return getattr(self, field.name)
+
+    def __iter__(self):
+        return (getattr(self, field.name) for field in fields(self))
+
+    def as_tuple(self) -> Tuple:
+        return tuple(self)
+
+    def as_dict(self) -> Dict:
+        return {field.name: getattr(self, field.name) for field in fields(self)}
+
+    @staticmethod
+    def from_dict(data: Dict, *, name: Optional[str] = None) -> TupleDataclass:
+        result_class = make_dataclass(
+            name or "TupleDataclass",
+            fields=[(key, type(value)) for key, value in data.items()],
+            bases=(TupleDataclass,),
+            frozen=True,
+        )
+        return result_class(**data)
--- a/starknet_py/utils/tuple_dataclass_test.py
+++ b/starknet_py/utils/tuple_dataclass_test.py
@@ -1,0 +1,22 @@
+from starknet_py.utils.tuple_dataclass import TupleDataclass
+
+
+def test_wrapped_named_tuple():
+    input_dict = {
+        "first": 1,
+        "second": 2,
+        "third": {"key": "value"},
+    }
+    input_tuple = tuple(input_dict.values())
+
+    result = TupleDataclass.from_dict(input_dict)
+    assert result.as_tuple() == input_tuple
+    assert result.as_dict() == input_dict
+    assert (result[0], result[1], result[2]) == input_tuple
+    assert (result.first, result.second, result.third) == input_tuple
+    assert str(result) == "TupleDataclass(first=1, second=2, third={'key': 'value'})"
+    assert repr(result) == "TupleDataclass(first=1, second=2, third={'key': 'value'})"
+
+    result = TupleDataclass.from_dict(input_dict, name="CustomClass")
+    assert str(result) == "CustomClass(first=1, second=2, third={'key': 'value'})"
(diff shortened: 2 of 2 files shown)


Solpatium:
I have just got an idea that might work, please don't review yet.

Solpatium:
Ok, it won't work, dataclasses are a better solution

starknet_py/utils/tuple_dataclass.py
```diff
@@ -0,0 +1,41 @@
...
+    Dataclass that behaves like a tuple at the same time.
+    """
+
+    # getattr is called when attribute is not found in object
+    # This way pyright will know that there might be some arguments it doesn't know about and will stop complaining
+    # about some fields that don't exist statically.
+    def __getattr__(self, item):
+        # This should always fail - only attributes that don't exist end up in here.
+        # We use __getattribute__ to get the native error.
+        return super().__getattribute__(item)
```
cptartur, line 18: Will it ever be called? From what I understand from docs:

Called when the default attribute access fails with an [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError) (either [__getattribute__()](https://docs.python.org/3/reference/datamodel.html#object.__getattribute__) raises an [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError) because name is not an instance attribute or an attribute in the class tree for self; or [__get__()](https://docs.python.org/3/reference/datamodel.html#object.__get__) of a name property raises [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError)). This method should either return the (computed) attribute value or raise an [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError) exception.

It will only get called if two other methods mentioned raise AttributeError. We don't implement any of these, only `__getitem__` which raises (at least should) IndexError.

starknet_py/utils/tuple_dataclass.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __getitem__(self, item):
+        field = fields(self)[item]
```
cptartur, line 21: How does it work? fields is supposed to return a tuple, shouldn't `item` be a index?

starknet_py/utils/tuple_dataclass.py
Solpatium: It will be called in case when someone wants to use an attribute that object doesn't have. Like this:
```
result = TupleDataclass.from_dict(input_dict)
result.unknown_property
```

starknet_py/utils/tuple_dataclass.py
Solpatium: Yes, item is an index. This way you can access seconds value with `result[1]`.

cptartur: Can we add type hint to that? 🙏

starknet_py/utils/tuple_dataclass.py
cptartur: Okay it makes sense. Perhaps info about `.` operator falling back to this could be included in the comment? It's kind of there but also it isn't.

starknet_py/utils/tuple_dataclass.py
Solpatium: Ok I thought pyright would complain

war-in (approved):
Seems good 🥇 Just to be sure, it is breaking for the users, right? Is `TupleDataclass.as_tuple` needed to compare TupleDataclass with a regular tuple?

Solpatium:
>Couldn't we override __eq__ to handle dicts and tuples?

I thought about doing that, but then realised that tuple won't know how to compare themselves to our dataclass. I checked and actually python prioritises custom __eq__, so it works 👌

starknet_py/utils/tuple_dataclass.py
```diff
@@ -0,0 +1,51 @@
...
+    def as_dict(self) -> Dict:
+        return {field.name: getattr(self, field.name) for field in fields(self)}
+
+    # Added for backward compatibility with previous implementation based on NamedTuple
+    def _asdict(self):
+        return self.as_dict()
+
+    def __eq__(self, other):
+        if isinstance(other, TupleDataclass):
+            return self.as_tuple() == other.as_tuple()
```
cptartur, line 39: Can we test this line? It's important for a module like this.

starknet_py/utils/tuple_dataclass.py
```diff
@@ -0,0 +1,51 @@
...
+    Dataclass that behaves like a tuple at the same time.
+    """
+
+    # getattr is called when attribute is not found in object. For instance when using object.unknown_attribute.
+    # This way pyright will know that there might be some arguments it doesn't know about and will stop complaining
+    # about some fields that don't exist statically.
+    def __getattr__(self, item):
+        # This should always fail - only attributes that don't exist end up in here.
+        # We use __getattribute__ to get the native error.
+        return super().__getattribute__(item)
```
cptartur, line 18: Same here

cptartur (changes requested):

cptartur (approved):

PR #592, opened 2022-12-14
Remove part of the cairo lang dependency

war-in:
## Introduced changes

Tracking PR containing changes connected with removing Cairo lang dependency

##


cptartur:
Maybe let's rename this PR? It's a bit misleading since cairo-lang is still staying with us for a while.

starknet_py/utils/crypto/transaction_hash.py
```diff
@@ -0,0 +1,145 @@
+from enum import Enum
```
drknzz, line 1: Not sure calculating transaction hashes should be in `crypto` module.

drknzz: https://github.com/software-mansion/starknet.py/pull/660#discussion_r1067997376

pyproject.toml
```diff
@@ -23,7 +23,9 @@ marshmallow = "^3.15.0"
 marshmallow-oneofschema = "^3.0.1"
 typing-extensions = "^4.3.0"
 cairo-lang = "==0.10.3"
+eth-utils = "^2.1.0"
```
cptartur, line 26: Why is this needed?

starknet_py/cairo/storage_test.py
```diff
@@ -0,0 +1,21 @@
...
+        ("a", [123, 456, 789], 404238469756318355939104484967681930221679732602307291786242730553081970461),
+        ("a" * 32, [], 1639766182393060139699536825713183380146349476460289054183902158918808727042),
+        ("a" * 64, [32, 64, 128, 256], 1170278195143236183253456127997214153696661952271500757743981570818429662143),
+        ("storage_name", [], 268122074306748686933226222576463058414657484382047525198367644523311212699),
+        ("storage_name", [2, 2, 2], 1815994203838325690189267524162006605165509742004447508084124379729518768247),
+        ("another_storage_name", [3], 3209489259786454249517264229792296517536124865381513663576107903160324151127),
+    ],
+)
+def test_get_storage_var_address(value, args, address):
+    assert get_storage_var_address(value, *args) == address
```
cptartur, line 21: Would a test for an address that would be higher than address bound be useful somehow?

starknet_py/constants.py
```diff
@@ -1,11 +1,29 @@
-# Address came from starkware-libs/starknet-addresses repository: https://github.com/starkware-libs/starknet-addresses
-FEE_CONTRACT_ADDRESS = (
-    "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
-)
+# pylint: disable=line-too-long
+# fmt: off
```
cptartur, line 2: Why is formatting disabled here?

starknet_py/constants.py
```diff
@@ -1,11 +1,29 @@
...
+DEFAULT_DEPLOYER_ADDRESS = "0x041a78e741e5aF2fEc34B695679bC6891742439f7AFB8484Ecd7766661aD02BF"
 
 RPC_INVALID_MESSAGE_SELECTOR_ERROR = 21
 RPC_CLASS_HASH_NOT_FOUND_ERROR = 28
+
+DEFAULT_ENTRY_POINT_NAME = "__default__"
+DEFAULT_L1_ENTRY_POINT_NAME = "__l1_default__"
+DEFAULT_ENTRY_POINT_SELECTOR = 0
+DEFAULT_DECLARE_SENDER_ADDRESS = 1
+MASK_250 = 2**250 - 1
```
cptartur, line 17: Is it used in more than 1 place? I'm not sure if it makes sense to define the mask as package constant.

starknet_py/net/full_node_client.py
```diff
@@ -422,7 +425,7 @@ def _create_broadcasted_txn(transaction: Union[Invoke, Declare, DeployAccount])
 
 
 def _create_broadcasted_declare_properties(transaction: Declare) -> dict:
-    contract_class = transaction.dump()["contract_class"]
+    contract_class = cast(Dict, DeclareSchema().dump(obj=transaction))["contract_class"]
```
cptartur, line 428: Why do we cast here?

starknet_py/net/full_node_client.py
```diff
@@ -443,7 +446,9 @@ def _create_broadcasted_invoke_properties(transaction: Invoke) -> dict:
 def _create_invoke_v0_properties(transaction: Invoke) -> dict:
     invoke_properties = {
         "contract_address": hash_to_felt(transaction.contract_address),
-        "entry_point_selector": hash_to_felt(transaction.entry_point_selector),
+        "entry_point_selector": hash_to_felt(
+            cast(int, transaction.entry_point_selector)
```
cptartur, line 450: Same here

starknet_py/net/models/transaction.py
```diff
@@ -1,29 +1,224 @@
...
+    """
+
+    max_fee: int = field(metadata={"marshmallow_field": Felt()})
+    signature: List[int] = field(
+        metadata={"marshmallow_field": fields.List(fields.String())}
+    )
+    nonce: int = field(metadata={"marshmallow_field": Felt()})
+
+
+# Used in the Account instead of Union[Invoke, Declare, DeployAccount]
```
cptartur, line 65: It's also used in Client, maybe use a better comment here?

starknet_py/net/models/transaction.py
```diff
@@ -1,29 +1,224 @@
...
+
+
+@dataclass(frozen=True)
+class Declare(AccountTransaction):
+    """
+    Represents a transaction in the StarkNet network that is a declaration of a StarkNet contract
+    class.
+    """
+
+    contract_class: ContractClass
```
cptartur, line 76: One think to keep in mind: eventually we'll have to replace ContractClass with our own implementation.

starknet_py/net/models/transaction.py
```diff
@@ -1,29 +1,224 @@
...
+class DeployAccount(AccountTransaction):
+    """
+    Represents a transaction in the StarkNet network that is a deployment of a StarkNet account
+    contract.
+    """
+
+    class_hash: int = field(metadata={"marshmallow_field": Felt()})
+    contract_address_salt: int = field(metadata={"marshmallow_field": Felt()})
+    constructor_calldata: List[int] = field(
+        metadata={"marshmallow_field": fields.List(fields.String())}
```
cptartur, line 119: Just to make sure, gateway expects this as list of strings not hex strings?

starknet_py/net/models/transaction.py
```diff
@@ -1,29 +1,224 @@
...
+    """
+    Represents a transaction in the StarkNet network that is an invocation of a Cairo contract
+    function.
+    """
+
+    contract_address: int = field(metadata={"marshmallow_field": Felt()})
+    calldata: List[int] = field(
+        metadata={"marshmallow_field": fields.List(fields.String())}
+    )
+    nonce: Optional[int] = field(metadata={"marshmallow_field": NoneFelt()})
```
cptartur, line 158: Let's make sure we make this non-optional once making v0 transactions is gone.

starknet_py/net/models/transaction.py
```diff
@@ -1,29 +1,224 @@
...
+    # A field element that encodes the signature of the invoked function.
+    # The entry_point_selector is deprecated for version 1 and above (transactions
+    # should go through the '__execute__' entry point).
+    entry_point_selector: Optional[int] = field(
+        default=None, metadata={"marshmallow_field": NoneFelt()}
+    )
```
cptartur, line 164: Same here

starknet_py/net/models/transaction.py
```diff
@@ -1,29 +1,224 @@
...
+        ), f"entry_point_selector should be None in version {version}."
+        del data["entry_point_selector"]
+
+        return data
+
+    def calculate_hash(self, chain_id: StarknetChainId) -> int:
+        """
+        Calculates the transaction hash in the StarkNet network.
+        """
+        if self.version == 0:
```
cptartur, line 190: Maybe some tests for this would be useful so codecov stops complaining? Also for the code above and below.

starknet_py/net/models/transaction.py
```diff
@@ -46,16 +241,21 @@ def compute_invoke_hash(
     :param version: Contract version
     :return: calculated hash
     """
+    warnings.warn(
+        "Function compute_invoke_hash is deprecated."
+        "To compute hash of an invoke transaction use compute_transaction_hash.",
+        category=DeprecationWarning,
+    )
```
cptartur, line 248: Once this is merged, let's create an issue for removing this.

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -56,33 +59,35 @@ def public_key(self) -> int:
 
     def sign_transaction(
         self,
-        transaction: Transaction,
+        transaction: AccountTransaction,
```
cptartur, line 62: This is not really following the BaseSigner interface. Either interface should be updated, or we should keep the typing here as it was (I think I prefer the 2nd option).

starknet_py/tests/e2e/utils.py
```diff
@@ -79,3 +75,11 @@ async def get_deploy_account_transaction(
         constructor_calldata=[key_pair.public_key],
         max_fee=MAX_FEE,
     )
+
+
+def _get_random_private_key() -> int:
```
cptartur, line 80: Maybe let's name it something like `_get_random_private_key_unsafe` so it's not accidentally used in library code.

pyproject.toml
drknzz: `_starknet_keccak` uses `keccak` from eth-utils

starknet_py/cairo/storage_test.py
drknzz: `get_storage_var_address` returns addresses `% ADDR_BOUND`.

starknet_py/constants.py
drknzz: To have long integer constants in one line instead this:
```
DEFAULT_DEPLOYER_ADDRESS = (
    "0x041a78e741e5aF2fEc34B695679bC6891742439f7AFB8484Ecd7766661aD02BF"
)
```

starknet_py/net/full_node_client.py
drknzz: Optional would be my guess.

starknet_py/net/models/transaction.py
drknzz: There is an existing issue for that.

starknet_py/tests/e2e/utils.py
drknzz: It's already in `tests` module

starknet_py/net/models/transaction.py
cptartur: Right. #661

starknet_py/tests/e2e/utils.py
cptartur: Yeah, but I think it'd still be possible to accidentally import that in your IDE. Review would most likely catch that but it is still risky.

drknzz: Okay, i'm down to change it ⬇️

starknet_py/net/models/transaction.py
```diff
@@ -1,29 +1,238 @@
...
+    @abstractmethod
+    def calculate_hash(self, chain_id: StarknetChainId) -> int:
+        """
+        Calculates the transaction hash in the StarkNet network - a unique identifier of the
+        transaction. See :ref:`compute_transaction_hash` docstring for more details.
+        """
+
+
+@dataclass(frozen=True)
+class AccountTransaction(Transaction):
```
cptartur, line 51: I think (or actually my IDE thinks), this should also have `ABC` as base class, since it's abstract.

starknet_py/net/signer/stark_curve_signer.py
war-in: I think the `BaseSigner` interface should change because `Transaction` doesn't even have a `signature` field

starknet_py/net/full_node_client.py
war-in: Because pyright complains :/ the returned value is a dict, but it sees a list. I can add `ignore`

cptartur: I prefer to cast than to ignore in this case.

starknet_py/net/models/transaction.py
cptartur: What about this?

starknet_py/net/signer/base_signer.py
```diff
@@ -20,7 +20,7 @@ def public_key(self) -> int:
         """
 
     @abstractmethod
-    def sign_transaction(self, transaction: Transaction) -> List[int]:
+    def sign_transaction(self, transaction: AccountTransaction) -> List[int]:
```
cptartur, line 23: Not sure if this is okay that BaseAccount accepts any `Transaction` in its methods, but BaseSigner only works with `AccountTransaction`.

starknet_py/net/models/transaction.py
war-in: Will add them

starknet_py/hash/class_hash_test.py
```diff
@@ -0,0 +1,24 @@
...
+        ("argent_proxy_compiled.json", 0x358ef62c72e99b3e94923826c2090f3875cee2f059e6236a8ec2e070ebd046b),
+        ("universal_deployer_compiled.json", 0x56679f765a989444f0db426e54ee779b7685ccfd4e39113b9c963eb903e20e3),
+        ("precompiled/oz_proxy_address_0.8.1_compiled.json", 0x413c36c287cb410d42f9e531563f68ac60a2913b5053608d640fb9b643acfe6),
+    ]
+)
+def test_compute_class_hash(contract_source, hash_):
+    compiled_contract = read_contract(contract_source)
+    contract_class = create_contract_class(compiled_contract)
+    class_hash = compute_class_hash(contract_class)
+    assert class_hash == hash_
```
cptartur, line 24: I wonder if it'd make sense (at least for the time being) to also compute these using cairo-lang methods and compare the results. Contract class computation historically changed quite often, and with our own implementation, we will be unaware of such changes.

But perhaps theres some other approach of verifying that (declaring contract on starknet and comparing class hash computed there maybe?)

starknet_py/utils/crypto/crypto_test.py
```diff
@@ -36,20 +32,3 @@ def test_invalid_crypto_path(monkeypatch, mocker):
...
-            3442134774288875752012730520904650962184640568595562887119811371865001706826,
-        ),
-        (
-            [28, 15, 39, 74],
-            1457535610401978056129941705021139155249904351968558303142914517100335003071,
-        ),
-    ),
-)
-def test_compute_hash_on_elements(data, calculated_hash):
-    assert compute_hash_on_elements(data) == calculated_hash
```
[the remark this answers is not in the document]
drknzz: It has been moved to `hash/utils/utils_test.py`

starknet_py/contract.py
```diff
@@ -647,10 +642,11 @@ def compute_address(
         compiled = create_compiled_contract(
             compilation_source, compiled_contract, search_paths
         )
+        assert compiled.abi is not None
```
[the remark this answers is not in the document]
drknzz: It's just for our typechecking purposes though. `CompiledContractSchema` has abi as a required key.
I added a comment explaining it.

starknet_py/net/client_models.py
```diff
@@ -345,6 +345,15 @@ class DeclaredContract:
     abi: Optional[AbiDictList] = None
 
 
+@dataclass
+class CompiledContract(ContractClass):
+    """
+    Dataclass representing ContractClass with required abi.
+    """
+
+    abi: AbiDictList = field(default_factory=list)
```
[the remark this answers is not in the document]
drknzz: Added a comment in the code.

starknet_py/hash/class_hash_test.py
drknzz: Added comparing with class_hash calculated using cairo-lang.

starknet_py/hash/class_hash_test.py
```diff
@@ -1,6 +1,10 @@
 # pylint: disable=line-too-long
 # fmt: off
 import pytest
+from starkware.starknet.core.os.class_hash import (
+    compute_class_hash as sw_compute_class_hash,
+)
```
cptartur, line 6: I think it's fine for now, but we should try to avoid adding new functionalities based on cairo-lang library, even when it comes to tests.

starknet_py/tests/e2e/docs/account_creation/test_deploy_prefunded_account.py
```diff
@@ -17,11 +17,11 @@ async def test_deploy_prefunded_account(
     # pylint: disable=import-outside-toplevel, too-many-locals
     # docs: start
     from starknet_py.hash.address import compute_address
-    from starknet_py.net import KeyPair
     from starknet_py.net.account.account import Account
     from starknet_py.net.gateway_client import GatewayClient
     from starknet_py.net.models import StarknetChainId
     from starknet_py.net.networks import TESTNET
+    from starknet_py.net.signer.stark_curve_signer import KeyPair
```
cptartur, line 24: Why was this changed?

drknzz: See [description of this](https://github.com/software-mansion/starknet.py/pull/760)

cptartur: Okay, makes sense.

cptartur (approved):

PR #596, opened 2022-12-14
Remove `ContractAddressSalt.get_random_value()` usages

### LINKED ISSUE #585: Replace `ContractAddressSalt.get_random_value()` with our implementation
### Feature Request

Consider the security of the new approach

war-in:
Closes #585 

## Introduced changes

Use `get_random_salt()` from utils/crypto/facade.py instead of `ContractAddressSalt.get_random_value()`

##


starknet_py/utils/crypto/facade.py
```diff
@@ -46,3 +48,7 @@ def pedersen_hash(left: int, right: int) -> int:
     if use_cpp_variant():
         return cpp_hash(left, right)
     return default_hash(left, right)
+
+
+def get_random_salt() -> int:
+    return random.Random().randrange(0, FIELD_PRIME)
```
drknzz, line 54: Shouldn't FIELD_PRIME also be included?

drknzz: I've checked, it should not.

drknzz (approved):
👍

starknet_py/utils/crypto/facade.py
Solpatium, line 54: This doesn't have anything with crypto, it doesn't use entropy based random values now. I think we can move it somewhere else.

Solpatium (changes requested):

cptartur (commented):
Agreed with @Solpatium comment

starknet_py/utils/crypto/facade.py
war-in: Does everyone agree that `get_random_salt` should be moved to the `udc_deployer/deployer.py` file? I'm quite unsure, but this is the only place where we use it (outside the tests)

cptartur (approved):

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -123,3 +125,7 @@ def _compute_address(
             salt=salt,
             deployer_address=deployer_address,
         )
+
+
+def get_random_salt() -> int:
```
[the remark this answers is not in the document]
war-in: And it is :hmmm: 
image

Solpatium (approved):

PR #597, opened 2022-12-14
Replace `get_random_private_key`

### LINKED ISSUE #583: Replace `get_random_private_key`
### Feature Request

Consider the security of the new implementation.

drknzz:
Closes #583 

## Introduced changes

- move `get_random_private_key` a test utils
- use simple `random.randint` similarly to [here](https://github.com/starkware-libs/starkex-resources/blob/master/crypto/starkware/crypto/signature/signature.py#L86)

##


cptartur (commented):
I'm unsure about exposing this functions to the user. Perhaps lets not include get_random_private_key at all, and use some basic generation for local tests.

starknet_py/tests/e2e/utils.py
```diff
@@ -79,3 +80,10 @@ async def get_deploy_account_transaction(
         constructor_calldata=[key_pair.public_key],
         max_fee=MAX_FEE,
     )
+
+
+def _get_random_private_key() -> int:
+    """
+    Returns a private key in the range [1, EC_ORDER).
+    """
+    return random.randint(1, EC_ORDER - 1)
```
cptartur, line 89: Maybe lets also remove the docstring? I know I always recommend adding docstring to private methods, but I'd like have a better indication that this is internal and shouldn't be touched. Perhaps lets also have EC_ORDER hardcoded in the function so someone cannot import it from our library?

cptartur (commented):
Except for the comment, I think it's okay.

I'd like @Solpatium to also review this one though.

starknet_py/tests/e2e/utils.py
drknzz: Is it really fair to go as far as removing a function description and hardcoding values into a function, just so a user does not use it themselves? It already is prefixed with `_` which by common convention is private. It also is not a part of our public api.
We could also start defining `__all__` in modules, but it still wouldn't stop users from using our private functions if they decided to.

Solpatium (approved):

war-in (approved):

cptartur (approved):

PR #598, opened 2022-12-15
Improve installation docs

### LINKED ISSUE #577: Improve docs about installation on ARM mac devices
### Feature Request

We should inform people how to install fastecdsa which is problematic on mac computers with arm processers.

cptartur:
Closes #577

## Introduced changes

- Extended MacOS with Apple silicon installation docs
- Extended docs for other platforms
- Added note on Windows being not supported

##

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -2,7 +2,7 @@ Installation
 ============
 
 To use StarkNet.py, ``ecdsa, fastecdsa, sympy`` dependencies are required. Depending on the operating system,
-different install steps must be performed.
+different installation steps must be performed.
 
 Linux
 ^^^^^
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -31,8 +31,7 @@ Apple silicon
 .. code-block:: console
 
     brew install gmp
-    CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip install ecdsa fastecdsa sympy
-    pip install starknet-py
+    CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip install starknet-py
 
 Windows
 -------
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -43,4 +43,4 @@ Windows
 ^^^^^^^
 
 This library is incompatible with Windows devices.
-Use virtual machine with Linux, Windows Subsystem for Linux 2 (WSL2) or other solutions.
+Use virtual machine with Linux, Windows Subsystem for Linux 2 (WSL2) or other solution.
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,11 +1,46 @@
 Installation
 ============
 
-To install this package run
+To use StarkNet.py, ``ecdsa, fastecdsa, sympy`` dependencies are required. Depending on the operating system,
+different install steps must be performed.
 
-``pip install starknet.py``
+Linux
+^^^^^
 
-or using Poetry:
+.. code-block:: console
 
-``poetry add starknet.py``
+    sudo apt install -y libgmp3-dev
+    pip install starknet-py
 
+MacOS with Intel processor
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Requires `Homebrew <https://brew.sh/>`_  installation.
+
+.. code-block:: console
+
+    brew install gmp
+    pip install starknet-py
+
+MacOS with Apple silicon
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Requires `Homebrew <https://brew.sh/>`_  installation.
+First install gmp:
+
+.. code-block:: console
+
+    brew install gmp
+
+Then install required dependencies and StarkNet.py itself:
+
+.. code-block:: console
+
+    CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip install ecdsa fastecdsa sympy
+    pip install starknet-py
+
+Windows
+^^^^^^^
+
+This library is incompatible with Windows devices.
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -1,11 +1,46 @@
...
+
+MacOS with Apple silicon
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Requires `Homebrew <https://brew.sh/>`_  installation.
+First install gmp:
+
+.. code-block:: console
+
+    brew install gmp
```
THenry14, line 33: I guess this could be resrtructured to have:

```
# Macos:
<install gmp>
## Apple silicon
<additional steps>
```
As the step with gmp is identical regardless of cpu.

docs/installation.rst
```diff
@@ -1,11 +1,46 @@
...
+.. code-block:: console
+
+    CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip install ecdsa fastecdsa sympy
+    pip install starknet-py
+
+Windows
+^^^^^^^
+
+This library is incompatible with Windows devices.
+Use virtual machine with Linux, Windows Subsystem for Linux 2 (WSL2) or other solutions.
```
THenry14, line 46:
```suggestion
Use virtual machine with Linux, Windows Subsystem for Linux 2 (WSL2) or other solution.
```

THenry14 (approved):
lgtm

docs/installation.rst
cptartur: I though about this, but I am unsure. I wanted to have the complete instruction for every platform in each section.

docs/installation.rst
```diff
@@ -1,11 +1,46 @@
...
+
+.. code-block:: console
+
+    brew install gmp
+
+Then install required dependencies and StarkNet.py itself:
+
+.. code-block:: console
+
+    CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip install ecdsa fastecdsa sympy
```
Solpatium, line 39: cmake is also needed for fastecdsa

docs/installation.rst
```diff
@@ -1,11 +1,46 @@
...
+    pip install starknet-py
+
+Windows
+^^^^^^^
+
+This library is incompatible with Windows devices.
+Use virtual machine with Linux, Windows Subsystem for Linux 2 (WSL2) or other solution.
```
Solpatium, line 46: WSL should be the only suggestion, it works well

docs/installation.rst
Solpatium: I would do 
```
#MacOS
## Apple silicon
## x86
```

docs/installation.rst
Solpatium, line 39: I think we can just do pip install starknet-py here. This way after uninstalling starknet-py dependencies won't be left in the system.

Solpatium (changes requested):
Some small tweaks might be great :)

docs/installation.rst
cptartur: Are you sure? And is it only on Apple silicon macs?

docs/installation.rst
cptartur: But using a VM is also a valid option. I specifically don't want to recommend a single solution so if someone has problems it doesn't look like "I used wsl2 as you said and it still doesn't work".

docs/installation.rst
cptartur: Does it work if you just do `CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip install starknet-py`? Personally I always did fastecdsa and the rest first, then starknet.py

docs/installation.rst
cptartur: I'll do something along the lines of this if you both prefer having subsections.

docs/installation.rst
THenry14: yeah, I remember the dark times when I was forced to use windows, and cygwin was pretty ok too

docs/installation.rst
```diff
@@ -1,11 +1,46 @@
...
+^^^^^
 
-or using Poetry:
+.. code-block:: console
 
-``poetry add starknet.py``
+    sudo apt install -y libgmp3-dev
+    pip install starknet-py
 
+MacOS with Intel processor
```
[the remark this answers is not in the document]
cptartur: I think it's overkill. If someone is unable to determine what kind of computer they have, there are better sources.

docs/installation.rst
Solpatium: Using VM is a huge overkill for using a python library. Using docker makes more sense.

docs/installation.rst
Solpatium: Kuba has been installing it yesterday and was missing cmake. Not sure if it is only needed for ARM ones though.

Solpatium: Yes it does, dependencies are installed with starknet.py and proper env variables are available.

Solpatium: I think we can add it to both versions just to be sure

Solpatium (approved):
LGTM with comments

war-in (approved):

docs/installation.rst
cptartur: It might be that the problem is caused by `wheel` not being installed causing pip to fall back to old installation method. I'm not really sure if I had to install cmake for fastecdsa to work.

docs/installation.rst
cptartur: this all falls into `or other solution`. If someone knows how to setup docker on windows (arguably harder than setting up a vm), they wouldn't need our recommendation anyways.

drknzz (approved):

docs/installation.rst
cptartur: Created #600 for this. Merging the current version.

PR #601, opened 2022-12-16
Feat/payload serializer

Solpatium:


starknet_py/cairo/serialization/data_serializers/payload_serializer.py
```diff
@@ -0,0 +1,85 @@
...
+
+        :return: Members serialized separately in SerializedPayload.
+        """
+        with SerializationContext.create() as context:
+            return SerializedPayload(
+                (name, list(values))
+                for name, values in serialize_from_dict_by_key(
+                    self.serializers, context, members
+                )
+            )
```
cptartur, line 66: I'm not sure if I understand possible use cases for this method. Could you elaborate on that?

starknet_py/cairo/serialization/serialized_payload.py
```diff
@@ -0,0 +1,15 @@
...
+class SerializedPayload(OrderedDict[str, CairoData]):
+    """
+    Dictionary for serialized payload members. Use ``to_calldata`` to get raw calldata.
+    """
+
+    def to_calldata(self) -> CairoData:
+        result = []
+        for entry_calldata in self.values():
+            result.extend(entry_calldata)
+        return result
```
cptartur, line 15:
```suggestion
        return [result for result in self.values()]
```
I think it's more concise.

drknzz: The suggestion is not equivalent, is it?
Originally there was `extend` method used, whereas the suggestion is the same as `append`.

cptartur: Hmm, so `entry_calldata` can be a list instead of a single item? Only then it should matter.

drknzz: The class inherits from `OrderedDict[str, CairoData]` where `CairoData = List[int]`.

starknet_py/cairo/serialization/data_serializers/payload_serializer.py
```diff
@@ -0,0 +1,85 @@
...
+@dataclass
+class PayloadSerializer(CairoDataSerializer[Dict, TupleDataclass]):
+    """
+    Serializer for payloads like function arguments/function outputs/events.
+    Can serialize a dictionary.
+    Deserializes data to a TupleDataclass.
+    """
+
+    # Value present only in constructor
+    input_serializers: InitVar[OrderedDict[str, CairoDataSerializer]]
```
war-in, line 38: Do we need this here? And if yes, tell me why 😅

Solpatium: I don't want to mutate the dictionary I accept in constructor. Will add comment.

starknet_py/cairo/serialization/_context.py
```diff
@@ -94,9 +92,11 @@ class SerializationContext(Context):
     Context used during serialization.
     """
 
+    # Type is iterator, because ContextManager doesn't work with pyright :|
+    # https://github.com/microsoft/pyright/issues/476
     @classmethod
     @contextmanager
-    def create(cls) -> ContextManager[SerializationContext]:
+    def create(cls) -> Iterator[SerializationContext]:
```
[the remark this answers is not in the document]
Solpatium: There is a comment above it

starknet_py/cairo/serialization/data_serializers/payload_serializer.py
Solpatium: The only reason I added it is: it was previously supported and exposed in PreparedFunctionCall. It complicates everything unfortunately. 

Now that I think about it, perhaps we should deprecate it there before the new data transformer is added and just ignore it completely? I hate that it adds more complexity for amost no gain. I didn't plan to add this at the beginning, but when I started integrating with contract.py I noticed it is missing. @war-in @drknzz please provide your view on this as well.

starknet_py/cairo/serialization/serialized_payload.py
Solpatium: We can turn it into flatmap list comprehension, but it might be less readable.

cptartur: If `extend` is necessary, I think it's okay to keep it the way it currently is.

starknet_py/cairo/serialization/data_serializers/payload_serializer.py
Solpatium: I remove serialize_members from PR and created https://github.com/software-mansion/starknet.py/pull/610.

war-in (approved):

cptartur (approved):

PR #602, opened 2022-12-16
Replace `compute_hash_on_elements`

### LINKED ISSUE #584: Replace `compute_hash_on_elements` and related methods
### Feature Request

* `compute_hash_on_elements`
* `calculate_contract_address_from_hash`
* `calculate_deploy_account_transaction_hash`
perhaps others

war-in:
Closes #584

## Introduced changes

There are two new functions in `facade.py`:
- compute_hash_on_elements
- calculate_contract_address_from_hash

##


starknet_py/utils/crypto/facade.py
```diff
@@ -1,19 +1,84 @@
...
+
+def calculate_contract_address_from_hash(
+    salt: int,
+    class_hash: int,
+    constructor_calldata: Sequence[int],
+    deployer_address: int,
+    hash_function: Callable[[int, int], int] = pedersen_hash,
+) -> int:
+    """
+    Calculates the contract address in the starkNet network - a unique identifier of the contract.
```
drknzz, line 58:
```suggestion
    Calculates the contract address in the StarkNet network - a unique identifier of the contract.
```

starknet_py/utils/crypto/facade.py
```diff
@@ -1,19 +1,84 @@
...
 from starkware.crypto.signature.signature import sign
 
+from starknet_py.common import int_from_bytes
 from starknet_py.net.client_models import Call
 
-# PREFIX_TRANSACTION = encoded 'StarkNet Transaction'
-PREFIX_TRANSACTION = 476441609247967894954472788179128007176248455022
+# From cairo-lang
+CONTRACT_ADDRESS_PREFIX = int_from_bytes(b"STARKNET_CONTRACT_ADDRESS")
+L2_ADDRESS_UPPER_BOUND = 2**251 - 256
```
drknzz, line 18: I've previously put constants like these in the `constants.py` file.
I know these probably won't be used anywhere else.
But after all the file `constants.py` is supposed to group and hold constants, so idk.

starknet_py/net/models/address.py
```diff
@@ -1,13 +1,10 @@
...
-from starkware.starknet.core.os.contract_address.contract_address import (
-    calculate_contract_address_from_hash,
-)
 from starkware.starknet.services.api.feeder_gateway.feeder_gateway_client import (
     BlockIdentifier,
 )
 
-
+from starknet_py.utils.crypto.facade import calculate_contract_address_from_hash
 from starknet_py.utils.docs import as_our_module
```
drknzz, line 9: So now we would have 2 functions that do the same. It previously made sense, as `compute_address` was our wrapper for cairo-lang's `calculate_contract_address_from_hash` function, but now that we implement it ourselves, we'd need to keep only one of these.

starknet_py/utils/crypto/facade.py
```diff
@@ -1,19 +1,84 @@
...
+    """
+    return functools.reduce(hash_func, [*data, len(data)], 0)
+
+
+def calculate_contract_address_from_hash(
+    salt: int,
+    class_hash: int,
+    constructor_calldata: Sequence[int],
+    deployer_address: int,
+    hash_function: Callable[[int, int], int] = pedersen_hash,
```
cptartur, line 51: Does it make sense to expose this parameter in our own function?

starknet_py/utils/crypto/facade.py
```diff
@@ -1,19 +1,80 @@
...
+            CONTRACT_ADDRESS_PREFIX,
+            deployer_address,
+            salt,
+            class_hash,
+            constructor_calldata_hash,
+        ],
+        hash_func=hash_function,
+    )
+
+    return raw_address % L2_ADDRESS_UPPER_BOUND
```
cptartur, line 77: Does it make sense to put in in facade.py? It is using compute_hash_on_elements but it is a higher level function. Perhaps a separate file or module would be better?

starknet_py/utils/crypto/crypto_test.py
```diff
@@ -37,3 +40,37 @@ def test_invalid_crypto_path(monkeypatch, mocker):
...
     file = get_cpp_lib_file()
     assert file is None
+
+
+@pytest.mark.parametrize(
+    "data, calculated_hash",
+    (
+        (
+            [1, 2, 3, 4, 5],
+            3442134774288875752012730520904650962184640568595562887119811371865001706826,
```
cptartur, line 50: How was this obtained?

starknet_py/net/models/address.py
war-in: Ok, but `compute_address` is in the public API. So there are two solutions:
- keep it as it is
- move `calculate_contract_addres_from_hash` to the `address.py` directory (and maybe rename it) because users can't import from `facade.py` (circular imports 🙈)

starknet_py/utils/crypto/facade.py
war-in: Ok, so we have a solution for duplicated functionality. Wait a second and I'll fix it

starknet_py/net/models/address.py
war-in: I will replace `compute_address` with implementation from the `calculate_contract_addres_from_hash`

war-in: Ok, check it now

cptartur: I'd prefer the moving solution.

drknzz (approved):

Solpatium (approved):
LGTM with one comment. I think we don't need to make hash configurable.

starknet_py/utils/crypto/facade.py
```diff
@@ -1,19 +1,45 @@
...
+    return not force_disable_ext and bool(cpp_lib_file)
+
+
+def pedersen_hash(left: int, right: int) -> int:
+    if use_cpp_variant():
+        return cpp_hash(left, right)
+    return default_hash(left, right)
+
+
+def compute_hash_on_elements(
```
cptartur, line 31: So we're leaving it in the facade.py after all?

war-in: Yup, `compute_hash_on_elements` would be here. Previously, we talked about compute_address

cptartur: Okay, makes sense

PR #544, opened 2022-11-28
Change docs to autoclass

drknzz:
Closes #535 

Also added `:member-order: bysource` as with the default, alphabetical order, properties were shuffled into methods.
It is possible to [set it as default in docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_default_options).

It looks cleaner with all the `_sync` methods at the bottom.

PR #611, opened 2022-12-19
Fix SentTransaction docs

cptartur:
Closes 

## Introduced changes

- This PR fixes the docs generation for fields of SentTransaction subclasses

Before 

image

After

image

##

### DIFF
--- a/docs/contract.rst
+++ b/docs/contract.rst
@@ -16,12 +16,15 @@ Contract
 
 .. autoclass:: InvokeResult
     :exclude-members: __init__, __new__
-    :members: wait_for_acceptance, wait_for_acceptance_sync
+    :members: wait_for_acceptance, wait_for_acceptance_sync, contract, invoke_transaction, hash, status, block_number
+    :member-order: groupwise
 
 .. autoclass:: DeployResult
     :exclude-members: __init__, __new__
-    :members: wait_for_acceptance, wait_for_acceptance_sync
+    :members: wait_for_acceptance, wait_for_acceptance_sync, deployed_contract, hash, status, block_number
+    :member-order: groupwise
 
 .. autoclass:: DeclareResult
     :exclude-members: __init__, __new__
-    :members: deploy, deploy_sync, wait_for_acceptance, wait_for_acceptance_sync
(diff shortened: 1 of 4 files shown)


Solpatium (approved):

docs/contract.rst
```diff
@@ -16,12 +16,15 @@ Contract
...
     :exclude-members: __init__, __new__
-    :members: wait_for_acceptance, wait_for_acceptance_sync
+    :members: wait_for_acceptance, wait_for_acceptance_sync, deployed_contract, hash, status, block_number
+    :member-order: groupwise
 
 .. autoclass:: DeclareResult
     :exclude-members: __init__, __new__
-    :members: deploy, deploy_sync, wait_for_acceptance, wait_for_acceptance_sync
+    :members: deploy, deploy_sync, wait_for_acceptance, wait_for_acceptance_sync, class_hash, compiled_contract, hash, status, block_number
+    :member-order: groupwise
```
drknzz, line 30: Maybe we should consider setting the member-order to `bysource`, as suggested [here](https://github.com/software-mansion/starknet.py/pull/544).
I think having the attribute members just below the class definition is more natural.

cptartur: I tried bysource and for some reason it was like: properties from class, methods and inherited methods, inherited properties.

So it'd put hash etc.  after wait_for_acceptance but class_hash before wait_for_acceptance (in DeclareResult case)

drknzz: I'll investigate this

cptartur: Okay, let me try that

cptartur: It works except for the DeclareResult.

Code like this

```rst
.. autoclass:: DeclareResult
    :members: class_hash, compiled_contract, hash, status, block_number, deploy, deploy_sync, wait_for_acceptance, wait_for_acceptance_sync
    :member-order: bysource
```

produces the output like

image

cptartur: Having it `groupwise` places methods on top, but at least properly separates them from the dataclass fields which I think is better.

drknzz (approved):
👍

war-in (approved):

PR #616, opened 2022-12-20
Unify error messages

### LINKED ISSUE #594: Replace all error messages to start with capital letters and end with full stops
### Feature Request

Unify error messages formatting.

drknzz:
Closes #594 

## Introduced changes

- Add capitalization and full stops to error messages.
- Some messages changed a bit more to match their format with other errors.

EDIT:
- Add tests for exception messages
- Change string searching, to match keyword when asserting error message in tests
- Remove long deprecated `AccountClient.get_code`

##

Theoretically each error message change is a breaking change.


starknet_py/contract.py
```diff
@@ -120,13 +120,14 @@ class DeclareResult(SentTransaction):
...
-                "None of the account, class_hash and compiled_contract fields can be None"
-            )
+        if self._account is None:
+            raise ValueError("Argument _account can't be None.")
+
+        if self.class_hash is None:
+            raise ValueError("Argument class_hash can't be None.")
+
+        if self.compiled_contract is None:
+            raise ValueError("Argument compiled_contract can't be None.")
```
war-in, line 130: Could we somehow unify them? Error message is the same, only argument name is different. Any ideas?

war-in (commented):
Nice 🔥

starknet_py/contract.py
cptartur: You could move ValueError creation to a method, but I think that would be overkill.

starknet_py/contract.py
```diff
@@ -120,13 +120,14 @@ class DeclareResult(SentTransaction):
...
-        if any(
-            field is None
-            for field in [self.class_hash, self._account, self.compiled_contract]
-        ):
-            raise ValueError(
-                "None of the account, class_hash and compiled_contract fields can be None"
-            )
+        if self._account is None:
+            raise ValueError("Argument _account can't be None.")
+
```
cptartur, line 125: We should test these

starknet_py/net/account/account_client.py
```diff
@@ -518,15 +520,17 @@ async def get_code(self, *args, **kwargs):
             category=DeprecationWarning,
         )
         if not isinstance(self.client, GatewayClient):
-            raise TypeError("AccountClient.get_code only supports using GatewayClient")
+            raise TypeError(
+                "Method AccountClient.get_code only supports using GatewayClient."
```
cptartur, line 524: I think previous version was okay. AccountClient starts with capital letter anyways.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -63,7 +63,7 @@ def create_deployment_call(
         :return: NamedTuple with call and address of the contract to be deployed
         """
         if not abi and calldata:
-            raise ValueError("calldata was provided without an abi")
+            raise ValueError("Argument calldata was provided without an abi.")
```
cptartur, line 66:
```suggestion
            raise ValueError("Argument calldata was provided without an ABI.")
```
If we are making RPC all caps, might as well do ABI I think.

@war-in

starknet_py/tests/e2e/contract_interaction/interaction_test.py
```diff
@@ -49,58 +49,34 @@ async def test_auto_fee_estimation(map_contract):
...
-
-@pytest.mark.asyncio
-async def test_throws_on_both_max_fee_in_prepare_and_auto_estimate(map_contract):
-    key = 2
-    value = 3
+    with pytest.raises(
+        ValueError,
+        match="Parameter max_fee must be specified when invoking a transaction.",
+    ):
+        await map_contract.functions["put"].invoke(key, value)
```
cptartur, line 59: This should still be their own separate tests. Combining them, you know nothing about what failed from just the test name.

starknet_py/contract.py
cptartur, line 125: These should be tested so codecov stops complaining.

cptartur (changes requested):

starknet_py/net/udc_deployer/deployer.py
war-in: Agree!

starknet_py/tests/e2e/contract_interaction/interaction_test.py
drknzz: I mean, the error messages are really expressive already.
We could split it into 2 tests, one for each error message, as they are repeated here.
How does that sound?

starknet_py/contract.py
drknzz: I think we can keep it that way, without doing any shenanigans. At least it's explicit.

starknet_py/tests/e2e/contract_interaction/interaction_test.py
cptartur: Two tests is better than one gigantic test, let's go for it.

starknet_py/net/account/account_client.py
```diff
@@ -309,7 +308,7 @@ async def _get_max_fee(
     ) -> int:
         if auto_estimate and max_fee is not None:
             raise ValueError(
-                "Max_fee and auto_estimate are exclusive and cannot be provided at the same time."
+                "Parameters max_fee and auto_estimate are mutually exclusive."
```
cptartur, line 311: Should we mix words `parameter` and `argument` or stick to a single one in every exception?

cc @Solpatium @war-in

cptartur (commented):
I think that the changes are fine, but the question about using `argument` or `parameter` has to be resolved before this is approved.

starknet_py/tests/e2e/contract_interaction/interaction_test.py
```diff
@@ -49,57 +49,30 @@ async def test_auto_fee_estimation(map_contract):
...
-    value = 3
-
     prepared_call = map_contract.functions["put"].prepare(key, value)
-
-    with pytest.raises(ValueError) as exinfo:
-        await prepared_call.invoke()
-
-    assert "Max_fee must be specified when invoking a transaction" in str(exinfo.value)
+    with pytest.raises(ValueError, match=error_message):
+        await prepared_call.invoke(max_fee=10, auto_estimate=True)
```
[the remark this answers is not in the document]
drknzz: If these were 2 separate test cases and something changed, one would have to update both as well, wouldn't they?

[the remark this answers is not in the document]
cptartur: Yes, but having two test failing in the first place indicates that two things also have to be fixed.

war-in (approved):

cptartur (approved):

PR #621, opened 2022-12-21
Replace transaction hash calculation

### LINKED ISSUE #607: Create module for hash calculation
### Feature Request

Currently, we use functions from the cairo lang:
- calculate_transaction_hash_common
- calculate_declare_transaction_hash
- calculate_deploy_account_transaction_hash

They should be replaced by our implementation

war-in:
Closes #607

## Introduced changes

Added:
- compute_transaction_hash
- compute_deploy_account_transaction_hash
- compute_declare_transaction_hash

Deprecated: 
- compute_invoke_hash

##


starknet_py/utils/crypto/transaction_hash.py
```diff
@@ -0,0 +1,96 @@
...
+    contract_class: ContractClass,
+    chain_id: int,
+    sender_address: int,
+    max_fee: int,
+    version: int,
+    nonce: int,
+) -> int:
+    class_hash = compute_class_hash(contract_class=contract_class)
+
+    if version in [0, QUERY_VERSION_BASE]:
```
cptartur, line 80: What's the purpose of `QUERY_VERSION_BASE` version?

starknet_py/utils/crypto/transaction_hash.py
```diff
@@ -0,0 +1,96 @@
...
+    ]
+
+    return compute_hash_on_elements(
+        data=data_to_hash,
+    )
+
+
+def compute_deploy_account_transaction_hash(
+    version: int,
+    contract_address: int,
```
cptartur, line 50: Wouldn't it make more sense not to take `contract_address` as a parameter and calculate it within the method instead?

starknet_py/utils/crypto/transaction_hash.py
```diff
@@ -0,0 +1,96 @@
+from enum import Enum
```
cptartur, line 1: Do we want the methods in this module to be public? If so, let's add docstrings.

starknet_py/utils/crypto/transaction_hash_test.py
```diff
@@ -0,0 +1,67 @@
...
+    compute_declare_transaction_hash,
+)
+
+
+@pytest.mark.parametrize(
+    "data, calculated_hash",
+    (
+        (
+            [TransactionHashPrefix.INVOKE, 2, 3, 4, [5], 6, 7, [8]],
+            1176541524183318710439453346174486593835028723304878846501762065493748312757,
```
cptartur, line 18: I'm assuming these were calculated using current cairo-lang implemenetations?

starknet_py/utils/crypto/transaction_hash.py
```diff
@@ -0,0 +1,96 @@
...
+    max_fee: int,
+    nonce: int,
+    salt: int,
+    chain_id: int,
+) -> int:
+    return compute_transaction_hash(
+        tx_hash_prefix=TransactionHashPrefix.DEPLOY_ACCOUNT,
+        version=version,
+        contract_address=contract_address,
+        entry_point_selector=0,
```
drknzz, line 62: Can we make a constant like [here](https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/starknet/public/abi.py#L21)?

starknet_py/utils/crypto/transaction_hash.py
drknzz: Once we uncover the meaning of it, it would be nice to also test this case.

starknet_py/utils/crypto/transaction_hash.py
war-in: There is a constant for that :o I've changed it

starknet_py/utils/crypto/transaction_hash.py
war-in: It's a constant from the cairo lang. Not sure what it means

starknet_py/utils/crypto/transaction_hash.py
cptartur: From cairo-lang code it seems that `DeployAccount.calculate_hash` method has `0` hardcoded for `deployer_address`. I'm not sure if it makes sense to expose this parameter specifically in the `DEPLOY_ACCOUNT` tx context.

cptartur: On the second though, perhaps it makes sense to just do it the way cairo-lang does it. This can be treated as lower level function so maybe it makes sense to not have this kind of implicit logic.

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -27,6 +21,12 @@
 from starknet_py.utils.crypto.facade import (
     message_signature,
 )
+from starknet_py.utils.crypto.transaction_hash import (
+    compute_transaction_hash,
+    TransactionHashPrefix,
+    compute_declare_transaction_hash,
+    compute_deploy_account_transaction_hash,
+)
```
drknzz, line 29: Maybe we could add [isort](https://pycqa.github.io/isort/) as a dev dependency, so it could run together with `poe format`?

drknzz (approved):

starknet_py/net/signer/stark_curve_signer.py
war-in: Ok, but in a separate PR. It will produce many changes

cptartur (approved):

starknet_py/utils/crypto/transaction_hash.py
cptartur: Should we crate an issue to investigate it?

war-in: Maybe? It's probably the highest transaction version

PR #629, opened 2023-01-04
Add serializer factory functions

Solpatium:


starknet_py/cairo/serialization/factory_test.py
```diff
@@ -0,0 +1,112 @@
...
+            regular=TupleType([FeltType()]),
+            named=NamedTupleType(OrderedDict(value=FeltType())),
+        ),
+    )
+    assert serializer_for_type(struct_with_tuples) == StructSerializer(
+        OrderedDict(
+            regular=TupleSerializer([FeltSerializer()]),
+            named=NamedTupleSerializer(OrderedDict(value=FeltSerializer())),
+        )
+    )
```
war-in, line 88:
```suggestion
@pytest.mark.parametrize(
    "structure, serializer",
    (
        (abi.defined_structures["Uint256"], Uint256Serializer()),
        (abi.defined_structures["PoolId"], pool_id_serializer),
        (abi.defined_structures["User"], user_serializer),
        (
            StructType(
                "structure",
                OrderedDict(
                    regular=TupleType([FeltType()]),
                    named=NamedTupleType(OrderedDict(value=FeltType())),
                ),
            ),
            StructSerializer(
                OrderedDict(
                    regular=TupleSerializer([FeltSerializer()]),
                    named=NamedTupleSerializer(OrderedDict(value=FeltSerializer())),
                )
            ),
        ),
    ),
)
def test_getting_type_serializer(structure, serializer):
    assert serializer_for_type(structure) == serializer
```

What do you think?

war-in (approved):
Fire 🔥 Just one comment

starknet_py/cairo/serialization/factory_test.py
Solpatium: I like current approach, because it is very simple and easy to follow. We create multiple serialisers here and it itself is complex. In this case it is very easy to track what went wrong, because there will be a line assigned. I am ok if others want me to parametrise it, though.

war-in: @cptartur ?

starknet_py/cairo/serialization/factory.py
```diff
@@ -0,0 +1,116 @@
...
+            OrderedDict(
+                (name, serializer_for_type(member_type))
+                for name, member_type in cairo_type.types.items()
+            )
+        )
+
+    raise InvalidTypeException(f"Received unknown Cairo type '{cairo_type}'.")
+
+
+# We don't want to require users to use OrderedDict. Regular python requires order since python 3.7.
```
cptartur, line 89: I wonder if it isn't a bit error prone. Would it be possible for someone to use something different that CPython that doesn't enforce order and have this function fail? 

I'm not sure if order is enforced by the language standard or is just a feature of default implementation.

starknet_py/cairo/serialization/factory_test.py
cptartur: I think parametrized tests produce nicer output in case of failures (the test name indicates what case failed). Maybe let's changed it.

starknet_py/cairo/serialization/factory.py
Solpatium: https://stackoverflow.com/a/39980744

python requires order since python 3.7.

cptartur: Right, it's fine to use dict then.

PR #630, opened 2023-01-04
Add cycle check in parser

Solpatium:


starknet_py/net/models/abi/parser.py
```diff
@@ -158,3 +171,15 @@ def _group_by_entry_name(
                 )
             grouped[name] = entry
         return grouped
+
+
+def _to_json(value):
+    class DataclassSupportingEncoder(json.JSONEncoder):
+        def default(self, o):
```
war-in, line 178: Maybe rename `o` to `object`?

war-in (approved):
Nice and simple 🚀

starknet_py/net/models/abi/parser.py
Solpatium: This is the name in parent class. If this method is called with keyword argument it would fail.

war-in: Oh, I see

cptartur (approved):

PR #631, opened 2023-01-04
Simplify `Contract` constructor

cptartur:
Closes #

## Introduced changes

- This PR combines Contract constructor arguments `account` and `client` into single `provider` and deprecates `client` argument.

##

- Contract constructor parameters changed

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -179,7 +179,7 @@ class DeclareResult(SentTransaction):
         )
 
         deployed_contract = Contract(
-            account=self._account,
+            provider=self._account,
             address=address,
             abi=abi,
         )
@@ -447,8 +447,9 @@ class Contract:
         self,
         address: AddressRepresentation,
         abi: list,
+        provider: Union[BaseAccount, Client] = None,  # pyright: ignore
+        *,
         client: Optional[Client] = None,
-        account: Optional[BaseAccount] = None,
     ):
         """
         Should be used instead of ``from_address`` when ABI is known statically.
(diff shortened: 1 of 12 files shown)


Solpatium (approved):
👌

starknet_py/contract_test.py
```diff
@@ -139,3 +141,46 @@ def test_deploy_result_post_init(gateway_client):
...
+
+
+def test_contract_create_with_account_client(gateway_account_client):
+    contract = Contract(address=0x1, abi=[], provider=gateway_account_client)
+    assert isinstance(contract.account, AccountProxy)
+    assert contract.client == gateway_account_client.client
+
+    contract = Contract(address=0x1, abi=[], client=gateway_account_client)
+    assert isinstance(contract.account, AccountProxy)
+    assert contract.client == gateway_account_client.client
```
war-in, line 176: That could be parametrized, but it will require using the `gateway_account_client` fixture in the `@pytest.mark.parametrize` which is not simple.

war-in: It is ok as it is

starknet_py/contract.py
```diff
@@ -447,8 +447,9 @@ def __init__(
         self,
         address: AddressRepresentation,
         abi: list,
+        provider: Union[BaseAccount, Client] = None,  # pyright: ignore
```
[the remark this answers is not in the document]
cptartur: 👍 #632

starknet_py/contract_test.py
cptartur: I'm not a fan of tests like this either. These could be their own separate tests, but `client` (and its tests) is going to be removed at some point, so I think it's fine to have it like this as a temporary solution.

war-in (approved):

PR #637, opened 2023-01-05
Use new serializer in contract

Solpatium:


starknet_py/contract.py
```diff
@@ -45,14 +52,23 @@
...
+    @cached_property
+    def identifier_manager(self) -> IdentifierManager:
+        warnings.warn(
+            "Using identifier_manager from ContractData has been deprecated. Consider using parsed_abi instead."
+        )
+        return identifier_manager_from_abi(self.abi)
+
+    @cached_property
+    def parsed_abi(self) -> Abi:
+        return AbiParser(self.abi).parse()
```
cptartur, line 65: Is it documented somewhere?

cptartur: Maybe a docstring would be good here?

Solpatium: This is basically a property of the object, but since it is frozen it is easier to do it this way.

Solpatium: I can add docstring

starknet_py/cairo/serialization/data_serializers/array_serializer_test.py
```diff
@@ -1,6 +1,6 @@
 import pytest
+from starkware.crypto.signature.signature import FIELD_PRIME
 
-from starknet_py.cairo.felt import FIELD_PRIME
```
war-in, line 3: But why? 🥺

war-in (approved):

starknet_py/cairo/serialization/data_serializers/array_serializer_test.py
Solpatium: Because it is just imported there.

PR #639, opened 2023-01-05
Make Compiler module pending deprecation

### LINKED ISSUE #635: Deprecate `Compiler` module
### Feature Request

We decided to remove the `Compiler`. It should be deprecated now

cptartur:
Closes #635 

## Introduced changes

- Adds pending deprecation warnings to Compiler module and methods with `compilation_source`.
- Removes the usage of Compiler in docs.

##

### DIFF
--- a/starknet_py/common.py
+++ b/starknet_py/common.py
@@ -1,12 +1,9 @@
+import warnings
 from typing import List, Literal, Optional, Union
 
 from starkware.starknet.services.api.contract_class import ContractClass
 
-from starknet_py.compile.compiler import (
-    Compiler,
-    StarknetCompilationSource,
-    create_contract_class,
-)
+from starknet_py.compile.compiler import Compiler, StarknetCompilationSource
 
 
 def create_compiled_contract(
@@ -14,6 +11,12 @@ def create_compiled_contract(
     compiled_contract: Optional[str] = None,
     search_paths: Optional[List[str]] = None,
 ) -> ContractClass:
+    warnings.warn(
+        "Function create_compiled_contract will be deprecated in the future. "
+        "Consider using create_contract_class instead.",
+        category=PendingDeprecationWarning,
+    )
+
     if not compiled_contract:
         if not compilation_source:
             raise ValueError(
@@ -25,6 +28,17 @@ def create_compiled_contract(
         ).compile_contract()
     definition = create_contract_class(compiled_contract)
     return definition
+
+
+def create_contract_class(
+    compiled_contract: str,
+) -> ContractClass:
+    """
+    Creates ContractDefinition either from already compiled contract
+
+    :return: a ContractDefinition
+    """
+    return ContractClass.loads(compiled_contract)
 
 
 def int_from_hex(number: Union[str, int]) -> int:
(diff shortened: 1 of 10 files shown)


starknet_py/common.py
```diff
@@ -27,6 +30,17 @@ def create_compiled_contract(
     return definition
 
 
+def create_contract_class(
+    compiled_contract: str,
+) -> ContractClass:
+    """
+    Creates ContractDefinition either from already compiled contract
+
+    :return: a ContractDefinition
```
war-in, line 39: Why `ContractDefinition`?

starknet_py/tests/e2e/block_test.py
```diff
@@ -36,17 +15,17 @@ async def declare_contract(account_client: AccountClient):
 
 
 @pytest.mark.asyncio
-async def test_pending_block(new_gateway_account_client):
+async def test_pending_block(new_gateway_account_client, map_compiled_contract):
     # TODO: change to new_account_client once devnet repaired
```
war-in, line 19: Is devnet repaired now? :hmm:

cptartur: It must be 🤔 And I think `new_gateway_account_client` should be changed to `account` to use a Account after refactor. Will address it in separate PR.

war-in (approved):

starknet_py/tests/e2e/fixtures/contracts.py
```diff
@@ -47,6 +46,14 @@ def map_compiled_contract() -> str:
     return read_contract("map_compiled.json")
 
 
+@pytest.fixture(scope="module")
+def simple_storage_with_event_compiled_contract() -> str:
+    """
+    Returns compiled simple storage contract that emits an event
```
drknzz, line 52:
```suggestion
    Returns compiled simple storage contract that emits an event.
```

starknet_py/net/signer/test_stark_curve_signer.py
```diff
@@ -8,8 +8,9 @@
 from starknet_py.net.models.transaction import Declare, DeployAccount, Invoke
 from starknet_py.net.signer.stark_curve_signer import KeyPair, StarkCurveSigner
 
-contract_source = (
-    Path(os.path.dirname(__file__)) / "../../tests/e2e/mock/contracts/erc20.cairo"
+compiled_contract = (
+    Path(os.path.dirname(__file__))
+    / "../../tests/e2e/mock/contracts_compiled/erc20_compiled.json"
```
drknzz, line 13: Can we do something about this path? Could it be somehow changed to relative to project root?

starknet_py/common.py
```diff
@@ -27,6 +30,17 @@ def create_compiled_contract(
     return definition
 
 
+def create_contract_class(
+    compiled_contract: str,
+) -> ContractClass:
+    """
+    Creates ContractDefinition either from already compiled contract
```
drknzz, line 37:
```suggestion
    Creates ContractDefinition either from already compiled contract.
```

starknet_py/common.py
```diff
@@ -27,6 +30,17 @@ def create_compiled_contract(
...
+
+    :return: a ContractClass
```
drknzz, line 39:
```suggestion
    :return: a ContractClass.
```

starknet_py/net/signer/test_stark_curve_signer.py
cptartur: We could change it to a fixture, but that would require indirect parametrisation of a fixture below. I can open an issue for that.

drknzz: Let's do something about it.

drknzz (approved):

drknzz (approved):

PR #642, opened 2023-01-09
Replace data transfomer in some modules

Solpatium:


starknet_py/contract.py
```diff
@@ -68,7 +68,7 @@ def parsed_abi(self) -> Abi:
         return AbiParser(self.abi).parse()
 
     @staticmethod
-    def from_abi(address: int, abi: ABI) -> "ContractData":
+    def from_abi(address: int, abi: ABI) -> ContractData:
```
Solpatium, line 71: We use __future__.annotations in this file, so we can use types normally.

starknet_py/net/models/abi/model.py
```diff
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-from collections import OrderedDict
 from dataclasses import dataclass
-from typing import Dict, Optional
+from typing import Dict, Optional, OrderedDict
```
Solpatium, line 4: We need OrderedDict type here (not class) :|

drknzz (approved):
👍

starknet_py/utils/contructor_args_translator.py
```diff
@@ -1,20 +1,16 @@
...
 
 
 def translate_constructor_args(
     abi: List, constructor_args: Optional[Union[List, dict]]
 ) -> List[int]:
-    constructor_abi = next(
-        (member for member in abi if member["type"] == "constructor"),
-        None,
-    )
+    parsed = AbiParser(abi).parse()
```
war-in, line 10: ❤️

war-in (approved):

PR #660, opened 2023-01-12
Support calculating class_hash

### LINKED ISSUE #655: Support calculating class_hash
### Feature Request

#

drknzz:
Closes #655

## Introduced changes

- **starknet_py.class_hash** module with `calculate_class_hash` function in place of implementation from cairo-lang
- up for possible suggestions with a better place for the module

##


starknet_py/class_hash_test.py
```diff
@@ -0,0 +1,27 @@
...
+        ("erc20_compiled.json", 0x6d157eb6e8cc2a587ed2cb22f29778c326a814d1375fc374981c82f6581332b),
+        ("oz_proxy_compiled.json", 0x601407cf04ab1fbab155f913db64891dc749f4343bc9e535bd012234a46dc61),
+        ("argent_proxy_compiled.json", 0x358ef62c72e99b3e94923826c2090f3875cee2f059e6236a8ec2e070ebd046b),
+        ("universal_deployer_compiled.json", 0x56679f765a989444f0db426e54ee779b7685ccfd4e39113b9c963eb903e20e3),
+        ("precompiled/oz_proxy_address_0.8.1_compiled.json", 0x413c36c287cb410d42f9e531563f68ac60a2913b5053608d640fb9b643acfe6),
+    ]
+)
+def test(contract_source, hash_):
+    compiled_contract = read_contract(contract_source)
+    contract_class = cast(DeclaredContract, DeclaredContractSchema().loads(compiled_contract))
```
war-in, line 25: pyright...

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+            result += " "
+        result += char
+    return result
+
+
+def _delete_attributes(program) -> dict:
+    if len(program["attributes"]) == 0:
+        # Remove attributes field from raw dictionary, for hash backward compatibility of
+        # contracts deployed prior to adding this feature.
+        del program["attributes"]
```
cptartur, line 115: Let's add an early return here instead of `if else`. It'll make this function a bit more readable here I think.

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
+import json
```
cptartur, line 1: Let's not have it in the main directory of the library.

starknet_py/constants.py
```diff
@@ -7,5 +7,7 @@
     "0x041a78e741e5aF2fEc34B695679bC6891742439f7AFB8484Ecd7766661aD02BF"
 )
 
+API_VERSION = 0
```
cptartur, line 10: I wonder how will we know that the api version changed. No that there's much we can do about it.

cptartur (changes requested):

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+
+    _encoded_builtins = [
+        encode_shortstring(builtin) for builtin in contract_class.program["builtins"]
+    ]
+    builtins_hash = compute_hash_on_elements(_encoded_builtins)
+
+    hinted_class_hash = _compute_hinted_class_hash(contract_class)
+
+    program_data_hash = compute_hash_on_elements(
+        [int(data_, 0) for data_ in contract_class.program["data"]]
```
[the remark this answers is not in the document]
war-in: Also `ContractClass` would change in the upcoming version :/

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+    program = contract_class.program
+    program["debug_info"] = None
+
+    if "attributes" in program:
+        program = _delete_attributes(program)
+
+    # If compiler_version is not present, this was compiled with a compiler before version 0.10.0.
+    # Use "(a : felt)" syntax instead of "(a: felt)" so that the class hash will be the same.
+    if "compiler_version" not in program:
+        program["identifiers"] = _fix_cairo_types(program["identifiers"])
```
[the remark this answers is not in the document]
war-in: Hmm makes sense. Let's do it and show an error when the field is missing

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
+import json
+from typing import List
+
+from starkware.cairo.common.hash_state import compute_hash_on_elements
```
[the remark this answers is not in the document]
drknzz: Yeah, I know, originated from development by mistake.

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+from starkware.starknet.public.abi import starknet_keccak
```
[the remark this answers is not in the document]
drknzz: These are implemented on our side on cario-lang dependency removal branch, so here it won't be a problem.

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+        ]
+    )
+
+
+def _entry_points_array(entry_points: List[EntryPoint]) -> List[int]:
+    entry_points_array = []
+    for entry_point in entry_points:
+        entry_points_array.extend([entry_point.selector, entry_point.offset])
+
+    return entry_points_array
```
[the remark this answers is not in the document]
drknzz: This list comprehension results in invalid syntax though 🙆‍♀️

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def _compute_hinted_class_hash(contract_class: DeclaredContract) -> int:
+    program = contract_class.program
+    program["debug_info"] = None
```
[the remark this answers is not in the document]
drknzz: [Docs](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-hash/#definition_of_a_class) -> Program hash:
```
$ starknet-compile --no_debug_info
```

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+    return identifiers
+
+
+def _add_backward_compatibility_space(cairo_type: str) -> str:
+    result = ""
+    for char in cairo_type:
+        if char == ":" and result[-1] != " ":
+            result += " "
+        result += char
+    return result
```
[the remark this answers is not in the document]
drknzz: Nice, will do 🏃‍♀️

starknet_py/constants.py
drknzz: We will know, since it's a part of class hash calculation.
If it changes, class hashes will change.

starknet_py/class_hash.py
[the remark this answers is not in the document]
drknzz: I think we absolutely should be supporting old versions of compiled contract.
What if someone fetches a declared contract from StarkNet that was compiled with compiler < 0.10.0?

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+    if "attributes" in program:
+        program = _delete_attributes(program)
+
+    # If compiler_version is not present, this was compiled with a compiler before version 0.10.0.
+    # Use "(a : felt)" syntax instead of "(a: felt)" so that the class hash will be the same.
+    if "compiler_version" not in program:
+        program["identifiers"] = _fix_cairo_types(program["identifiers"])
+
+    # ContractClass stores keys of `hints` field as integers, resulting in different sorting order
+    program["hints"] = {int(key): val for key, val in program["hints"].items()}
```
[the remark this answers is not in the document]
drknzz: The cairo-lang implementation of `calculate_class_hash` uses **strongly** defined `ContractClass`.
It, among other differences, stores keys of `hints` field as integers (and not strings).
Line 78 below uses `sort_keys=True`, so the keys have to be changed to ints (['0', '22', '3'] and [0, 3, 22] are both considered sorted).

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+    if len(program["attributes"]) == 0:
+        # Remove attributes field from raw dictionary, for hash backward compatibility of
+        # contracts deployed prior to adding this feature.
+        del program["attributes"]
+    else:
+        # Remove accessible_scopes and flow_tracking_data fields from raw dictionary, for hash
+        # backward compatibility of contracts deployed prior to adding this feature.
+        for attr in program["attributes"]:
+            if "accessible_scopes" in attr and len(attr["accessible_scopes"]) == 0:
+                del attr["accessible_scopes"]
```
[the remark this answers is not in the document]
drknzz: I don't think so, the class hash should get calculated correctly then.

starknet_py/class_hash.py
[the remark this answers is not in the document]
cptartur: We could use nested comprehensions here. Though I'm not sure if that would be readable.

starknet_py/class_hash.py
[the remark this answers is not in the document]
cptartur: Shouldn't it just be deleted then?

starknet_py/class_hash.py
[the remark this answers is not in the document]
cptartur: Hmm, this is valid concern, but I'm not sure how common of a use case would that be.

starknet_py/class_hash.py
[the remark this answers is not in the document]
cptartur: So am I correctly understanding, that even though starknet documentation says that the class hash is
```


The starknet_keccak, of the class’s program. The class’s program is the abi and program part of the .json file that the StarkNet compiler outputs when you run the following command:$ starknet-compile --no_debug_info
--


The [starknet_keccak](https://docs.starknet.io/documentation/architecture_and_concepts/Hashing/hash-functions/#starknet-keccak), of the class’s program. The class’s program is the abi and program part of the .json file that the StarkNet compiler outputs when you run the following command:

$ starknet-compile --no_debug_info
```

It actually doesn't use the format outputted by the compiler (which uses strings for hints keys), but instead it uses the format of `ContractClass` from cairo-lang?

starknet_py/class_hash.py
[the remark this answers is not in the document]
cptartur: Okay, maybe it's just has to be done like this. 

It's kind of the same question as above, should we be really supporting old versions of compiled contract?

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+    # If compiler_version is not present, this was compiled with a compiler before version 0.10.0.
+    # Use "(a : felt)" syntax instead of "(a: felt)" so that the class hash will be the same.
+    if "compiler_version" not in program:
+        program["identifiers"] = _fix_cairo_types(program["identifiers"])
+
+    # ContractClass stores keys of `hints` field as integers, resulting in different sorting order
+    program["hints"] = {int(key): val for key, val in program["hints"].items()}
+
+    class_ = dict(program=program, abi=contract_class.abi)
+    serialized_contract_class = json.dumps(obj=class_, sort_keys=True)
```
cptartur, line 78: Let's verify if we really need to sort the keys.

starknet_py/class_hash.py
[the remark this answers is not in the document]
war-in: It won't. Resolved

starknet_py/class_hash.py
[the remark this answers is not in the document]
drknzz: Setting it to None results in the serialized class containing:

```
"debug_info": null,
```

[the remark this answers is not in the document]
cptartur: If that's required for correct class hash calculation, let's keep it.

starknet_py/class_hash.py
drknzz: Turned out that the output from `starknet-compile` has all keys sorted and the only keys not sorted at this point in code were `program` and `abi` (dict with them is created in the line before).

starknet_py/class_hash.py
drknzz: Any idea where to put it? 👁️ 👄 👁️

starknet_py/class_hash.py
```diff
@@ -0,0 +1,124 @@
...
+from starkware.starknet.public.abi import starknet_keccak
+
+from starknet_py.cairo.felt import encode_shortstring
+from starknet_py.constants import API_VERSION
+from starknet_py.net.client_models import DeclaredContract, EntryPoint
+
+
+def compute_class_hash(contract_class: DeclaredContract) -> int:
+    """
+    Calculate class hash of a declared contract.
```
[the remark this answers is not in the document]
drknzz: What exactly do you mean by a not declared contract?
We do not have a `ContractClass` currently in the code. We do have `DeclaredContract` that is just a dataclass with the same 3 fields (program, entry_points_by_type, abi) as ContractClass from cairo-lang.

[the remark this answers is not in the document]
drknzz: ContractClass from cairo-lang would be used only in this place. Is there a point in creating a **ContractClass** of our own (since we don't want to depend on cairo-lang anymore) and have it's fields defined now (when changes to it are coming)? Maybe we could just go with **TypedDict ContractClass** for now?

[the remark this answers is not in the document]
war-in: Unfortunately, it will be used in the `transactions.py`. `Declare` needs it

starknet_py/contract.py
```diff
@@ -185,6 +180,7 @@ async def deploy(
         """
         # pylint: disable=too-many-arguments
         abi = create_compiled_contract(compiled_contract=self.compiled_contract).abi
+        assert abi is not None
```
drknzz, line 183: [Comment from other PR](https://github.com/software-mansion/starknet.py/pull/686#discussion_r1083806982)

starknet_py/compile/compiler.py
```diff
@@ -62,17 +61,6 @@ def compile_contract(self) -> str:
         )
 
 
-def create_contract_class(
```
war-in, line 65: Is it ok to remove it from here? Can't we just leave this and add the new one somewhere?

starknet_py/net/client_models.py
```diff
@@ -325,7 +325,7 @@ class EntryPointsByType:
 
 
 @dataclass
-class DeclaredContract:
+class ContractClass:
     """
     Dataclass representing contract declared to starknet
     """
```
war-in, line 331: What with the program field? Don't we have to specify this? Would it be useful?

starknet_py/net/models/transaction.py
```diff
@@ -72,7 +72,9 @@ class Declare(AccountTransaction):
     class.
     """
 
-    contract_class: ContractClass
+    contract_class: ContractClass = field(
+        metadata={"marshmallow_field": fields.Nested(ContractClassSchema())}
```
war-in, line 76: Are you sure `fields.Nested` is required here? And if so please tell me why.

starknet_py/hash/class_hash.py
```diff
@@ -0,0 +1,115 @@
...
+
+from starknet_py.cairo.felt import encode_shortstring
+from starknet_py.constants import API_VERSION
+from starknet_py.hash.utils import _starknet_keccak, compute_hash_on_elements
+from starknet_py.net.client_models import ContractClass, EntryPoint
+
+
+def compute_class_hash(contract_class: ContractClass) -> int:
+    """
+    Calculate class hash of a declared contract.
```
war-in, line 13: Please update the docstring 🙏 (not only declared contract)

starknet_py/compile/compiler.py
drknzz: Right, these should be duplicated for now.

starknet_py/net/client_models.py
drknzz: There are no use cases for typed program currently. Maybe in the future, after 0.11.0.

starknet_py/net/models/transaction.py
drknzz: What alternative do you see here?

starknet_py/hash/class_hash.py
drknzz, line 13:
```suggestion
    Calculate class hash of a ContractClass.
```

starknet_py/compile/compiler.py
```diff
@@ -66,11 +68,11 @@ def create_contract_class(
     compiled_contract: str,
 ) -> ContractClass:
     """
-    Creates ContractDefinition either from already compiled contract
+    Creates ContractClass from already compiled contract.
 
-    :return: a ContractDefinition
+    :return: a ContractClass.
```
cptartur, line 73:
```suggestion
    :return: a ContractClass instance.
```
maybe?

starknet_py/contract.py
```diff
@@ -646,10 +642,11 @@ def compute_address(
         compiled = create_compiled_contract(
             compilation_source, compiled_contract, search_paths
         )
+        assert compiled.abi is not None
```
cptartur, line 645: I still don't like the fact that `create_compiled_contract` returns an Optional abi. I don't think there's ever the case that ABI is None when building `ContractClass` from the source code. 

Having `assert compiled.abi is not None` in multiple places of the library is the proof that this is not an ergonomic approach.

starknet_py/hash/class_hash.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if "attributes" in program:
+        program = _delete_attributes(program)
+
+    # If compiler_version is not present, this was compiled with a compiler before version 0.10.0.
+    # Use "(a : felt)" syntax instead of "(a: felt)" so that the class hash will be the same.
+    if "compiler_version" not in program:
+        program["identifiers"] = _fix_cairo_types(program["identifiers"])
+
+    class_ = dict(abi=contract_class.abi, program=program)
```
cptartur, line 73: Why not use a dict literal here? 
```suggestion
    class_ = {"abi": contract_class.abi, "program": program}
```

starknet_py/hash/class_hash.py
```diff
@@ -0,0 +1,115 @@
...
+        identifiers[name] = _fix_cairo_types(value)
+
+    return identifiers
+
+
+def _add_backward_compatibility_space(cairo_type: str) -> str:
+    return re.sub(r"(?<! ):", " :", cairo_type)
+
+
+def _delete_attributes(program) -> dict:
```
cptartur, line 101: Maybe let's rename it to something else? Deleting attributes is a small part of this function.

cptartur (changes requested):

starknet_py/net/client_models.py
war-in: Ok 👌

starknet_py/net/models/transaction.py
war-in: Just `ContractClassSchema()`?

drknzz: Well it has to be a `dataclasses.field`, doesn't it? It wouldn't work with just
```
contract_class: ContractClass = ContractClassSchema()
```
:/

war-in: Ok, you're right

starknet_py/contract.py
drknzz: I fully agree. However the asserts are used only after `create_compiled_contract`, which is already **deprecated**.
Introducing a subclass (probably with its schema as well) and extra logic would just increase complexity of a function that is already going to be removed.

cptartur: Does the same problem exist if using `create_contract_class`? Because it's here to stay I think, and if abi is typed to be `Optional` the same issue will exist even when compiling contracts with external tool.

war-in (approved):

starknet_py/contract.py
drknzz: https://github.com/software-mansion/starknet.py/pull/708

starknet_py/net/client_models.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
cptartur, line 2: Is it still necessary in this module?

cptartur (approved):

starknet_py/net/client_models.py
drknzz: No, good catch 👍

PR #665, opened 2023-01-13
Update ProxyChecks

drknzz:
Closes #

## Introduced changes

Update resolving proxy contracts to be more flexible.
- `ProxyChecks` implement the logic without catching exceptions, they are now caught a level higher
- fixed a bug where, when given a list of proxy_checks, if the first proxy_check that was found invalid implementation, the resolving would throw
- fix `isort` - `black` compliance

##


starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -97,34 +102,58 @@ async def resolve_abi(self) -> Abi:
...
+            try:
+                if implementation_type == ImplementationType.CLASS_HASH:
+                    contract_class = await self.client.get_class_by_hash(implementation)
+                else:
+                    contract_class = await self._get_class_by_address(
+                        address=implementation
+                    )
+
+                if contract_class.abi is None:
+                    raise AbiNotFoundError()
```
cptartur, line 116: Doesn't it produce the same bug as before? This will raise if the first implementation from generator is not working.

starknet_py/proxy/proxy_check.py
```diff
@@ -38,48 +35,18 @@ class ArgentProxyCheck(ProxyCheck):
...
-        return implementation
-
-    @staticmethod
-    def _get_implementation_call(address: Address) -> Call:
-        return Call(
-            to_addr=address,
-            selector=get_selector_from_name("get_implementation"),
-            calldata=[],
-        )
+        return await self.implementation_address(address, client)
```
cptartur, line 49: How do we distinguish if the returned value was address or class hash now?

starknet_py/tests/e2e/docs/guide/test_resolving_proxies.py
```diff
@@ -12,21 +28,54 @@ async def test_resolving_proxies(
...
     # After that contract can be used as usual
+
+    # To resolve proxy contract other than OpenZeppelin or ArgentX, a custom ProxyCheck is needed
+    # The ProxyCheck below resolves proxy contracts which have implementation
+    # stored in implementation() function as address
+    class CustomProxyCheck(ProxyCheck):
+        async def implementation_address(
+            self, address: Address, client: Client
+        ) -> Optional[int]:
+            return None
```
cptartur, line 52: A comment about why it is returning none could be useful here considering that it is in documentation.

starknet_py/tests/e2e/fixtures/proxy.py
```diff
@@ -70,6 +70,21 @@ async def deploy_proxy_to_contract_custom(
     )
 
 
+@pytest_asyncio.fixture(
+    params=[("oz_proxy_exposed_compiled.json", "map_compiled.json")]
+)
+async def deploy_proxy_to_contract_exposed(
```
cptartur, line 76: Maybe a better name would be nice? We could rename `deploy_proxy_to_contract` function to something else and name this fixture `deploy_proxy_to_contract`.

starknet_py/proxy/contract_abi_resolver.py
drknzz: This line means that, some contract_class has been found (either through `self.client.get_class_by_hash` or `self._get_class_by_address`) but the abi is None. It should not happen but `abi` field in **DeclaredContract** is *optional* (I think it's because of it being optional in RPC spec? - a contract without abi does not make much sense though).

starknet_py/proxy/proxy_check.py
drknzz: We don't.
Higher level will try to get class_hash of implementation by calling `ProxyCheck.implementation_hash` and then, if no success, will try to get address by calling `ProxyCheck.implementation_address`.

starknet_py/tests/e2e/fixtures/proxy.py
drknzz: There already is a function `deploy_proxy_to_contract`.
Solving this would require indirect parametrization + it is used twice sometimes in one test and I don't know if it's possible to even have 2 variables parametrized by the same fixture in one test.

Maybe we could just leave it

cptartur: It's not a big problem. It's just a bit weird name.

starknet_py/proxy/proxy_check.py
cptartur: Okay, makes sense. But if a user wants to define a custom proxy check where they distinguish between hash and address, that would still work right?

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -97,34 +102,58 @@ async def resolve_abi(self) -> Abi:
...
         """
-        impl = await self._get_implementation_from_proxy()
-        implementation, implementation_type = impl
-
-        if implementation_type == ImplementationType.CLASS_HASH:
-            contract_class = await self.client.get_class_by_hash(implementation)
-        else:
-            contract_class = await self._get_class_by_address(address=implementation)
+        implementation_generator = self._get_implementation_from_proxy()
+        async for implementation, implementation_type in implementation_generator:
```
war-in, line 106: Does someone have a better idea for the name of the `implementation` variable? I do not have one, but `implementation = address | class_hash` and it wasn't obvious for me :/ maybe a comment will be enough

starknet_py/proxy/proxy_check.py
war-in: Shouldn't we return `class_hash` in the `implementation_hash`? It is confusing when `implementation_hash` returns an address 😕

drknzz: **OpenZeppelinProxyCheck** has different implementations of `implementation_address` and `implementation_hash`

drknzz: > Shouldn't we return class_hash in the implementation_hash? It is confusing when implementation_hash returns an address 😕

The methods are just instructions for the resolver on how to retrieve the class_hash or address from the Proxy.

war-in: I know, but I can't see the profit of making `implementation_hash` returning an address

drknzz: It just means that the methods have the same logic. If you mean that it does not make much sense for the `implementation_hash` method to return some *address*, you're right.
I will add one more function and call it from both `implementation_hash` and `implementation_address`.

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -105,6 +105,7 @@ async def resolve_abi(self) -> Abi:
         implementation_generator = self._get_implementation_from_proxy()
         async for implementation, implementation_type in implementation_generator:
             try:
+                # implementation is either a class_hash or address
```
war-in, line 108: Please move it above the `async for` 🙏

war-in (approved):

starknet_py/proxy/contract_abi_resolver.py
cptartur: Okay, makes sense. Do you think a comment would be useful here?

starknet_py/tests/e2e/docs/guide/test_resolving_proxies.py
```diff
@@ -12,21 +28,56 @@ async def test_resolving_proxies(
...
     # After that contract can be used as usual
+
+    # To resolve proxy contract other than OpenZeppelin or ArgentX, a custom ProxyCheck is needed
+    # The ProxyCheck below resolves proxy contracts which have implementation
+    # stored in implementation() function as address
+    class CustomProxyCheck(ProxyCheck):
+        async def implementation_address(
+            self, address: Address, client: Client
+        ) -> Optional[int]:
+            # Note that None is returned, since the Proxy uses
```
cptartur, line 52:
```suggestion
            # Note that None is returned, since our custom Proxy uses
```

cptartur (approved):

PR #673, opened 2023-01-17
Move migration guide into a single page

Solpatium:
Migration guide is a single page:
image

This is how page navigation looks:
image

Added a line between sections:
image


docs/migration_guide.rst
```diff
@@ -59,28 +64,36 @@ Breaking changes
...
 |
+.. raw:: html
+
+   <hr>
 |
 
+**********************
 0.12.0 Migration guide
-======================
+**********************
```
drknzz, line 75: The docs build failed because of warnings.
Maybe we could do
```
|

----

|
```

I think these lines would be less visible though, so maybe we could also remove "Migration guide" from every section and leave just the version, so it would be more visible.

drknzz:
Do we plan on having the latest version migration guide on the top level? @cptartur

drknzz (approved):

war-in (approved):

PR #674, opened 2023-01-17
Split guide into multiple pages

Solpatium:
image


war-in (approved):

docs/guide/serialization.rst
```diff
@@ -0,0 +1,94 @@
+#############
+Serialization
```
cptartur, line 2: Not sure about naming this Serialization. Maybe there's some better name for this page?

Solpatium: This page covers how to serialize data to other formats (cairo, python). I think it is a good name. Will be even a better fit once new serialization module is merged.
Any suggestions?

cptartur: Let's keep it as `Serialized` for now. We can change it if there's some confusion about it.

cptartur (approved):

PR #676, opened 2023-01-17
Code examples! (demo)

### LINKED ISSUE #383: Code examples in the documentation

When users are reading an [API](https://starknetpy.readthedocs.io/en/latest/api.html) they do not see how to use this methods.

Code example below every method

war-in:
Closes #383 

## Introduced changes

- Adds decorator to include code_examples in the API docs

Check out `API -> Deployer`

##


drknzz (commented):
Great :fireworks:, some thoughts:
> 1. I'd prefer we keep snake case file naming style. It should be easy to implement a function that changes snake case into camel case ([like here for example](https://stackoverflow.com/a/19053800/19725666))
> 2. I don't think code examples are necessary for each function. For example I would keep only the constructor code example as there happens something more than just calling the function with basic arguments.
> 3. If we changed `literalinclude` to `codesnippet`, would including multiple parts of one test work?

war-in:
> 2. I don't think code examples are necessary for each function. For example I would keep only the constructor code example as there happens something more than just calling the function with basic arguments.

How should we define which examples are necessary? Also, it is not trivial to find out methods without examples in the decorator :/

cptartur (commented):
It require a bit of effort to maintain these examples, but they could be useful.

cptartur:
> Great 🎆, some thoughts:
> 
> > 1. I'd prefer we keep snake case file naming style. It should be easy to implement a function that changes snake case into camel case ([like here for example](https://stackoverflow.com/a/19053800/19725666))

Where do you want to use snake case names? 

> > 2. I don't think code examples are necessary for each function. For example I would keep only the constructor code example as there happens something more than just calling the function with basic arguments.

They are not necessary, but if were to start adding examples at all ideally we'd eventually have them for all functions. As @war-in it'd be difficult to decide what functions should have examples and what should not.

drknzz:
> > Great 🎆, some thoughts:
> > > 1. I'd prefer we keep snake case file naming style. It should be easy to implement a function that changes snake case into camel case ([like here for example](https://stackoverflow.com/a/19053800/19725666))
> 
> Where do you want to use snake case names?

Currently, the class names are are contained in the file name 1:1, meaning they are camelCase (see `test_Deployer.py`).

> > > 2. I don't think code examples are necessary for each function. For example I would keep only the constructor code example as there happens something more than just calling the function with basic arguments.
> 
> They are not necessary, but if were to start adding examples at all ideally we'd eventually have them for all functions. As @war-in it'd be difficult to decide what functions should have examples and what should not.

We've agreed with @war-in that function examples like this:
```
    contract_deployment = deployer.create_deployment_call(
        class_hash=0x123, salt=1, abi=abi, calldata=calldata
    )
```
do not give much insight. Maybe if each argument was typed inline, it would make more sense.

Other thing is, if a method uses multiple nested structures, it would be hard to show its execution without showing the whole context (creating the structures before).

cptartur:
> > > Great 🎆, some thoughts:
> > > > 1. I'd prefer we keep snake case file naming style. It should be easy to implement a function that changes snake case into camel case ([like here for example](https://stackoverflow.com/a/19053800/19725666))
> > 
> > 
> > Where do you want to use snake case names?
> 
> Currently, the class names are are contained in the file name 1:1, meaning they are camelCase (see `test_Deployer.py`).

Okay, it makes sense.

> 
> > > > 2. I don't think code examples are necessary for each function. For example I would keep only the constructor code example as there happens something more than just calling the function with basic arguments.
> > 
> > 
> > They are not necessary, but if were to start adding examples at all ideally we'd eventually have them for all functions. As @war-in it'd be difficult to decide what functions should have examples and what should not.
> 
> We've agreed with @war-in that function examples like this:
> 
> ```
>     contract_deployment = deployer.create_deployment_call(
>         class_hash=0x123, salt=1, abi=abi, calldata=calldata
>     )
> ```
> 
> do not give much insight. Maybe if each argument was typed inline, it would make more sense.
> 
> Other thing is, if a method uses multiple nested structures, it would be hard to show its execution without showing the whole context (creating the structures before).

Fair argument about types. But while for things like calldata we could show `[1, 2, 3]`, for more complicated types like `abi` it gets difficult.

starknet_py/utils/decorators/code_examples/code_examples.py
```diff
@@ -0,0 +1,51 @@
...
+        
+            .. codesnippet:: ../../starknet_py/tests/e2e/docs/code_examples/{file_name}
+                :language: python
+                :start-after: docs: {method_name}_start
+                :end-before: docs: {method_name}_end
+                :dedent: 4
+        """
+
+
+def _extract_file_properties(class_name: str) -> Tuple[str, str]:
```
drknzz, line 42: Some simple docstrings would be useful (at least specifying what the functions return - here for example file properties are the name and its content)

starknet_py/utils/decorators/code_examples/code_examples.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def add_code_examples(original_class: Any) -> Any:
+    base_class = original_class.__base__
+
+    file_name, file_content = _extract_file_properties(original_class.__name__)
+
+    for method_name, method in original_class.__dict__.items():
+        stripped_method_name = method_name.strip("_")
+        if f"""docs: {stripped_method_name}_start""" in file_content:
```
drknzz, line 14: Maybe it would be easier to track the functions if it was something like `docs-start: method_name`?

starknet_py/utils/decorators/code_examples/code_examples.py
```diff
@@ -0,0 +1,51 @@
...
+                method.__doc__ = (method.__doc__ or "") + hint
+
+    return original_class
+
+
+def create_hint(file_name: str, method_name: str) -> str:
+    return f"""
+        .. hint::
+        
+            .. codesnippet:: ../../starknet_py/tests/e2e/docs/code_examples/{file_name}
```
drknzz, line 34: Could use the new root path here

war-in: Unfortunately no, codesnippet adds a `docs` path to the path, so it has to be relative

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,94 @@
...
+
+from docutils.nodes import Node
+from sphinx.ext.autodoc.directive import AutodocDirective
+
+from starknet_py.constants import ROOT_PATH
+
+
+class AutoclassWithExamples(AutodocDirective):
+    """
+    Directive class extending AutodocDirective and adding code examples.
```
drknzz, line 13: Let's maybe write a short description of how the class works (like the fact that it adds raw `codesnippet` and uses code examples defined in ...../code_examples)

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    def run(self) -> List[Node]:
+        original_class = getattr(
+            sys.modules[self.env.ref_context.get("py:module")], self.arguments[0]
+        )
+        add_code_examples(original_class)
+
+        self.name = self.name.replace("-with-examples", "")  # remove `-with-examples`
```
drknzz, line 22: Maybe we could make this a static method `get_class_name`?

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,94 @@
...
+
+            # if method does not have __doc__ take it from the base method
+            if (
+                callable(method)
+                and method.__doc__ is None
+                and method_name in base_class.__dict__
+            ):
+                method.__doc__ = getattr(base_class, method_name).__doc__ + hint
+            elif callable(method):
+                method.__doc__ = (method.__doc__ or "") + hint
```
drknzz, line 48: Maybe something like this would be easier to understand
```
if not callable:
   continue
stripped_method_name = method_name.strip("_")
if exists_code_example(stripped_method_name, file_content):
   hint = create_hint
   append_hint(method, hint)
```

docs/api/abi.rst
```diff
@@ -21,6 +21,9 @@ Model
     :exclude-members: __init__, __new__
     :members: defined_structures, functions, constructor, l1_handler, events
 
-.. autoclass:: starknet_py.abi.Abi.Function
+.. py:module:: starknet_py.abi.Abi
+    :noindex:
```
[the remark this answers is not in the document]
war-in: Yes, it specifies the module for the next `autoclass`es. Will be important when `autoclass-with-examples` is used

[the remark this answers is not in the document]
drknzz: And the `:noindex:` part? It is used only in this `py:module`

docs/_ext/autoclass_with_examples.py
war-in: Not really, AutodocDirective is using it this way

docs/api/abi.rst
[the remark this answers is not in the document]
war-in: Sphinx is complaining about starknet_py.abi.Abi usage. It says this is added two times and suggested adding :noindex: to one of them

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,103 @@
...
+
+        stripped_method_name = method_name.strip("_")
+        if _code_example_exists(stripped_method_name, file_content):
+            hint = create_hint(file_name, stripped_method_name)
+
+            # if method does not have __doc__ take it from the base method
+            if method.__doc__ is None and method_name in base_class.__dict__:
+                method.__doc__ = getattr(base_class, method_name).__doc__ + hint
+            else:
+                method.__doc__ = (method.__doc__ or "") + hint
```
drknzz, line 53: Can we have this as a separate method? 5 indents is a little much 🙈

drknzz (approved):

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,103 @@
...
+    return re.sub(r"(?<!^)(?=[A-Z])", "_", text).lower()
+
+
+def setup(app) -> Dict[str, Any]:
+    app.add_directive("autoclass-with-examples", AutoclassWithExamples)
+
+    return {
+        "version": "0.1",
+        "parallel_read_safe": True,
+        "parallel_write_safe": True,
```
cptartur, line 107: Always unsure about this. Is it really parallel safe if we're modifying files?

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,103 @@
...
+
+    This class runs before the AutodocDirective class and pulls code snippets
+    from the starknet_py/tests/e2e/docs/code_examples directory to include in the documentation.
+    This allows developers to easily see and understand how the methods
+    being documented are intended to be used in a practical context.
+    """
+
+    def run(self) -> List[Node]:
+        original_class = getattr(
+            sys.modules[self.env.ref_context.get("py:module")], self.arguments[0]
```
cptartur, line 24: Some comment about this could be useful, I think this is one of the least obvious lines in the whole PR.

docs/api/abi.rst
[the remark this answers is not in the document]
cptartur: I'm not sure if I understand what it does. Will we have to add it to other modules?

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def _append_hint(method_name: str, method, base_class: Any, hint: str):
+    """
+    If method does not have the __doc__, takes it from the base method.
+    """
+    if method.__doc__ is None and method_name in base_class.__dict__:
+        method.__doc__ = getattr(base_class, method_name).__doc__ + hint
+    else:
+        method.__doc__ = (method.__doc__ or "") + hint
```
cptartur, line 58: This wouldn't work if the docstring was several classes "above" in the hierarchy, right?

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,108 @@
...
+
+def _camel_to_snake(text: str) -> str:
+    """
+    Transforms camelCase to the snake_case.
+    """
+    return re.sub(r"(?<!^)(?=[A-Z])", "_", text).lower()
+
+
+def _code_example_exists(method_name: str, file_content: str):
+    return f"""test_{method_name}(""" in file_content
```
cptartur, line 83: Why use multiline string here?

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def _create_hint(file_name: str, method_name: str) -> str:
+    """
+    Constructs a hint with code example.
+    """
+    return f"""
+        .. hint::
```
cptartur, line 91: Not sure about using hint for this. Couldn't it be a normal code block as a part of documentation? Or something else that is not a hint?

docs/api/abi.rst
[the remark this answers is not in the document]
cptartur: This kind of seems like a workaround.

starknet_py/constants.py
```diff
@@ -10,3 +12,5 @@
 RPC_CONTRACT_NOT_FOUND_ERROR = 20
 RPC_INVALID_MESSAGE_SELECTOR_ERROR = 21
 RPC_CLASS_HASH_NOT_FOUND_ERROR = 28
+
+ROOT_PATH = Path(__file__).parent
```
cptartur, line 16: Isn't the result of this dependent on how `constants.py` is imported?

docs/_ext/autoclass_with_examples.py
drknzz: [Discussion in Codesnippet](https://github.com/software-mansion/starknet.py/pull/485/files#r1012775772)

docs/_ext/autoclass_with_examples.py
drknzz: It is a *hint* on how to use it though.
Plus hints have a nice, more contrasty color

starknet_py/constants.py
drknzz: We could add an **assert** below
```
assert ROOT_PATH.name == "starknet_py"
```

docs/_ext/autoclass_with_examples.py
war-in: Yes. Since we do not currently use deep inheritance I would not implement this

starknet_py/constants.py
war-in: > Isn't the result of this depends on how constants.py is imported?

It is but constants.py is in the top-level directory. Should I add an `assert` as @drknzz sad?

docs/api/abi.rst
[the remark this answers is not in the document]
cptartur: Okay, makes sense? Why the `noindex` though?

docs/_ext/autoclass_with_examples.py
cptartur: Maybe let's at least open issue for that so we don't forget? I can imagine the situation where docs are missing because of this.

docs/_ext/autoclass_with_examples.py
cptartur: If you both think its okay, we can leave it as a hint. It just doesn't make 100% sense to me to have a hint that is just an example usage.

starknet_py/constants.py
cptartur: Isn't this 
```py
assert ROOT_PATH.name == "starknet_py"\
```
pretty much equal to just hardocoding the path (i.e. not using `__file__` and parent) at all?

docs/_ext/autoclass_with_examples.py
war-in: I can rename it to `Example`

starknet_py/constants.py
war-in: You're right. I will rewrite that

war-in: The problem with hardcoding is that we can hardcode only the part of the path ("starknet_py"). Path(__file__) takes the entire path from the Home directory which is more useful

docs/_ext/autoclass_with_examples.py
war-in: Although, I agree it should be somehow `remembered` I don't think the issue is the best place for that. I can add a comment or include information about it in the docstring.

docs/_ext/autoclass_with_examples.py
war-in: image

starknet_py/constants.py
drknzz: > Isn't this
> 
> ```python
> ROOT_PATH = Path(__file__).parent
> assert ROOT_PATH.name == "starknet_py"\
> ```
> 
> pretty much equal to just hardocoding the path (i.e. not using `__file__` and parent) at all?

So just
```
ROOT_PATH = "starknet_py"
```
?

Would this be any useful? Would this work here?
```
file_path = ROOT_PATH / "tests/e2e/docs/code_examples" / file_name
```

war-in: Nope, it wouldn't. (tested that)

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,116 @@
...
+    Extracts file content for given class name.
+
+    :param class_name: A string representing the name of the class where examples will be added to.
+    :returns: A tuple containing the name of the file as a string, and its content as a string.
+    """
+    file_name = "test_" + _camel_to_snake(class_name) + ".py"
+    file_path = ROOT_PATH / "tests/e2e/docs/code_examples" / file_name
+
+    with open(file_path, "r", encoding="utf-8") as file:
+        return file_name, file.read()
```
cptartur, line 69: wouldn't it be sufficient to just do `file_path.read_text("utf-8")?

docs/_ext/autoclass_with_examples.py
```diff
@@ -0,0 +1,116 @@
...
+    add code examples to the method docstrings in the documentation.
+
+    This class runs before the AutodocDirective class and pulls code snippets
+    from the starknet_py/tests/e2e/docs/code_examples directory to include in the documentation.
+    This allows developers to easily see and understand how the methods
+    being documented are intended to be used in a practical context.
+
+    .. note::
+
+        The docstrings will not be inherited if the hierarchy is `deeper` than one step.
```
cptartur, line 23: Let's create an issue for that.

war-in: https://github.com/software-mansion/starknet.py/issues/705

cptartur (approved):

PR #677, opened 2023-01-17
Hash module

drknzz:
## Introduced changes

- Create `starknet_py.hash` module
- **starknet.js** uses similiar [structure](https://github.com/0xs34n/starknet.js/blob/develop/src/utils/hash.ts)

##

| Function | Old Path | New Path |
| ------------- | ------------- | -- |
| compute_address | starknet_py.net.models.address  | starknet_py.hash.address |
| get_selector_from_name |  starknet_py.cairo.selector  | starknet_py.hash.selector |
| get_storage_var_address |  starknet_py.cairo.storage  | starknet_py.hash.storage |
| transaction hashes |  starknet_py.utils.crypto.transaction_hash  | starknet_py.hash.transaction |
| _starknet_keccak |  starknet_py.cairo.utils  | starknet_py.hash.utils |
| |   | |
| compute_hash_on_elements |  starknet_py.utils.crypto.facade  | starknet_py.hash.utils |
| message_signature |  starknet_py.utils.crypto.facade  | starknet_py.hash.utils |
| pederse_hash |  starknet_py.utils.crypto.facade  | starknet_py.hash.utils |


starknet_py/hash/utils.py
```diff
@@ -0,0 +1,40 @@
...
+from starknet_py.common import int_from_bytes
+from starknet_py.constants import MASK_250
+from starknet_py.utils.crypto.facade import use_cpp_variant
+
+
+def _starknet_keccak(data: bytes) -> int:
+    """
+    A variant of eth-keccak that computes a value that fits in a StarkNet field element.
+    """
+    return int_from_bytes(keccak(data)) & MASK_250
```
cptartur, line 18: Do we want to keep this a private function?

starknet_py/hash/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def pedersen_hash(left: int, right: int) -> int:
```
cptartur, line 21: This could use a docstring

starknet_py/hash/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    if use_cpp_variant():
+        return cpp_hash(left, right)
+    return default_hash(left, right)
```
cptartur, line 24: Let's add some kind of test for it.

starknet_py/hash/utils.py
```diff
@@ -0,0 +1,40 @@
...
+        h(h(h(h(0, data[0]), data[1]), ...), data[n-1]), n).
+
+    The hash is initialized with 0 and ends with the data length appended.
+    The length is appended in order to avoid collisions of the following kind:
+    H([x,y,z]) = h(h(x,y),z) = H([w, z]) where w = h(x,y).
+    """
+    return functools.reduce(pedersen_hash, [*data, len(data)], 0)
+
+
+def message_signature(msg_hash, priv_key, seed: Optional[int] = 32) -> ECSignature:
```
cptartur, line 39: Docstring

cptartur (commented):
I like the changes. 👍 

I'm not 100% sold on the `utils` name. Technically I think it'd make sense to name this file `hash` but then we'd have `from ...hash.hash` imports which I don't really like. 

If there are some alternatives to `utils` you've considered I'd be interesting in seeing them though.

starknet_py/hash/utils.py
drknzz: Brother 💀 I was just moving stuff around in this pr.

starknet_py/hash/utils.py
drknzz: Any reasons for why we should make it public?

cptartur: I wonder, probably no.

drknzz:
> I like the changes. 👍
> 
> I'm not 100% sold on the `utils` name. Technically I think it'd make sense to name this file `hash` but then we'd have `from ...hash.hash` imports which I don't really like.
> 
> If there are some alternatives to `utils` you've considered I'd be interesting in seeing them though.

How about `common`? They are not that common though. And I think utils is better than that

cptartur:
> > I like the changes. 👍
> > I'm not 100% sold on the `utils` name. Technically I think it'd make sense to name this file `hash` but then we'd have `from ...hash.hash` imports which I don't really like.
> > If there are some alternatives to `utils` you've considered I'd be interesting in seeing them though.
> 
> How about `common`? They are not that common though. And I think utils is better than that

I also prefer `utils` to `common`.

starknet_py/hash/contract_class.py
```diff
@@ -2,10 +2,9 @@
 from typing import List
```
war-in, line 2: Why this file is named `contract_class`?

starknet_py/hash/address.py
```diff
@@ -0,0 +1,35 @@
+from typing import Sequence
```
war-in, line 1: Do you think having two `address` files is ok?

war-in (commented):
I like the structure 🔥 
Although `utils` is not the best name, it is located in the `hash` directory (fair enough).

starknet_py/hash/address.py
drknzz: As they are in different places and contain other logic, I'd say yes

starknet_py/hash/contract_class.py
drknzz: It cannot be called `class.py` as it's some internal name 😿

war-in: I would name it `class_hash`. It is a very well-known name in StarkNet and import 
```from starknet.py.hash.class_hash import compute_class_hash```
isn't too bad

war-in (approved):
Just rename `contract_class` to `class_hash` and it's perfect 🥇

starknet_py/hash/contract_class.py
drknzz: @cptartur, you mentioned the repeating names in module path above, what do you think?

cptartur (approved):

starknet_py/hash/contract_class.py
cptartur: I think it's okay in this case. It'd be worse if it was `hash.hash`.

cptartur: Let's name it as @war-in suggested.

drknzz: Okay, I will do it in next pr though

PR #683, opened 2023-01-20
Replace compress and decompress program

### LINKED ISSUE #658: Replace compress_program and decompress_program
### Feature Request

It should be done after #649 is merged

cptartur:
Closes #658 

## Introduced changes

- Replaces compress and decompress program from cairo lang

##

- Decompress program failure will no longer raise a StarkException

### DIFF
--- a/starknet_py/net/models/transaction.py
+++ b/starknet_py/net/models/transaction.py
@@ -88,7 +88,11 @@ class Declare(AccountTransaction):
         name = Declare._contract_attribute_name(data)
 
         program = data[name]["program"]
-        data[name]["program"] = Declare._compress_program(program)
+
+        compressed_program = json.dumps(program)
+        compressed_program = gzip.compress(data=compressed_program.encode("ascii"))
+        compressed_program = base64.b64encode(compressed_program)
+        data[name]["program"] = compressed_program.decode("ascii")
 
         return data
 
@@ -125,14 +129,6 @@ class Declare(AccountTransaction):
         return (
             "contract_definition" if "contract_definition" in data else "contract_class"
         )
-
-    @staticmethod
-    def _compress_program(program: Dict[str, Any]):
-        compressed_program = json.dumps(program)
-        compressed_program = gzip.compress(data=compressed_program.encode("ascii"))
-        compressed_program = base64.b64encode(compressed_program)
-
-        return compressed_program.decode("ascii")
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


starknet_py/net/models/transaction.py
```diff
@@ -105,6 +120,20 @@ def calculate_hash(self, chain_id: StarknetChainId) -> int:
             nonce=self.nonce,
         )
 
+    @staticmethod
+    def _contract_attribute_name(data: Dict[str, Any]):
+        return (
+            "contract_definition" if "contract_definition" in data else "contract_class"
```
drknzz, line 126: I think it's backward compatibility, no?

cptartur: It is. It's not expensive to have it though, so I think it's fine.

drknzz (approved):

starknet_py/net/models/transaction.py
```diff
@@ -105,6 +120,20 @@ def calculate_hash(self, chain_id: StarknetChainId) -> int:
...
+        )
+
+    @staticmethod
+    def _compress_program(program: Dict[str, Any]):
+        compressed_program = json.dumps(program)
+        compressed_program = gzip.compress(data=compressed_program.encode("ascii"))
+        compressed_program = base64.b64encode(compressed_program)
+
+        return compressed_program.decode("ascii")
```
war-in, line 135: Do we really need these two as static methods?

war-in: Ok, I see what is the problem. Let's keep it as it is

war-in (approved):

starknet_py/net/models/transaction.py
cptartur: Compress program could be inlined. Attribute name would be duplicated in 2 places.

starknet_py/net/models/transaction.py
```diff
@@ -105,6 +124,12 @@ def calculate_hash(self, chain_id: StarknetChainId) -> int:
             nonce=self.nonce,
         )
 
+    @staticmethod
+    def _contract_attribute_name(data: Dict[str, Any]):
+        return (
+            "contract_definition" if "contract_definition" in data else "contract_class"
+        )
```
cptartur, line 131: I wanted to test serializing contract that has `contract_definition` defined, but I went through a couple of old compiled contract I had and none of them had this. It must have been changed a while ago or I am looking in the wrong place.

drknzz (approved):

PR #649, opened 2023-01-09
Add dataclasses for transactions

war-in:
Closes #640 
Closes #634 

## Introduced changes

- Transactions from cairo-lang are replaced with our implementation
- `QUERY_VERSION_BASE` is removed

##

- [ ] This PR contains breaking changes

PR #693, opened 2023-01-23
Add signing transactions for fee estimations

cptartur:
Closes #

## Introduced changes

- This PR adds new method in Account for signing transactions for fee estimation purpose
- Changed StarkCurveSigner logic to only use standard logic for transactions with all version `!= 0`.

##


starknet_py/net/account/base_account.py
```diff
@@ -56,6 +61,18 @@ async def get_balance(
         :return: Token balance.
         """
 
+    @abstractmethod
+    async def sign_for_fee_estimate(
+        self, transaction: TypeAccountTransaction
+    ) -> TypeAccountTransaction:
+        """
+        Sign a transaction for a purpose of fee_estimation.
+        Uses transaction version calculated like ``transaction.version + 2 ** 128``.
```
drknzz, line 70: Should the exact way of calculating the version be specified here?

starknet_py/net/account/account.py
```diff
@@ -229,6 +228,16 @@ async def get_balance(
 
         return (high << 128) + low
 
+    async def sign_for_fee_estimate(
+        self, transaction: TypeAccountTransaction
+    ) -> TypeAccountTransaction:
+        version = self.supported_transaction_version + 2**128
```
drknzz, line 234: Should we add a constant?

drknzz: Also we should test it.

starknet_py/net/account/account.py
```diff
@@ -229,6 +228,16 @@ async def get_balance(
...
+
+        transaction = dataclasses.replace(transaction, version=version)
+        signature = self.signer.sign_transaction(transaction)
+
+        return _add_signature_to_transaction(tx=transaction, signature=signature)
```
drknzz, line 239: We use `dataclasses.replace` in line 236, while line 239 uses the same but wrapped into a function.
Let's maybe change `_add_signature_to_transaction` to also contain line 237 (since this logic is also repeated in a few places)

starknet_py/net/account/base_account.py
war-in: Good point. Maybe `Should use ...`? And a little explanation of why this is a good idea to estimate_fee with a different signature.

starknet_py/net/account/_account_proxy.py
```diff
@@ -36,6 +42,16 @@ async def get_balance(
     ) -> int:
         return await self._account_client.get_balance(token_address=token_address)
 
+    async def sign_for_fee_estimate(
+        self, transaction: TypeAccountTransaction
+    ) -> TypeAccountTransaction:
+        version = self.supported_transaction_version + 2**128
+        signature = self._account_client.signer.sign_transaction(transaction)
```
war-in, line 49: Version should be added to the transaction before calculating a signature

starknet_py/net/account/account.py
war-in: And add the docstring

war-in (changes requested):

starknet_py/net/account/account.py
cptartur: Docstring is already there in the `BaseAccount`.

cptartur: I don't want to define a constant for this, this is Account implementation detail.

starknet_py/net/account/account.py
cptartur: Don't want to change the Account logic more than its necessary in this PR, we can create an issue for that if you think its necessary.

starknet_py/net/account/_account_proxy.py
cptartur: Fixed

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -73,13 +73,13 @@ def _sign_transaction(self, transaction: Invoke):
             version=transaction.version,
             contract_address=self.address,
             entry_point_selector=DEFAULT_ENTRY_POINT_SELECTOR
-            if transaction.version == 1
+            if transaction.version != 0
```
war-in, line 76: What if someone wants to sign a transaction whit `version = 2**128`? It should be created exactly like the `0` version

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -328,3 +329,22 @@ async def test_deploy_account_uses_custom_calldata(
     tx = await client.get_transaction(deploy_result.hash)
     assert isinstance(tx, DeployAccountTransaction)
     assert tx.constructor_calldata == calldata
+
+
+@pytest.mark.asyncio
+async def test_sign_tx_for_fee_estimation(account, map_contract):
+    call = map_contract.functions["put"].prepare(key=40, value=50)
+    transaction = await account.sign_invoke_transaction(calls=call, max_fee=MAX_FEE)
```
war-in, line 337: What about other transactions? `Declare` and `DeployAccount`? They should be tested, because we don't know if the logic works for them

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -328,3 +331,88 @@ async def test_deploy_account_uses_custom_calldata(
...
+
+@pytest.mark.asyncio
+async def test_sign_declare_tx_for_fee_estimation(account, map_compiled_contract):
+    transaction = await account.sign_declare_transaction(
+        compiled_contract=map_compiled_contract, max_fee=MAX_FEE
+    )
+
+    signed_transaction = await account.sign_for_fee_estimate(transaction)
+
+    estimation = await account.client.estimate_fee(signed_transaction)
```
war-in, line 363: That flow is the default for estimating fees now. It means a user has to sign the transaction properly and then resign it for estimation. What if someone only wants to estamete_fee? Isn't it too complex? Shouldn't we add an option to create a transaction from scratch here? Any ideas?

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -139,3 +139,7 @@ def sign_message(
         # pylint: disable=invalid-name
         r, s = message_signature(msg_hash=msg_hash, priv_key=self.private_key)
         return [r, s]
+
+
+def _is_old_transaction_version(version: int):
+    return version in (0, 2**128)
```
drknzz, line 145: This reminds me of the `QUERY_VERSION_BASE` 👀 
https://github.com/software-mansion/starknet.py/pull/649/files#diff-82627af50ed667c0bddd33a217deb26c43ff95cfecca09ec50351d1545e13c81L129

starknet_py/tests/e2e/account/account_test.py
drknzz: eh. Well I suppose if someone estimates the fee of a transaction, they probably do want to send it later.

drknzz: An optional flag would keep the flow the same, but the method would have to be used twice when wanting to send the tx, so no benefit.

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -328,3 +331,88 @@ async def test_deploy_account_uses_custom_calldata(
...
+    else:
+        account = AccountProxy(
+            AccountClient(
+                address=address,
+                client=client,
+                key_pair=key_pair,
+                chain=StarknetChainId.TESTNET,
+                supported_tx_version=1,
+            )
+        )
```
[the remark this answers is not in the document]
cptartur: I agree its not very good. Defining a custom fixture for single tests seemed overkill though. This could be split into two tests, but we would have a bit of duplication though.

starknet_py/tests/e2e/account/account_test.py
cptartur: I think there are little to now use-cases for creating transactions only for purpose of fee estimation.

starknet_py/net/signer/stark_curve_signer.py
cptartur: I does seem to be the same value, but I'm unsure if should be used here. Do we want this logic to still apply if `QUERY_VERSION_BASE` was updated?

starknet_py/net/account/_account_proxy.py
```diff
@@ -36,6 +42,17 @@ async def get_balance(
     ) -> int:
         return await self._account_client.get_balance(token_address=token_address)
 
+    async def sign_for_fee_estimate(
+        self, transaction: TypeAccountTransaction
+    ) -> TypeAccountTransaction:
+        version = self.supported_transaction_version + 2**128
```
drknzz, line 48: Are we sure we do not want to make it a constant value somewhere?
It's used in more than one place, in the same context, implicating a constant.

cptartur: This is an mostly an implementation detail. We could define a constant within Account maybe, but I'm not sure exposing it as a general constant in the library.

starknet_py/tests/e2e/account/account_test.py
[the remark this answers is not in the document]
war-in: But the fixture could be defined above the test. Or a helper function? to clarify the test :)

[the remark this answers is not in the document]
cptartur: Added a helper function.

starknet_py/net/account/_account_proxy.py
cptartur: Once we confirm that using `2**128` is fine for the purpose of non-executable transctions, let's reintroduce `FEE_VERSION_BASE = 2**128` constant and use it there.

war-in (dismissed):

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -328,3 +332,96 @@ async def test_deploy_account_uses_custom_calldata(
...
+    # Verify that the transaction signed for fee estimation cannot be sent
+    with pytest.raises(ClientError):
+        await account.client.declare(estimate_fee_transaction)
+
+    # Verify that original transaction can be sent
+    result = await account.client.declare(transaction)
+    await account.client.wait_for_tx(result.transaction_hash)
+
+
+def _account_by_type(
```
drknzz, line 376: Could use just the class itself instead of string

drknzz (dismissed):

war-in (approved):

drknzz (approved):

war-in (approved):

PR #703, opened 2023-01-26
Update checks

### LINKED ISSUE #700: Split CI tests
### Feature Request

The workflow runs started to take a bit longer.
We could use some inspiration from [actions in protostar](https://github.com/software-mansion/protostar/actions) to speed up the process.

war-in:
Closes #700 

## Introduced changes

This PR splits CI tests into:
- installing dependencies 
- checking format etc
- unit & e2e tests
- docs tests

The last three are running in parallel 🔥 

Comparison:
- before
image

- after
image

##

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,3 +1,5 @@
+name: Checks
+
 on:
   push:
     branches:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -54,6 +54,18 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install poetry
+        run: |
+          python -m pip install --upgrade pip
+          pip install poetry
+
+      - name: Restore caches
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pypoetry
+          key: poetry-${{ hashFiles('poetry.lock') }}
+
       - name: Check formatting with black
         run: |
           poetry run poe format_check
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -8,49 +8,108 @@ on:
...
+      matrix:
+        python-version: [ "3.8", "3.9" ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
```
drknzz, line 56: Why do we set up python in every job? Is setting it in `setup` not enough?

war-in: Nope, it has to be set up for every job

.github/workflows/checks.yml
```diff
@@ -8,49 +8,128 @@ on:
...
+    needs: setup
+    strategy:
+      matrix:
+        python-version: [ "3.8", "3.9" ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Install poetry
+        run: |
```
cptartur, line 46: Remove newline

.github/workflows/checks.yml
```diff
@@ -8,49 +8,128 @@ on:
...
+    runs-on: ubuntu-latest
+    needs: setup
+    strategy:
+      matrix:
+        python-version: [ "3.8", "3.9" ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Install poetry
```
cptartur, line 45: Maybe a comment why we need to reinstall poetry even though we are using cache?

.github/workflows/checks.yml
```diff
@@ -8,49 +8,128 @@ on:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install poetry
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'poetry'
+
```
cptartur, line 124: unnecessary newline

cptartur (approved):

drknzz (approved):
Great. I think we might need to change the required checks though?

PR #708, opened 2023-01-27
Add CompiledContract

drknzz:
## Introduced changes

- Add `CompiledContract` - `ContractClass` subclass, with required `abi`

### DIFF
--- a/starknet_py/common.py
+++ b/starknet_py/common.py
@@ -2,8 +2,8 @@ import warnings
 from typing import List, Literal, Optional, Union, cast
 
 from starknet_py.compile.compiler import Compiler, StarknetCompilationSource
-from starknet_py.net.client_models import CompiledContract, ContractClass
-from starknet_py.net.schemas.gateway import CompiledContractSchema, ContractClassSchema
+from starknet_py.net.client_models import ContractClass
+from starknet_py.net.schemas.gateway import ContractClassSchema
 
 
 def create_compiled_contract(
@@ -28,15 +28,6 @@ def create_compiled_contract(
         ).compile_contract()
     definition = create_contract_class(compiled_contract)
     return definition
-
-
-def _create_compiled_contract(compiled_contract: str) -> CompiledContract:
-    """
-    Creates CompiledContract from already compiled contract.
-
-    :return: a CompiledContract instance.
-    """
-    return cast(CompiledContract, CompiledContractSchema().loads(compiled_contract))
 
 
 def create_contract_class(
--- a/starknet_py/common.py
+++ b/starknet_py/common.py
@@ -2,8 +2,8 @@ import warnings
 from typing import List, Literal, Optional, Union, cast
 
 from starknet_py.compile.compiler import Compiler, StarknetCompilationSource
-from starknet_py.net.client_models import ContractClass
-from starknet_py.net.schemas.gateway import ContractClassSchema
+from starknet_py.net.client_models import CompiledContract, ContractClass
+from starknet_py.net.schemas.gateway import CompiledContractSchema, ContractClassSchema
 
 
 def create_compiled_contract(
@@ -28,6 +28,15 @@ def create_compiled_contract(
         ).compile_contract()
     definition = create_contract_class(compiled_contract)
     return definition
+
+
+def _create_compiled_contract(compiled_contract: str) -> CompiledContract:
+    """
+    Creates CompiledContract from already compiled contract.
+
+    :return: a CompiledContract instance.
+    """
+    return cast(CompiledContract, CompiledContractSchema().loads(compiled_contract))
 
 
 def create_contract_class(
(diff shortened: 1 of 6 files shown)


starknet_py/common.py
```diff
@@ -2,8 +2,8 @@
 from typing import List, Literal, Optional, Union, cast
 
 from starknet_py.compile.compiler import Compiler, StarknetCompilationSource
-from starknet_py.net.client_models import ContractClass
-from starknet_py.net.schemas.gateway import ContractClassSchema
+from starknet_py.net.client_models import CompiledContract, ContractClass
+from starknet_py.net.schemas.gateway import CompiledContractSchema, ContractClassSchema
 
 
 def create_compiled_contract(
```
drknzz, line 9: After this gets removed, `_create_compiled_contract` would be ranamed to `create_compiled_contract` as it's just the perfect name, that is currently taken

starknet_py/net/client_models.py
```diff
@@ -332,7 +332,16 @@ class ContractClass:
 
     program: dict
     entry_points_by_type: EntryPointsByType
-    abi: Optional[AbiDictList] = None
+    abi: Optional[AbiDictList]
```
war-in, line 335: Shouldn't we leave abi default to None?

starknet_py/common.py
```diff
@@ -30,6 +30,15 @@ def create_compiled_contract(
...
+def _create_compiled_contract(compiled_contract: str) -> CompiledContract:
+    """
+    Creates CompiledContract from already compiled contract.
+
+    :return: a CompiledContract instance.
+    """
+    return cast(CompiledContract, CompiledContractSchema().loads(compiled_contract))
+
+
 def create_contract_class(
```
cptartur, line 42: Perhaps this could get deprecated? Are there valid use cases for the user to create `ContractClass`?

starknet_py/common.py
cptartur: It will be problematic to properly deprecate `create_compile_contract` and then suddenly reintroduce it with different parameters.

cptartur: Actually I'd consider an alternative approach, but that would be for a separate bigger issue. 

What if we started introducing methods like this as `@classmethod`s in the models themselves.
```py
CompiledContract:
  ...

  @classmethod
  def from_contract(compiled_contract: str) -> CompiledContract:
    ...
```
(though I'm unsure about the exact name of the method).

starknet_py/net/client_models.py
drknzz: It cannot be `= None` in the current implementation, as **CompiledContract** inherits from **ContractClass** and it has would cause "non default argument after a default argument"

war-in: Right, forget about it

starknet_py/net/client_models.py
```diff
@@ -332,7 +335,25 @@ class ContractClass:
...
+    abi: AbiDictList
+
+    @classmethod
+    def from_contract(cls, contract: str) -> CompiledContract:
+        """
+        Creates CompiledContract from already compiled contract.
+
+        :return: a CompiledContract instance.
+        """
+        return cast(CompiledContract, CompiledContractSchema().loads(contract))
```
cptartur, line 356: Actually, if we're not using `cls` at all here, this could be changed to `@staticmethod`.

Or we could write 
```suggestion
    @classmethod
    def from_contract(cls, contract: str) -> CompiledContract:
        """
        Creates CompiledContract from already compiled contract.

        :return: a CompiledContract instance.
        """
        return cast(cls, CompiledContractSchema().loads(contract))
```

starknet_py/common.py
drknzz: Why not leave it? It's used in some places around the library.

cptartur: I think 99% of these use cases are create by the classmethod now.

war-in (approved):

starknet_py/net/client_models.py
```diff
@@ -332,7 +334,16 @@ class ContractClass:
 
     program: dict
     entry_points_by_type: EntryPointsByType
-    abi: Optional[AbiDictList] = None
+    abi: Optional[AbiDictList]
```
cptartur, line 337: I think technically this is a breaking change, please include it in the migration guide.

cptartur (approved):

starknet_py/net/client_models.py
drknzz: What if we leave `= None` here and make `abi` in **CompiledContract** default to an empty list?

starknet_py/net/client_models.py
```diff
@@ -335,6 +337,15 @@ class ContractClass:
     abi: Optional[AbiDictList] = None
 
 
+@dataclass
+class CompiledContract(ContractClass):
+    """
+    Dataclass representing ContractClass with required abi.
+    """
+
+    abi: AbiDictList = field(default_factory=list)
```
cptartur, line 346: I think we could add some sort of post load verification to assert that abi is not an empty list, otherwise it's an okay approach.

drknzz: I don't think we need to. `abi` is a required key in the schema, and its type is `list`. So the `default_factory=list` actually does nothing. It's just a fix for keyword arguments before positional ones.

PR #711, opened 2023-01-27
Change fee estimate multiplier

### LINKED ISSUE #702: Investigate max_fee not high enough when `auto_estimate`
### Feature Request

Perhaps a multiplier should be increased from `1.1` that it is currently.

drknzz:
Closes #702

## Introduced changes

- Increase automatic estimated fee multiplier from `1.1` to `1.5`


starknet_py/net/account/account_client.py
```diff
@@ -326,7 +326,7 @@ async def _get_max_fee(
 
         if auto_estimate:
             estimate_fee = await self.estimate_fee(transaction)
-            max_fee = int(estimate_fee.overall_fee * 1.1)
+            max_fee = int(estimate_fee.overall_fee * ESTIMATED_FEE_MULTIPLIER)
```
cptartur, line 329: I wouldn't change it in AccountClient: it is deprecated anyways.

starknet_py/net/account/account.py
```diff
@@ -110,7 +111,7 @@ async def _get_max_fee(
 
         if auto_estimate:
             estimate_fee = await self._estimate_fee(transaction)
-            max_fee = int(estimate_fee.overall_fee * 1.1)
+            max_fee = int(estimate_fee.overall_fee * ESTIMATED_FEE_MULTIPLIER)
```
cptartur, line 114: I think we could make `fee_multiplier` a class variable, but don't expose it in constructor. This way `Account` would use default of `1.5` but could adjusted by the user like `account.fee_multiplier = 2.0`.

cc @war-in

cptartur (changes requested):

starknet_py/net/account/account.py
war-in: Yes, I would like to have it too

starknet_py/net/account/account.py
```diff
@@ -55,6 +55,8 @@ class Account(BaseAccount):
     Default Account implementation.
     """
 
+    ESTIMATED_FEE_MULTIPLIER: float = 1.5
```
cptartur, line 58: I wouldn't use uppercase here and perhaps add a docstring.

docs/guide/using_existing_contracts.rst
```diff
@@ -61,7 +61,7 @@ Automatic fee estimation
 
 For testing purposes it is possible to enable automatic fee estimation when making
 a transaction. Starknet.py will then use ``estimate_fee()`` internally and use value
-returned by it multiplied by ``1.1`` as a ``max_fee``.
+returned by it multiplied by ``1.5`` as a ``max_fee``.
```
cptartur, line 64: Would add information about adjusting the default multiplier here or somewhere else.

starknet_py/net/account/account.py
drknzz: https://stackoverflow.com/a/45702782

cptartur: But is it really a constant if it is intended to be changed by the user?

docs/guide/using_existing_contracts.rst
drknzz: Adjusting? You mean the fact that it has been changed or? It cannot be adjusted manually

starknet_py/net/account/account.py
drknzz: It's not changeable yet though

docs/guide/using_existing_contracts.rst
cptartur: I'd like it to be adjustable by the user. That's why I wanted the  `ESTIMATED_FEE_MULTIPLIER` to be a part of Account now.

starknet_py/net/account/account.py
drknzz: Well it is if one sets account.attr = 2.0

war-in (approved):

starknet_py/net/account/account.py
```diff
@@ -53,8 +53,12 @@
 class Account(BaseAccount):
     """
     Default Account implementation.
+
+    :ivar ESTIMATED_FEE_MULTIPLIER: Amount by which each estimated fee is multiplied when using `auto_estimate`.
     """
 
+    ESTIMATED_FEE_MULTIPLIER: float = 1.5
```
cptartur, line 60: Wouldn't the docs for this generate correctly if you just added a docstring directly below `ESTIMATED_FEE_MULTIPLIER`, without the `:ivar ESTIMATED_FEE_MULTIPLIER:`?

That's how it worked with dataclasses.

docs/guide/using_existing_contracts.rst
```diff
@@ -71,3 +71,8 @@ returned by it multiplied by ``1.1`` as a ``max_fee``.
 .. code-block:: python
 
     await contract.functions["put"].invoke(k, v, auto_estimate=True)
+
+
+.. note::
+    It is possible to set the value by which the estimated fee is multiplied,
```
cptartur, line 77:
```suggestion
    It is possible to configure the value by which the estimated fee is multiplied,
```

cptartur (approved):

starknet_py/net/account/account.py
drknzz: I think it wouldn't show at all in the class

cptartur:
```py
ESTIMATED_FEE_MULTIPLIER: float = 1.5
"""Amount by which each estimated fee is multiplied when using `auto_estimate`."""
```
should work and display as 
image

PR #714, opened 2023-01-30
Fix API docs (order and undocumented members)

war-in:
Closes #

## Introduced changes

- Adds `groupwise` order (sync methods are just below `async` ones.
- Methods/parameters without docstrings are now present in the API docs.

##

### DIFF
--- a/docs/api/abi.rst
+++ b/docs/api/abi.rst
@@ -22,5 +22,11 @@ Model
     :members: defined_structures, functions, constructor, l1_handler, events
 
 .. autoclass:: starknet_py.abi.Abi.Function
+    :members:
+    :undoc-members:
+    :member-order: groupwise
 
 .. autoclass:: starknet_py.abi.Abi.Event
+    :members:
+    :undoc-members:
+    :member-order: groupwise
--- a/docs/api/account.rst
+++ b/docs/api/account.rst
@@ -1,9 +1,9 @@
 Account
 =======
 
----------
-Interface
----------
+---------------------
+BaseAccount interface
+---------------------
 
 .. py:module:: starknet_py.net.account.base_account
 
@@ -12,9 +12,9 @@ Interface
     :undoc-members:
     :member-order: groupwise
 
-----------------------
-Default implementation
-----------------------
+----------------------------------
+BaseAccount default implementation
+----------------------------------
 
 .. py:module:: starknet_py.net.account.account
 
--- a/docs/api/account.rst
+++ b/docs/api/account.rst
@@ -9,7 +9,8 @@ Interface
 
 .. autoclass:: BaseAccount
     :members:
-    :member-order: bysource
+    :undoc-members:
+    :member-order: groupwise
 
 ----------------------
 Default implementation
@@ -19,7 +20,8 @@ Default implementation
 
 .. autoclass-with-examples:: Account
     :members:
-    :member-order: bysource
+    :undoc-members:
+    :member-order: groupwise
 
 ------------------
 Account deployment
--- a/docs/api/contract.rst
+++ b/docs/api/contract.rst
@@ -4,27 +4,36 @@ Contract
 .. py:module:: starknet_py.contract
 
 .. autoclass:: Contract
-    :members: from_address, __init__, functions, from_address_sync, compute_address, compute_contract_hash
+    :members:
+    :undoc-members:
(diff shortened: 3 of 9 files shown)


drknzz (commented):
I feel like having all the members sorted `bysource` is much better.
Current approach shuffles properties of a class all over the place (where bysource keeps them at the top). Also I think it is just cleaner having all the `_sync` methods at the bottom, as they do not add any useful information to already documented async function.

war-in:
We discussed it internally and decided to leave `_sync` methods in the API docs. We will create a sphinx extension to move `properties` to the top (#716)

docs/api/account.rst
```diff
@@ -1,25 +1,27 @@
...
+---------------------
+BaseAccount interface
+---------------------
 
 .. py:module:: starknet_py.net.account.base_account
 
 .. autoclass:: BaseAccount
     :members:
-    :member-order: bysource
+    :undoc-members:
```
cptartur, line 12: Why was this added?

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -13,14 +13,12 @@
...
 from starknet_py.utils.contructor_args_translator import translate_constructor_args
 from starknet_py.utils.crypto.facade import pedersen_hash
-from starknet_py.utils.sync import add_sync_methods
 
 ContractDeployment = NamedTuple(
     "ContractDeployment", [("call", Call), ("address", int)]
 )
 
 
-@add_sync_methods
```
cptartur, line 23: Isn't this breaking?

war-in: Nope, because there isn't any `async` method in the Deployer 😅

docs/api/account.rst
war-in: To include methods (or properties) without a docstrings

docs/api/contract.rst
```diff
@@ -4,27 +4,36 @@ Contract
...
+    :members:
+    :undoc-members:
     :member-order: groupwise
 
 .. autoclass:: DeclareResult
     :exclude-members: __init__, __new__
-    :members: deploy, deploy_sync, wait_for_acceptance, wait_for_acceptance_sync, class_hash, compiled_contract, hash, status, block_number
+    :members:
+    :undoc-members:
     :member-order: groupwise
```
drknzz, line 39: Don't we want to set it in configuration so it does not have to be repeated every time (if we always use these parameters)?

docs/api/account.rst
cptartur: I'd prefer if we only included methods (and properties) with docstrings and removed `undoc-members` from the config.

docs/api/contract.rst
war-in: We will not use the `:undoc-members` because @cptartur suggested to create docstring for all API methods/properties

drknzz: Yes, but the `:member-order: groupwise` appears in (almost?) every autoclass.

war-in: I would like to keep it explicit

docs/api/signer.rst
```diff
@@ -1,16 +1,34 @@
...
+    :members:
+    :member-order: groupwise
+
+-------
+KeyPair
+-------
 
 .. autoclass:: KeyPair
+    :members:
+    :undoc-members:
```
drknzz, line 32: Was this omitted on purpose?

docs/api/abi.rst
```diff
@@ -22,5 +22,11 @@ Model
...
 
 .. autoclass:: starknet_py.abi.Abi.Function
+    :members:
+    :undoc-members:
+    :member-order: groupwise
 
 .. autoclass:: starknet_py.abi.Abi.Event
+    :members:
+    :undoc-members:
+    :member-order: groupwise
```
drknzz, line 32: Same here

docs/api/signer.rst
war-in: Yes, there is no need to add docstring for `private_key` and `public_key`, but would be nice to have them in the docs

docs/api/abi.rst
war-in: Same as above. Function and Event have undocumented fields

drknzz (approved):

PR #715, opened 2023-01-30
Update `Account creation` docs

### LINKED ISSUE #709: Update `Account creation` section
### Feature Request

`Account` has a `deploy_account` static method which is not present in the `Guide` (or anywhere in the docs). We should include it in the `Account creation` section.

war-in:
Closes #709 

## Introduced changes

- use the `Account.deploy_account` simplified flow

##

### DIFF
--- a/starknet_py/tests/e2e/docs/account_creation/test_deploy_prefunded_account.py
+++ b/starknet_py/tests/e2e/docs/account_creation/test_deploy_prefunded_account.py
@@ -43,26 +43,21 @@ async def test_deploy_prefunded_account(
     await res.wait_for_acceptance()
     # docs: start
 
-    # Create an Account instance
-    account = Account(
+    # Use `Account.deploy_account` static method to deploy an account
+    account_deployment_result = await Account.deploy_account(
         address=address,
-        client=GatewayClient(net=network),
-        key_pair=key_pair,
-        chain=StarknetChainId.TESTNET,
-    )
-
-    # Create and sign DeployAccount transaction
-    deploy_account_tx = await account.sign_deploy_account_transaction(
         class_hash=class_hash,
-        contract_address_salt=salt,
+        salt=salt,
+        key_pair=key_pair,
+        client=GatewayClient(net=network),
+        chain=StarknetChainId.TESTNET,
         constructor_calldata=[key_pair.public_key],
         max_fee=int(1e15),
     )
+    # Wait for deployment transaction to be accepted
+    await account_deployment_result.wait_for_acceptance()
 
-    resp = await account.client.deploy_account(transaction=deploy_account_tx)
-    await account.client.wait_for_tx(resp.transaction_hash)
-
-    # Since this moment account can be used to sign other transactions
+    # Since this moment account can be used as usual
+    account = account_deployment_result.account
     # docs: end
-
-    assert address == resp.address
+    assert account.address == address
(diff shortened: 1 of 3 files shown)


starknet_py/tests/e2e/docs/account_creation/test_deploy_prefunded_account.py
```diff
@@ -43,26 +43,21 @@ async def test_deploy_prefunded_account(
...
-    account = Account(
+    # Use `Account.deploy_account` static method to deploy an account
+    account_deployment_result = await Account.deploy_account(
         address=address,
-        client=GatewayClient(net=network),
+        class_hash=class_hash,
+        salt=salt,
         key_pair=key_pair,
+        client=GatewayClient(net=network),
         chain=StarknetChainId.TESTNET,
```
drknzz, line 53: Shouldn't we maybe use here (and possibly in other places in documentation) `chain_from_network()`? It's currently hardcoded to `StarknetChainId.TESTNET`, but the chain should be dependent on the network, as it's not specified above to be `testnet`.

war-in: I would rather change the network to `TESTNET`. Our examples are based on testnet

drknzz: But shouldn't the examples be as universal as possible? The network might as well be `TESTNET2`.

war-in: Right, but if someone sees how to use TESTNET it can use TESTNET2 as well. With the `network`, we would have to show what it really is

starknet_py/tests/e2e/docs/account_creation/test_deploy_prefunded_account.py
```diff
@@ -43,26 +49,30 @@ async def test_deploy_prefunded_account(
...
-        key_pair=key_pair,
-        chain=StarknetChainId.TESTNET,
-    )
+    # Define the client to be used to interact with StarkNet
+    client = GatewayClient(net=TESTNET)
+    chain = StarknetChainId.TESTNET
+    # docs: end
+
+    client = gateway_client
+    chain = chain_from_network(net=network, chain=StarknetChainId.TESTNET)
```
drknzz, line 58: I'm having second thoughts about using this function, as #719 is a thing now, and also I don't really like the way it works (and the fact that it's chain from network but it still takes chain). Maybe let's not include it in docs, my bad 😶

war-in: It is not in the docs ;)

starknet_py/tests/e2e/docs/account_creation/test_deploy_prefunded_account.py
```diff
@@ -43,26 +49,30 @@ async def test_deploy_prefunded_account(
...
         max_fee=int(1e15),
     )
+    # Wait for deployment transaction to be accepted
+    await account_deployment_result.wait_for_acceptance()
 
-    resp = await account.client.deploy_account(transaction=deploy_account_tx)
-    await account.client.wait_for_tx(resp.transaction_hash)
-
-    # Since this moment account can be used to sign other transactions
+    # Since this moment account can be used as usual
```
drknzz, line 75:
```suggestion
    # From now on, account can be used as usual
```
Since this moment sounds weird to me, but maybe it's just me 🤷

cptartur (approved):
Do we want to have a separate guide on how to use `sign_deploy_account_transaction` or is it not necessary?

war-in:
I think it is not. There is no reason to use it directly

drknzz:
> Do we want to have a separate guide on how to use `sign_deploy_account_transaction` or is it not necessary?

It's documented in Account API, and pretty much self explanatory from this example.

PR #723, opened 2023-01-30
Add docs about `client.call_contract`

drknzz:
## Introduced changes

- Add docs about `client.call_contract`

### DIFF
--- a/docs/guide/using_existing_contracts.rst
+++ b/docs/guide/using_existing_contracts.rst
@@ -1,6 +1,9 @@
 Using existing contracts
 ========================
 
+Existing contracts
+------------------
+
 Although it is possible to use :obj:`Client <starknet_py.net.gateway_client.GatewayClient>` to interact with contracts, it requires translating python values into Cairo
 values. Contract offers that and some other utilities.
 
@@ -17,6 +20,19 @@ This is how we can interact with it:
 .. codesnippet:: ../../starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
     :language: python
     :dedent: 4
+
+
+Raw contract calls
+------------------
+
+If you do not have ABI statically, but you know the interface of the contract on StarkNet, you can make a raw call:
+
+.. codesnippet:: ../../starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
+    :language: python
+    :dedent: 4
+    :start-after: docs-raw-call: start
+    :end-before: docs-raw-call: end
+
 
 Fees
 ----
--- a/starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
+++ b/starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
@@ -92,6 +92,7 @@ async def test_using_existing_contracts(account, erc20_contract):
 
 @pytest.mark.asyncio
 async def test_raw_call(account):
+    # pylint: disable=import-outside-toplevel
     # docs-raw-call: start
     from starkware.starknet.public.abi import get_selector_from_name
 
--- a/starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
+++ b/starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
@@ -98,11 +98,12 @@ async def test_raw_call(account):
 
     from starknet_py.net.client_models import Call
 
-    eth_token_address = (
-        0x049D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7
-    )
     # docs-raw-call: end
-    eth_token_address = int(FEE_CONTRACT_ADDRESS, 0)
+    # fmt: off
+    # docs-raw-call: start
+    eth_token_address = 0x049D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7
+    # docs-raw-call: end
+    # fmt: on
     # docs-raw-call: start
 
     # Create a call to function "balanceOf" at address `eth_token_address`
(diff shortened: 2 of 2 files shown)


starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```diff
@@ -86,3 +88,26 @@ async def test_using_existing_contracts(account, erc20_contract):
     # docs: end
 
     assert balance == 200
+
+    # docs-raw-call: start
+    from starkware.starknet.public.abi import get_selector_from_name
+
```
war-in, line 94: That newline is probably innecessary

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```diff
@@ -86,3 +88,26 @@ async def test_using_existing_contracts(account, erc20_contract):
...
+    from starknet_py.net.client_models import Call
+
+    eth_token_address = (
+        0x049D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7
+    )
+    # docs-raw-call: end
+    eth_token_address = int(FEE_CONTRACT_ADDRESS, 0)
+    # docs-raw-call: start
```
war-in, line 102: Why do we have to do this? Isn't the first line enough?

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```diff
@@ -86,3 +88,26 @@ async def test_using_existing_contracts(account, erc20_contract):
     # docs: end
 
     assert balance == 200
+
+    # docs-raw-call: start
+    from starkware.starknet.public.abi import get_selector_from_name
```
war-in, line 93: Let's maybe create another test case for that. `test_using_call_contract`

docs/guide/using_existing_contracts.rst
```diff
@@ -1,6 +1,9 @@
 Using existing contracts
 ========================
 
+Existing contracts
+------------------
+
```
war-in, line 6: I'm not sure if it is necessary. Do you have a strong opinion on that?

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
drknzz: We don't, I thought it could be a little change-proof this way, as the `FEE_CONTRACT_ADDRESS` constant could change, and I wanted to show some address in the docs. Should we somehow change it?

docs/guide/using_existing_contracts.rst
drknzz: It's currently not visible on the right side.
image

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
war-in: I would leave that `0x049D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7` address only. If it changes we will know about it.

docs/guide/using_existing_contracts.rst
war-in: Right, but the title is about that section.

drknzz: Yes, but it's nice to have every topic on the right side.
+ it's already like this in serialization section for example
image

cptartur: I think it's okay to add a subsection here if you think it increases readability.

docs/guide/using_existing_contracts.rst
```diff
@@ -18,6 +21,19 @@ This is how we can interact with it:
     :language: python
     :dedent: 4
 
+
+Raw contract calls
+------------------
+
+If you do not have ABI statically, but you know the interface of the contract on StarkNet, you can make a raw call:
```
cptartur, line 28: Let's extend this description to indicate that it's impossible to use `Contract` api without an abi (contract functions won't generate), and that's why client has to be used directly.

cptartur (changes requested):

docs/guide/using_existing_contracts.rst
war-in: Fine, I will not complain about it, since it makes the docs more readable 👍

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```diff
@@ -86,3 +86,34 @@ async def test_using_existing_contracts(account, erc20_contract):
...
+
+    from starknet_py.net.client_models import Call
+
+    # docs-raw-call: end
+    # fmt: off
+    # docs-raw-call: start
+    eth_token_address = 0x049D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7
+    # docs-raw-call: end
+    # fmt: on
```
drknzz, line 104: The formats comments are so the address is in single line instead of 3

starknet_py/tests/e2e/docs/guide/test_using_existing_contracts.py
```diff
@@ -112,7 +110,10 @@ async def test_raw_call(account):
         selector=get_selector_from_name("balanceOf"),
         calldata=[account.address],
     )
+    # Pass the created call to Client.call_contract
     account_balance = await account.client.call_contract(call)
+
+    # Note that a Contract instance cannot be used here, since it needs ABI to generate the functions
```
cptartur, line 116: I think it'd be better to include a proper explanation in the docs themselves instead of comment in the snippet.

drknzz: I think it's kind of obvious, since there exists a subsection with the raw calling. Here it is close to the context.

cptartur: Okay, we can leave it as it is.

cptartur (approved):

war-in (approved):

PR #733, opened 2023-02-01
Fix `Client.get_state_update`

### LINKED ISSUE #731: [BUG] `FullNodeClient.get_state_update` doesn't match RPC specs
### What happened

I've tried to use `FullNodeClient.get_state_update` when there was storage diff.
Marshmallow threw a validation error.

### Stack trace

E           marshmallow.exceptions.ValidationError: {'state_diff': {'storage_diffs': {0: {'key': ['Missing data for required field.'], 'value': ['Missing data for required field.'], 'storage_entries': ['Unknown field.']}, 1: {'key': ['Missing data for required field.'], 'value': ['Missing data for required field.'], 'storage_entries': ['Unknown field.']}}}}

### Steps to reproduce

1. Use devnet and perform some action to change the storage.
2. Use `FullNodeClient.get_state_update` to get the state.
3. Get en error.

### SDK Version

0.13.0

### Python version

3.9.13

Mac

war-in:
Closes #731 

## Introduced changes

- Client.get_state_update works how it should
- Dataclasses are compatible with RPC API
- Tests are added to avoid such a situation next time

- Tests for the `nonces` field can't be added because `devnet` doesn't include them in the block

##

- Dataclasses for BlockStateUpdate are different (compatible with RPC API)

### DIFF
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -273,7 +273,7 @@ class StorageEntry:
 
 
 @dataclass
-class StorageDiff:
+class StorageDiffItem:
     address: int
     storage_entries: List[StorageEntry]
 
@@ -301,7 +301,7 @@ class ContractsNonce:
 class StateDiff:
     deployed_contracts: List[DeployedContract]
     declared_contract_hashes: List[int]
-    storage_diffs: List[StorageDiff]
+    storage_diffs: List[StorageDiffItem]
     nonces: List[ContractsNonce]
 
 
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -266,6 +266,7 @@ class BlockTransactionTraces:
     traces: List[BlockSingleTransactionTrace]
 
 
+@dataclass
 class StorageEntry:
     key: int
     value: int
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -266,11 +266,15 @@ class BlockTransactionTraces:
     traces: List[BlockSingleTransactionTrace]
 
 
+class StorageEntry:
+    key: int
+    value: int
+
+
 @dataclass
 class StorageDiff:
     address: int
-    key: int
-    value: int
+    storage_entries: List[StorageEntry]
 
 
 @dataclass
@@ -287,6 +291,20 @@ class DeployedContract:
 
 
 @dataclass
+class UpdatedNonce:
+    contract_address: int
+    nonce: int
+
+
+@dataclass
+class StateDiff:
+    deployed_contracts: List[DeployedContract]
+    declared_contract_hashes: List[int]
+    storage_diffs: List[StorageDiff]
+    nonces: List[UpdatedNonce]
+
+
+@dataclass
 class BlockStateUpdate:
     """
     Dataclass representing a change in state of a block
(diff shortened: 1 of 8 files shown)


starknet_py/net/client_models.py
```diff
@@ -267,12 +267,17 @@ class BlockTransactionTraces:
...
 @dataclass
-class StorageDiff:
-    address: int
+class StorageEntry:
     key: int
     value: int
 
 
+@dataclass
+class StorageDiff:
```
cptartur, line 276: Perhaps naming this `StorageDiffItem` like in the rpc spec or a would make more sense? Since it's not a whole storage diff but a part of one.

starknet_py/net/schemas/gateway.py
```diff
@@ -275,61 +278,64 @@ def make_dataclass(self, data, **kwargs):
...
+        Felt(),
+        data_key="declared_contracts",
+        required=True,
+    )
+    storage_diffs = fields.Dict(
+        keys=fields.String(),
+        values=fields.Raw(),
+        data_key="storage_diffs",
+        required=True,
+    )
```
cptartur, line 306: Would it be possible to define a proper schema for this?

starknet_py/net/schemas/gateway.py
```diff
@@ -275,61 +278,64 @@ def make_dataclass(self, data, **kwargs):
...
-
-        del data["state_diff"]
-
+        storage_diffs: Dict = data["state_diff"].storage_diffs
+        fixed_storage_diffs: List[StorageDiff] = []
+        for address in storage_diffs.keys():
+            entries = []
+            for entry in storage_diffs[address]:
+                entries.append(StorageEntrySchema().load(entry))
+            fixed_storage_diffs.append(StorageDiff(address, entries))
```
cptartur, line 328: I know it's been like this before this PR, but I'd consider moving this to a separate method so it's a bit more readable.

starknet_py/net/schemas/gateway.py
```diff
@@ -275,61 +278,64 @@ def make_dataclass(self, data, **kwargs):
...
+
+        nonces: Dict = data["state_diff"].nonces
+        fixed_nonces: List[ContractsNonce] = []
+        for address, nonce in nonces.items():
+            fixed_nonces.append(ContractsNonce(address, nonce))
```
cptartur, line 333: Same with this

starknet_py/net/schemas/rpc.py
```diff
@@ -261,24 +287,10 @@ class BlockStateUpdateSchema(Schema):
     old_root = Felt(data_key="old_root", required=True)
     state_diff = fields.Nested(StateDiffSchema(), data_key="state_diff", required=True)
 
-    @pre_load
-    def preprocess(self, data, **kwargs):
-        # Remove this when support for nonces is added
```
cptartur, line 266: I guess it's now 😅

starknet_py/net/schemas/gateway.py
war-in: Sure, makes a lot of sense!

cptartur (approved):
Looks good codewise. You can consider moving `StorageEntrySchema` to common since its used by both rpc and gateway.

PR #734, opened 2023-02-01
Auto assign author to PR

war-in:
Closes #

## Introduced changes

Assigning authors to PRs is tough and sometimes we forget about it. That is why I created the workflow :)

##

### DIFF
--- a/.github/workflows/auto-author-assign.yml
+++ b/.github/workflows/auto-author-assign.yml
@@ -2,8 +2,7 @@
 name: Auto Author Assign
 
 on:
-  pull_request_target:
-    types: [ opened, reopened ]
+  pull_request:
 
 permissions:
   pull-requests: write
--- a/.github/workflows/auto-author-assign.yml
+++ b/.github/workflows/auto-author-assign.yml
@@ -3,6 +3,7 @@ name: Auto Author Assign
 
 on:
   pull_request:
+    types: [ opened, reopened ]
 
 permissions:
   pull-requests: write
--- a/.github/workflows/auto-author-assign.yml
+++ b/.github/workflows/auto-author-assign.yml
@@ -13,4 +13,4 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.repository == 'software-mansion/starknet.py' }}  # do not run on forks
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-author-assign.yml
```diff
@@ -0,0 +1,15 @@
...
+
+permissions:
+  pull-requests: write
+
+jobs:
+  assign-author:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'software-mansion/starknet.py' }}  # do not run on forks
+    steps:
+      - uses: toshimaru/auto-author-assign@v1.6.2
```
cptartur, line 15: Looks fine I think, what I'd change is fixing `@version` to a specific version of the commit for extra security, [see here how it's done](https://github.com/software-mansion/starknet.py/blob/development/.github/workflows/dependabot.yml#L26)

.github/workflows/auto-author-assign.yml
```diff
@@ -0,0 +1,15 @@
+# .github/workflows/auto-author-assign.yml
+name: Auto Author Assign
+
+on:
+  pull_request:
```
THenry14, line 5: Correct me if I'm wrong, but wouldn't it trigger this workflow at many events? I guess we could run it only for `opened` and maybe `reopened`

.github/workflows/auto-author-assign.yml
THenry14: We should do it for every action, at the very least not official ones

.github/workflows/auto-author-assign.yml
```diff
@@ -0,0 +1,16 @@
...
+
+permissions:
+  pull-requests: write
+
+jobs:
+  assign-author:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'software-mansion/starknet.py' }}  # do not run on forks
+    steps:
+      - uses: toshimaru/auto-author-assign/push@2daaeb2988aef24bf37e636fe733f365c046aba0  # v1.6.2
```
THenry14, line 16:
```suggestion
      - uses: toshimaru/auto-author-assign@2daaeb2988aef24bf37e636fe733f365c046aba0  # v1.6.2
```

war-in: Right, found it too 😅

THenry14 (approved):

PR #736, opened 2023-02-02
Optional StarknetChainId

### LINKED ISSUE #719: StarknetChainId not Optional
### Feature Request

Currently, chain parameter in `Account.deploy_account` is required. It seems though, it could be optional, as the `chain` could be taken from required `client` parameter.

drknzz:
Closes #719

### DIFF
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -91,6 +91,18 @@ class Account(BaseAccount):
 
         self.signer = signer or self._create_signer(key_pair, chain)
 
+    @staticmethod
+    def _assert_no_key_pair_chain(
+        key_pair: Optional[KeyPair], chain: Optional[StarknetChainId]
+    ):
+        if chain is not None:
+            warnings.warn(
+                "Arguments signer and chain are mutually exclusive. "
+                "This is going to be an Error in the future version."
+            )
+        if key_pair is not None:
+            raise ValueError("Arguments signer and key_pair are mutually exclusive.")
+
     def _create_signer(
         self, key_pair: Optional[KeyPair], chain: Optional[StarknetChainId]
     ) -> StarkCurveSigner:
@@ -105,18 +117,6 @@ class Account(BaseAccount):
         return StarkCurveSigner(
             account_address=self.address, key_pair=key_pair, chain_id=chain
         )
-
-    @staticmethod
-    def _assert_no_key_pair_chain(
-        key_pair: Optional[KeyPair], chain: Optional[StarknetChainId]
-    ):
-        if chain is not None:
-            warnings.warn(
-                "Arguments signer and chain are mutually exclusive. "
-                "This is going to be an Error in the future version."
-            )
-        if key_pair is not None:
-            raise ValueError("Arguments signer and key_pair are mutually exclusive.")
 
     @property
     def address(self) -> int:
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -73,6 +73,7 @@ class Account(BaseAccount):
         Creates an Account instance.
         If only `signer` is given, it will be used in Account.
         If only `key_pair` is given, `client.net` will be used to create `chain`.
+
         :param address: Address of the account contract.
         :param client: Instance of Client which will be used to add transactions.
         :param signer: Custom signer to be used by Account.
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -1,5 +1,6 @@
 import dataclasses
 import re
+import warnings
 from collections import OrderedDict
 from typing import Dict, Iterable, List, Optional, Tuple, TypeVar, Union
 
@@ -86,6 +87,8 @@ class Account(BaseAccount):
         self._client = client
 
         if signer is not None:
+            if chain is not None:
+                warnings.warn("Arguments signer and chain are mutually exclusive.")
             if key_pair is not None:
                 raise ValueError(
                     "Arguments signer and key_pair are mutually exclusive."
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -86,8 +86,6 @@ class Account(BaseAccount):
         self._client = client
 
         if signer is not None:
-            if chain is not None:
-                raise ValueError("Arguments signer and chain are mutually exclusive.")
             if key_pair is not None:
                 raise ValueError(
                     "Arguments signer and key_pair are mutually exclusive."
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -25,7 +25,6 @@ from starknet_py.net.client_models import (
 from starknet_py.net.models import (
     AddressRepresentation,
     StarknetChainId,
-    chain_from_network,
     compute_address,
     parse_address,
 )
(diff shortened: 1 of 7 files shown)


starknet_py/net/account/account.py
```diff
@@ -71,6 +70,9 @@ def __init__(
         chain: Optional[StarknetChainId] = None,
     ):
         """
+        Creates an Account instance.
+        If only `signer` is given, it will be used in Account.
+        If only `key_pair` is given, `client.net` will be used to create `chain`.
```
war-in, line 75: That is quite strange. Although I know what is the purpose of that PR I'm not sure if I understand that 😅

starknet_py/net/models/chains.py
```diff
@@ -10,10 +13,33 @@ class StarknetChainId(Enum):
...
+        :return: StarknetChainId instance.
+        """
+        net_to_chain = {
+            MAINNET: StarknetChainId.MAINNET,
+            TESTNET: StarknetChainId.TESTNET,
+            TESTNET2: StarknetChainId.TESTNET2,
+        }
+
+        if not isinstance(net, str) or net not in net_to_chain:
+            raise ValueError("Unknown Network.")
```
war-in, line 31: I would change the error message to be more descriptive. When we use it in the Account init and the net is devnet it will raise an error `Unknown Network.` but the user doesn't specify the network parameter in the Account init

starknet_py/tests/e2e/docs/code_examples/test_account.py
```diff
@@ -17,16 +17,24 @@
...
+        chain_id=StarknetChainId.TESTNET,
+    )
     account = Account(
         address=0x123,
         client=GatewayClient(net=TESTNET),
-        signer=StarkCurveSigner(
-            account_address=0x123,
-            key_pair=KeyPair(12, 34),
-            chain_id=StarknetChainId.TESTNET,
-        ),
```
war-in, line 27: I would not extract it from here

starknet_py/net/account/account.py
```diff
@@ -79,26 +83,31 @@ def __init__(
...
         self._client = client
 
-        if signer is not None and key_pair is not None:
-            raise ValueError("Arguments signer and key_pair are mutually exclusive.")
-
-        if signer is None:
-            if key_pair is None:
+        if signer is not None:
+            if chain is not None:
+                warnings.warn("Arguments signer and chain are mutually exclusive.")
```
cptartur, line 91: Let's reword the error message so it clearly indicates that this combination will raise an error in future versions.

starknet_py/net/account/account.py
```diff
@@ -79,26 +83,31 @@ def __init__(
...
                 raise ValueError(
-                    "Either a signer or a key_pair must be provided in Account constructor."
+                    "Arguments signer and key_pair are mutually exclusive."
                 )
 
-            chain = chain_from_network(net=client.net, chain=chain)
-            signer = StarkCurveSigner(
-                account_address=self.address, key_pair=key_pair, chain_id=chain
+            self.signer = signer
+            return
```
cptartur, line 98: I fear that relying on return here might cause some problems in the future. What if someone adds additional code below these changes without realizing that there's an early return here.

starknet_py/net/account/account.py
```diff
@@ -79,26 +83,31 @@ def __init__(
...
+                "Either a signer or a key_pair must be provided in Account constructor."
             )
-        self.signer: BaseSigner = signer
+
+        if chain is None:
+            chain = StarknetChainId.from_network(net=client.net)
+
+        self.signer = StarkCurveSigner(
+            account_address=self.address, key_pair=key_pair, chain_id=chain
+        )
```
cptartur, line 110: This logic has gotten quite complex now. I think it's a good time to separate parts of this into helper functions so its easier to follow.

starknet_py/net/account/account_test.py
```diff
@@ -62,13 +62,14 @@ def test_create_account_from_signer():
     assert account.signer == signer
 
 
-def test_create_account_raises_on_no_chain_and_signer():
```
cptartur, line 65: So i get that chain can be somehow constructed from the `net` even if signer is not provided, but what if non-standard net is used and chain creation fails? We should test for a case like this.

starknet_py/net/models/chains.py
```diff
@@ -10,10 +13,33 @@ class StarknetChainId(Enum):
     TESTNET = int_from_bytes(b"SN_GOERLI")
     TESTNET2 = int_from_bytes(b"SN_GOERLI2")
 
+    @classmethod
+    def from_network(cls, net: Network) -> StarknetChainId:
```
cptartur, line 17: `cls` is not used anywhere, this could be changed to a staticmethod

cptartur (changes requested):
Acc

cptartur:
I'd hold merging this until we discuss `net` parameter in `Client` interface.

starknet_py/net/account/account.py
```diff
@@ -79,26 +83,40 @@ def __init__(
...
-            signer = StarkCurveSigner(
-                account_address=self.address, key_pair=key_pair, chain_id=chain
+    @staticmethod
+    def _assert_no_key_pair_chain(
+        key_pair: Optional[KeyPair], chain: Optional[StarknetChainId]
+    ):
+        if chain is not None:
+            warnings.warn(
+                "Arguments signer and chain are mutually exclusive. "
+                "This is going to be an Error in the future version."
```
cptartur, line 101:
```suggestion
                "This will raise an Error in the future version."
```

starknet_py/net/account/account.py
```diff
@@ -79,26 +83,40 @@ def __init__(
...
-                raise ValueError(
-                    "Either a signer or a key_pair must be provided in Account constructor."
-                )
+        self.signer = signer or self._create_signer(key_pair, chain)
 
-            chain = chain_from_network(net=client.net, chain=chain)
-            signer = StarkCurveSigner(
-                account_address=self.address, key_pair=key_pair, chain_id=chain
+    @staticmethod
+    def _assert_no_key_pair_chain(
```
cptartur, line 95:
```suggestion
    def _assert_no_key_pair_and_chain(
```

starknet_py/net/account/account.py
```diff
@@ -79,26 +83,40 @@ def __init__(
...
-            raise ValueError("Arguments signer and key_pair are mutually exclusive.")
+        if signer is not None:
+            self._assert_no_key_pair_chain(key_pair, chain)
 
-        if signer is None:
-            if key_pair is None:
-                raise ValueError(
-                    "Either a signer or a key_pair must be provided in Account constructor."
-                )
+        self.signer = signer or self._create_signer(key_pair, chain)
```
cptartur, line 92: What if we rewrote this to something like

```py
if signer is not None:
  self._assert_no_key_pair_chain(key_pair, chain)
  self.signer = signer
else:
  self.signer = self._create_signer(key_pair, chain)
```
I think it makes it easier to follow what happens in cases signer is and isn't provided.

starknet_py/net/models/chains.py
```diff
@@ -10,10 +13,35 @@ class StarknetChainId(Enum):
...
+        """
+        net_to_chain = {
+            MAINNET: cls.MAINNET,
+            TESTNET: cls.TESTNET,
+            TESTNET2: cls.TESTNET2,
+        }
+
+        if not isinstance(net, str) or net not in net_to_chain:
+            raise ValueError(
+                f"Cannot create StarknetChainId from unknown Network '{net}'."
```
cptartur, line 32:
```suggestion
                f"Cannot create StarknetChainId from unknown net '{net}'."
```

I think it's more descriptive if we use an argument name instead of a type.

PR #742, opened 2023-02-06
Show using custom nonce logic

### LINKED ISSUE #739: Document using custom nonces
### Feature Request

Document replacing `get_nonce` function implementation with custom one.

cptartur:
Closes #739 

## Introduced changes

This PR adds a example in guide showing using custom nonce logic in `Account`.

##

### DIFF
--- a/docs/guide/account_and_client.rst
+++ b/docs/guide/account_and_client.rst
@@ -14,6 +14,18 @@ Here is an example:
     :dedent: 4
 
 
+Custom nonce logic
+------------------
+
+By default, :ref:`Account` calls StarkNet for nonce every time a new transaction is signed or executed.
+This is okay in for most users, but in case your applications needs to pre-sing multiple transactions
+for execution, deals with high amount of transactions or for whatever reason needs to support different nonce
+logic, it is possible to so with :ref:`Account. Simply overwrite the
+:meth:`~starknet_py.net.account.account.Account.get_nonce` with your own logic.
+
+.. codesnippet:: ../../starknet_py/tests/e2e/docs/guide/test_custom_nonce.py
+    :language: python
+    :dedent: 4
 
 FullNodeClient usage
 --------------------
--- a/docs/guide/account_and_client.rst
+++ b/docs/guide/account_and_client.rst
@@ -18,10 +18,10 @@ Custom nonce logic
 ------------------
 
 By default, :ref:`Account` calls StarkNet for nonce every time a new transaction is signed or executed.
-This is okay in for most users, but in case your applications needs to pre-sing multiple transactions
-for execution, deals with high amount of transactions or for whatever reason needs to support different nonce
-logic, it is possible to so with :ref:`Account. Simply overwrite the
-:meth:`~starknet_py.net.account.account.Account.get_nonce` with your own logic.
+This is okay for most users, but in case your applications needs to pre-sign multiple transactions
+for execution, deals with high amount of transactions or just needs to support different nonce
+logic, it is possible to so with :ref:`Account`. Simply overwrite the
+:meth:`~starknet_py.net.account.account.Account.get_nonce` method with your own logic.
 
 .. codesnippet:: ../../starknet_py/tests/e2e/docs/guide/test_custom_nonce.py
     :language: python
--- a/docs/guide/account_and_client.rst
+++ b/docs/guide/account_and_client.rst
@@ -14,19 +14,6 @@ Here is an example:
     :dedent: 4
 
 
-Custom nonce logic
-------------------
-
-By default, :ref:`Account` calls StarkNet for nonce every time a new transaction is signed or executed.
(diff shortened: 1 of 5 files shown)


docs/guide/account_and_client.rst
```diff
@@ -14,6 +14,18 @@ Here is an example:
     :dedent: 4
 
 
+Custom nonce logic
+------------------
+
+By default, :ref:`Account` calls StarkNet for nonce every time a new transaction is signed or executed.
+This is okay in for most users, but in case your applications needs to pre-sing multiple transactions
+for execution, deals with high amount of transactions or for whatever reason needs to support different nonce
```
drknzz, line 22:
```suggestion
for execution, deals with high amount of transactions or just needs to support different nonce
```

docs/guide/account_and_client.rst
```diff
@@ -14,6 +14,18 @@ Here is an example:
...
+logic, it is possible to so with :ref:`Account. Simply overwrite the
```
drknzz, line 23:
```suggestion
logic, it is possible to so with :ref:`Account`. Simply overwrite the
```

docs/guide/account_and_client.rst
```diff
@@ -14,6 +14,18 @@ Here is an example:
...
+:meth:`~starknet_py.net.account.account.Account.get_nonce` with your own logic.
```
drknzz, line 24:
```suggestion
:meth:`~starknet_py.net.account.account.Account.get_nonce` method with your own logic.
```

starknet_py/tests/e2e/docs/guide/test_custom_nonce.py
```diff
@@ -0,0 +1,58 @@
...
+        address=address,
+        client=client,
+        key_pair=KeyPair.from_private_key(private_key),
+        chain=StarknetChainId.TESTNET,
+    )
+    # docs: end
+
+    assert account.nonce_counter == 0
+    await account.sign_invoke_transaction(calls=Call(0x1, 0x1, []), max_fee=10000000000)
+    assert account.nonce_counter == 1
```
drknzz, line 58: Maybe this could also be included?

docs/guide/account_and_client.rst
```diff
@@ -14,6 +14,18 @@ Here is an example:
...
+
+By default, :ref:`Account` calls StarkNet for nonce every time a new transaction is signed or executed.
+This is okay for most users, but in case your applications needs to pre-sign multiple transactions
+for execution, deals with high amount of transactions or just needs to support different nonce
+logic, it is possible to so with :ref:`Account`. Simply overwrite the
+:meth:`~starknet_py.net.account.account.Account.get_nonce` method with your own logic.
+
+.. codesnippet:: ../../starknet_py/tests/e2e/docs/guide/test_custom_nonce.py
+    :language: python
+    :dedent: 4
```
drknzz, line 28: Maybe we should put it at the end of this section? I feel like it's not that common to use use it, and it is the second subsection in the whole guide.
Or we could put it somewhere in `signing` section maybe?

starknet_py/tests/e2e/docs/guide/test_custom_nonce.py
cptartur: Don't think it's important to have that in guide. What this example is supposed to show is that it is possible to have custom nonces, not how to do it exactly.

docs/guide/account_and_client.rst
cptartur: We can move it to the end.

starknet_py/tests/e2e/docs/guide/test_custom_nonce.py
drknzz: Right, but it shows how the custom class works in practice. But I'm fine with not having it though.

cptartur: Yeah, I don't think it matters how it works. Imo, the guide could even be showing `...` in place of the custom implementation.

docs/guide/account_and_client.rst
```diff
@@ -39,3 +38,17 @@ You can use :class:`starknet_py.net.client_errors.ClientError` to catch errors f
...
+
+Custom nonce logic
+------------------
+
+By default, :ref:`Account` calls StarkNet for nonce every time a new transaction is signed or executed.
+This is okay for most users, but in case your applications needs to pre-sign multiple transactions
+for execution, deals with high amount of transactions or just needs to support different nonce
+logic, it is possible to so with :ref:`Account`. Simply overwrite the
+:meth:`~starknet_py.net.account.account.Account.get_nonce` method with your own logic.
+
```
drknzz, line 51: Should we write here that the code below is just an example implementation of a nonce counter?

drknzz (approved):

PR #746, opened 2023-02-09
Update starknet-devnet to 0.4.5

cptartur:
Closes #

## Introduced changes

- Updated starknet-devnet to 0.4.5 (bumped locked version)
- Fixed RPC spec compliance issues in FullNodeClient

##


starknet_py/net/full_node_client.py
```diff
@@ -474,11 +473,32 @@ def _create_broadcasted_txn_common_properties(transaction: AccountTransaction) -
...
-        "version": hash_to_felt(transaction.version),
-        "signature": [hash_to_felt(sig) for sig in transaction.signature],
-        "nonce": hash_to_felt(transaction.nonce)
+        "max_fee": _to_rpc_felt(transaction.max_fee),
+        "version": _to_rpc_felt(transaction.version),
+        "signature": [_to_rpc_felt(sig) for sig in transaction.signature],
+        "nonce": _to_rpc_felt(transaction.nonce)
         if transaction.nonce is not None
-        else None,
+        else "0x00",
```
cptartur, line 481: Specification requires the nonce to always be there (which doesn't make sens in tx version 0). I had to introduce a default value in place of `None`. 

This is not ideal, but tx version 0 will be gone soon anyways.

cptartur:
This needs some tests I think. Especially the `_to_storage_key` function.

starknet_py/net/full_node_client.py
```diff
@@ -474,11 +473,32 @@ def _create_broadcasted_txn_common_properties(transaction: AccountTransaction) -
...
     }
     return broadcasted_txn_common_properties
+
+
+def _to_storage_key(key: int) -> str:
+    hashed_key = hex(key).lstrip("0x")
+
+    if hashed_key[0] not in ("0", "1", "2", "3", "4", "5", "6", "7"):
```
war-in, line 489:
```suggestion
    if hashed_key[0] not in str(range(8)):
```
?

starknet_py/net/full_node_client.py
```diff
@@ -474,11 +473,32 @@ def _create_broadcasted_txn_common_properties(transaction: AccountTransaction) -
...
+        hashed_key = "0" + hashed_key
+
+    if not 0 < len(hashed_key) <= 63:
+        raise ValueError(f"Value {key} cannot be represented as RPC storage key.")
```
war-in, line 493: Let's say why it can't be represented (add a comment about the length)

starknet_py/net/full_node_client.py
```diff
@@ -474,11 +473,32 @@ def _create_broadcasted_txn_common_properties(transaction: AccountTransaction) -
...
+        "signature": [_to_rpc_felt(sig) for sig in transaction.signature],
+        "nonce": _to_rpc_felt(transaction.nonce)
         if transaction.nonce is not None
-        else None,
+        else "0x00",
     }
     return broadcasted_txn_common_properties
+
+
+def _to_storage_key(key: int) -> str:
```
war-in, line 486: Please add a docstring with current `storage_key` format

starknet_py/net/full_node_client.py
```diff
@@ -474,11 +473,32 @@ def _create_broadcasted_txn_common_properties(transaction: AccountTransaction) -
...
+    if hashed_key[0] not in ("0", "1", "2", "3", "4", "5", "6", "7"):
+        hashed_key = "0" + hashed_key
+
+    if not 0 < len(hashed_key) <= 63:
+        raise ValueError(f"Value {key} cannot be represented as RPC storage key.")
+
+    return "0x0" + hashed_key
+
+
+def _to_rpc_felt(value: Hash):
```
war-in, line 498: Docstring with `rpc_felt` format would be useful

starknet_py/net/full_node_client.py
```diff
@@ -474,11 +473,32 @@ def _create_broadcasted_txn_common_properties(transaction: AccountTransaction) -
...
+def _to_storage_key(key: int) -> str:
+    hashed_key = hex(key).lstrip("0x")
+
+    if hashed_key[0] not in ("0", "1", "2", "3", "4", "5", "6", "7"):
+        hashed_key = "0" + hashed_key
+
+    if not 0 < len(hashed_key) <= 63:
+        raise ValueError(f"Value {key} cannot be represented as RPC storage key.")
+
+    return "0x0" + hashed_key
```
drknzz, line 495: I would validate the result against the regex before returning.

starknet_py/net/full_node_client.py
drknzz, line 489: Wouldn't this throw on key = 0? You re accessing index 0 of hashed_key, but it can be an empty string.

cptartur: That won't work, it'd evaluate to `Range(0, 8)` or something like this. We could do `[str(i) for i in range(8)]` if you think that's better?

drknzz: Maybe change it to just `0 <= int(x) <= 7`

drknzz: Oh, it could be `a - f` as well

starknet_py/net/full_node_client.py
cptartur: I'll be changing the check to verify agains the RPC regex as per @drknzz suggestion, so that kind of error won't be possible.

starknet_py/net/full_node_client.py
cptartur: It shouldn't now

cptartur: > Maybe change it to just `0 <= int(x) <= 7`

That'd have to be `0 <= int(hashed_key[0], 16) <= 7`. Is that more readable?

drknzz: To be fair, the original code is more readable 🙈

cptartur: Let's keep it then.

drknzz: Hmm, what if we take `key = 222`?
```
hex(222) = 0xde
```
Current logic would say that it is invalid key_storage. But isn't it equivalent to `0x00de`, which complies with the spec?

starknet_py/net/client_test.py
```diff
@@ -24,3 +25,20 @@ def test_handle_rpc_error_server_error():
 
     with pytest.raises(ServerError, match="RPC request failed."):
         RpcHttpClient.handle_rpc_error(no_error_dict)
+
+
+@pytest.mark.parametrize(
+    "key,expected",
```
war-in, line 31:
```suggestion
    "key, expected",
```

war-in (approved):

starknet_py/net/client_test.py
```diff
@@ -24,3 +25,26 @@ def test_handle_rpc_error_server_error():
...
+    assert _to_storage_key(key) == expected
+
+
+@pytest.mark.parametrize(
+    "key",
+    [int(1e100), -1, 0x8FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF],
+)
+def test_get_rpc_storage_key_raises_on_non_representable_key(key):
+    with pytest.raises(ValueError, match="cannot be represented"):
+        _to_storage_key(key)
```
drknzz, line 50: Maybe one more test case for `ADDR_BOUND - 1` would be handy? It's used to calculate the storage_var address.

drknzz (approved):

PR #747, opened 2023-02-10
Autoclass docstring deep inheritance

### LINKED ISSUE #705: Update `autoclass-with-examples` to deeply inherit docstrings
### Feature Request

Currently `autoclass-with-examples` takes docstring only from the parent class. Would be perfect to inherit it from the parents of parent if the parent's docstring is empty.

drknzz:
Closes #705


war-in (commented):
Did you test it in any way?

drknzz:
> Did you test it in any way?

Yes, locally.

war-in:
Is it possble to include a test here?

docs/_ext/autoclass_with_examples.py
```diff
@@ -100,14 +96,18 @@ def _create_hint(file_name: str, method_name: str) -> str:
...
     """
-    If method does not have the __doc__, takes it from the base method.
+    If method does not have the __doc__, takes it from the ancestor method.
     """
-    if method.__doc__ is None and method_name in base_class.__dict__:
-        method.__doc__ = getattr(base_class, method_name).__doc__ + hint
-    else:
-        method.__doc__ = (method.__doc__ or "") + hint
+    parent_method = method
+    while parent_method and parent_method.__doc__ is None:
```
cptartur, line 104:
```suggestion
    while parent_method is not None and parent_method.__doc__ is None:
```

docs/_ext/autoclass_with_examples.py
```diff
@@ -100,14 +96,18 @@ def _create_hint(file_name: str, method_name: str) -> str:
...
+        parent_method = getattr(parent_class, method_name, None)
+        parent_class = parent_class.__base__
```
cptartur, line 106: I think this method taking `parent_class` as argument and renaming `parent_method = method` makes this more confusing that it needs to be. How about this function accepted a method name and related class and did the resolving internally? 

I don't see a good reason why this has to take in `base_class` as an argument.

cptartur (changes requested):

docs/_ext/autoclass_with_examples.py
drknzz: I didn't want to change the rest of the logic much, but okay 👍

docs/_ext/autoclass_with_examples.py
```diff
@@ -100,14 +94,18 @@ def _create_hint(file_name: str, method_name: str) -> str:
         """
 
 
-def _append_hint(method_name: str, method, base_class: Any, hint: str):
+def _append_hint(method: Any, class_: Any, hint: str) -> None:
```
cptartur, line 97: I don't think `-> None` is necessary.

cptartur (approved):

docs/_ext/autoclass_with_examples.py
```diff
@@ -100,14 +94,18 @@ def _create_hint(file_name: str, method_name: str) -> str:
...
-    else:
-        method.__doc__ = (method.__doc__ or "") + hint
+    parent_method = method
+    while parent_method is not None and parent_method.__doc__ is None:
+        class_ = class_.__base__
+        parent_method = getattr(class_, method.__name__, None)
+
+    if parent_method is None:
+        method.__doc__ = hint
+    method.__doc__ = (parent_method.__doc__ or "") + hint
```
war-in, line 108: If `parent_method` is None it will fail. Probably a return under the `method.__doc__ = hint` will fix the issue

drknzz: Actually, `None.__doc__` returns `None`. But I'll reorder the if statement.

docs/_ext/autoclass_with_examples.py
drknzz: I'd rather keep it explicit. It's easier to read later, as it shows that the return type hint wasn't just omitted.

war-in (approved):
👌

PR #748, opened 2023-02-10
Fix most of `aiohttp` errors (while running parallel tests)

war-in:
Closes #

## Introduced changes

Use `pytest-rerunfailures` to automatically rerun aiohttp errors

##


pyproject.toml
```diff
@@ -46,11 +46,12 @@ pyright = "^1.1.270"
...
 isort = "^5.11.4"
+pytest-rerunfailures = "^11.1"
 
 [tool.poe.tasks]
-test.shell = "pytest -n auto -v --cov=starknet_py starknet_py"
-test_ci.shell = "coverage run -m pytest starknet_py --ignore=starknet_py/tests/e2e/docs"
-test_ci_docs.shell = "coverage run -m pytest starknet_py/tests/e2e/docs"
+test.shell = "pytest -n auto -v --reruns 10 --only-rerun aiohttp.client_exceptions.ClientConnectorError --cov=starknet_py starknet_py"
+test_ci.shell = "coverage run -m pytest -v --reruns 2 --only-rerun aiohttp.client_exceptions.ClientConnectorError starknet_py --ignore=starknet_py/tests/e2e/docs"
+test_ci_docs.shell = "coverage run -m pytest -v --reruns 2 --only-rerun aiohttp.client_exceptions.ClientConnectorError starknet_py/tests/e2e/docs"
```
drknzz, line 54: What's with the different choices for the rerun numbers?

cptartur, line 54: I think I'd avoid automatic reruns on CI. Failures due to aiohttp errors in non-parallel tests are not frequent.

war-in: Parallel tests are faster than regular tests, so aiohttp errors are frequent there. `ci` tests don't usually fail with aiohttp so a low number of reruns is reasonable. I can increase it if you think it will be better

cptartur: I wouldn't increase it above 2 in CI if we are going to keep it at all. I wouldn't want to risk our CI taking very long time because of multiple reruns.

cptartur (approved):

drknzz (approved):

PR #752, opened 2023-02-13
Add information about supported python versions

drknzz:


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -40,7 +40,7 @@ body:
     id: python-version
     attributes:
       label: Python version
-      description: What version of Python are you using?
+      description: What version of Python are you using? Note that the currently supported Python versions are 3.8 and 3.9.
```
war-in, line 43: What do you think about adding two checkboxes?
- [ ] 3.8...
- [ ] 3.9...

drknzz: I thought about restricting the choice, maybe to a dropdown list.
But the problem with this approach is that it does not specify which specific python version is being used, which might be crucial info when reproducing some bugs.
We might list every python version for 3.8 / 3.9 though 🙈

war-in: I don't think listing every version is a good idea :/ Let's maybe move the new sentence under the `What version of Python are you using?`

drknzz: I don't think it's possible in these forms

war-in (approved):

PR #754, opened 2023-02-13
`Client.net` update

### LINKED ISSUE #738: Consider changing `FullNodeClient`'s constructor
### Feature Request

Currently, FullNodeClient requires two arguments (node_url and net). `node_url` is used often, but `net` is necessary only for `chain_id` resolving. We should discuss how to fix that issue

war-in:
Closes #738 

## Introduced changes

- `Network` is declared as dataclass which holds `address` and `chain_id` of the net

##


starknet_py/net/models/chains_test.py
```diff
@@ -1,11 +0,0 @@
...
-
-from starknet_py.net.models.chains import chain_from_network
-
-
-def test_no_chain_for_custom_network():
-
-    with pytest.raises(
-        ValueError, match="Chain is required when not using predefined networks."
-    ):
-        chain_from_network("some_custom_url")
```
drknzz, line 11: Has this test been moved?

starknet_py/net/models/chains.py
```diff
@@ -1,29 +1,9 @@
...
-        TESTNET2: StarknetChainId.TESTNET2,
-    }
-
-    if isinstance(net, str) and net in mapping:
-        return mapping[net]
-
-    if not chain:
-        raise ValueError("Chain is required when not using predefined networks.")
-
-    return chain
```
drknzz, line 29: We should leave it for now and deprecate it.

starknet_py/net/networks.py
```diff
@@ -26,9 +39,28 @@ def net_address_from_net(net: str) -> str:
 
 
 def default_token_address_for_network(net: Network) -> str:
-    if net not in [TESTNET, TESTNET2, MAINNET]:
+    if net.address not in [TESTNET, TESTNET2, MAINNET]:
```
drknzz, line 42: Let's maybe have a test for this case

starknet_py/net/gateway_client.py
```diff
@@ -64,13 +64,13 @@ def __init__(
         :param session: Aiohttp session to be used for request. If not provided, client will create a session for
```
drknzz, line 64: Should the docstring be updated?

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -274,7 +275,7 @@ async def test_deploy_account_raises_on_no_enough_funds(deploy_account_details_f
                 class_hash=class_hash,
                 salt=salt,
                 key_pair=key_pair,
-                client=GatewayClient(net="testnet"),
+                client=GatewayClient(net=Network("testnet")),
```
drknzz, line 278: It's a huge breaking change, as practically almost every use case uses client somehow.
Deprecations would for sure be needed.

Also I'm not a fan of creating a new network every time a client is constructed. Maybe we could create Network objects of `testnet`, `testnet2` and `mainnet` somewhere, and just import them, the same way it was done previously. However it would still require users to change the way they use clients currently.

Another, less breaking approach, would be not to change the constructor, and create Network internally from the predefined networks.

starknet_py/net/networks.py
```diff
@@ -26,9 +39,28 @@ def net_address_from_net(net: str) -> str:
...
+        MAINNET: StarknetChainId.MAINNET,
+        TESTNET: StarknetChainId.TESTNET,
+        TESTNET2: StarknetChainId.TESTNET2,
+    }
+
+    if isinstance(net, str) and net in mapping:
+        return mapping[net]
+
+    if not chain:
+        raise ValueError("Chain is required when not using predefined networks.")
```
drknzz, line 64: I think the test was deleted 👀

starknet_py/net/networks.py
```diff
@@ -1,32 +1,38 @@
-from typing import Literal, TypedDict, Union
+from dataclasses import dataclass
+from typing import TypedDict, Union
 
 from starknet_py.constants import FEE_CONTRACT_ADDRESS
-
-MAINNET = "mainnet"
-TESTNET = "testnet"
-TESTNET2 = "testnet2"
```
drknzz, line 7: Were these used only in the context of net in client?

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -299,7 +300,7 @@ async def test_deploy_account_passes_on_enough_funds(deploy_account_details_fact
             class_hash=class_hash,
             salt=salt,
             key_pair=key_pair,
-            client=GatewayClient(net="testnet"),
+            client=GatewayClient(net=TESTNET),
```
drknzz, line 303: The problem with this approach is that it's still breaking.

starknet_py/net/models/chains.py
drknzz: Please bring it back and deprecate it, so we don't forget.

starknet_py/net/networks.py
```diff
@@ -1,32 +1,38 @@
...
+    address: Union[str, CustomGatewayUrls]
+    chain_id: StarknetChainId
 
 
-def net_address_from_net(net: str) -> str:
-    return {
-        MAINNET: "https://alpha-mainnet.starknet.io",
-        TESTNET: "https://alpha4.starknet.io",
-        TESTNET2: "https://alpha4-2.starknet.io",
-    }.get(net, net)
```
drknzz, line 25: This should be deprecated first.

PR #766, opened 2023-02-15
`Client.net` deprecation

war-in:
Closes #

## Introduced changes

- Client.net is deprecated
- Account.get_balance has chain_id parameter
- net parameter in FullNodeClient's constructor is deprecated

##

### DIFF
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -218,9 +218,13 @@ class Account(BaseAccount):
         )
 
     async def get_balance(
-        self, token_address: Optional[AddressRepresentation] = None, chain_id: Optional[StarknetChainId] = None,
+        self,
+        token_address: Optional[AddressRepresentation] = None,
+        chain_id: Optional[StarknetChainId] = None,
     ) -> int:
-        token_address = token_address or default_token_address_for_chain(chain_id or self._chain_id)
+        token_address = token_address or default_token_address_for_chain(
+            chain_id or self._chain_id
+        )
 
         low, high = await self._client.call_contract(
             Call(
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -1,7 +1,7 @@
 import dataclasses
 import re
 from collections import OrderedDict
-from typing import Dict, Iterable, List, Optional, Tuple, Union, cast
+from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 from starknet_py.common import create_compiled_contract
 from starknet_py.constants import FEE_CONTRACT_ADDRESS, QUERY_VERSION_BASE
@@ -70,9 +70,6 @@ class Account(BaseAccount):
         :param key_pair: Key pair that will be used to create a default `Signer`.
         :param chain: ChainId of the chain used to create the default signer.
         """
-        if chain is None and signer is None:
-            raise ValueError("One of chain or signer must be provided.")
-
         self._address = parse_address(address)
         self._client = client
 
@@ -84,13 +81,13 @@ class Account(BaseAccount):
                 raise ValueError(
                     "Either a signer or a key_pair must be provided in Account constructor."
                 )
+            if chain is None:
+                raise ValueError("One of chain or signer must be provided.")
 
             signer = StarkCurveSigner(
                 account_address=self.address,
                 key_pair=key_pair,
-                chain_id=cast(
-                    StarknetChainId, chain
-                ),  # cast is required, because pyright doesn't know chain is not None here
+                chain_id=chain
             )
         self.signer: BaseSigner = signer
         self._chain_id = chain
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -222,9 +222,9 @@ class Account(BaseAccount):
         token_address: Optional[AddressRepresentation] = None,
         chain_id: Optional[StarknetChainId] = None,
     ) -> int:
-        token_address = token_address or default_token_address_for_chain(
-            chain_id or self._chain_id
-        )
+        if token_address is None:
+            chain_id = chain_id or self._chain_id
+            token_address = default_token_address_for_chain(chain_id)
 
         low, high = await self._client.call_contract(
             Call(
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -4,7 +4,7 @@ from collections import OrderedDict
 from typing import Dict, Iterable, List, Optional, Tuple, Union, cast
 
 from starknet_py.common import create_compiled_contract
-from starknet_py.constants import QUERY_VERSION_BASE
+from starknet_py.constants import FEE_CONTRACT_ADDRESS, QUERY_VERSION_BASE
 from starknet_py.hash.address import compute_address
 from starknet_py.hash.selector import get_selector_from_name
 from starknet_py.net.account.account_deployment_result import AccountDeploymentResult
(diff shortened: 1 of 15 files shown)


starknet_py/net/account/account.py
```diff
@@ -96,11 +86,15 @@ def __init__(
...
                 )
 
-            chain = chain_from_network(net=client.net, chain=chain)
             signer = StarkCurveSigner(
-                account_address=self.address, key_pair=key_pair, chain_id=chain
+                account_address=self.address,
+                key_pair=key_pair,
+                chain_id=cast(
+                    StarknetChainId, chain
+                ),  # cast is required, because pyright doesn't know chain is not None here
```
drknzz, line 94: How come it wasn't required before?

starknet_py/net/models/chains.py
```diff
@@ -27,3 +28,16 @@ def chain_from_network(
...
+
+
+def default_token_address_for_chain(chain_id: Optional[StarknetChainId] = None) -> str:
+    if chain_id not in [
+        StarknetChainId.TESTNET,
+        StarknetChainId.TESTNET2,
+        StarknetChainId.MAINNET,
+    ]:
+        raise ValueError(
+            "Argument token_address must be specified when using a custom net."
```
drknzz, line 40: Should it still be net?

starknet_py/net/account/account.py
```diff
@@ -224,10 +218,12 @@ async def get_nonce(self) -> int:
...
     async def get_balance(
-        self, token_address: Optional[AddressRepresentation] = None
+        self,
+        token_address: Optional[AddressRepresentation] = None,
+        chain_id: Optional[StarknetChainId] = None,
     ) -> int:
-        token_address = token_address or default_token_address_for_network(
-            self._client.net
+        token_address = token_address or default_token_address_for_chain(
+            chain_id or self._chain_id
```
drknzz, line 226: Could we maybe do something more explicit like
```
if token_address is None:
   chain_id = chain_id or self._chain_id
   token_address = default_token_address_for_chain(chain_id)
```

starknet_py/tests/e2e/utils.py
```diff
@@ -37,7 +37,7 @@ async def get_deploy_account_details(
     )
 
     res = await fee_contract.functions["transfer"].invoke(
-        recipient=address, amount=int(1e15), max_fee=MAX_FEE
+        recipient=address, amount=int(1e20), max_fee=MAX_FEE
```
drknzz, line 40: Why this change?

starknet_py/net/account/account.py
war-in: `chain_from_network()` returns StarknetChainId or raises an error, so the chain was always StarknetChainId here

starknet_py/net/models/chains.py
war-in: Maybe `network`

starknet_py/net/account/account.py
drknzz: We could try doing an `assert` 👀

drknzz: Casting in situations like these always requires an additional comment explaining the reason behind it.
I think
```
assert chain is not None
```
is shorter and self explanatory

war-in: I know who would not like it 🙈

drknzz: 😆
It wasn't about the use of `assert` though but rather the same `assert` in multiple places in code, implying a flaw in the structure of the code

war-in: I'll leave it as open discussion 😅

Arcticae: It is perfectly viable for `chain` to be `None` here, so i don't know why the `cast` is here - py**right** seems to be **right** 🙃

starknet_py/net/account/base_account.py
```diff
@@ -51,13 +51,16 @@ async def get_nonce(self) -> int:
...
+        token_address: Optional[AddressRepresentation] = None,
+        chain_id: Optional[StarknetChainId] = None,
     ) -> int:
         """
         Checks account's balance of specified token.
 
         :param token_address: Address of the ERC20 contract.
-            If not specified it will be payment token address.
+        :param chain_id: Identifier of the Starknet chain used.
+            If token_address is not specified it will be used to determine network's payment token address.
```
Arcticae, line 63: it would be also irrelevant when `token_address` is provided - worth to note here

starknet_py/net/models/chains.py
```diff
@@ -27,3 +28,16 @@ def chain_from_network(
...
     return chain
+
+
+def default_token_address_for_chain(chain_id: Optional[StarknetChainId] = None) -> str:
+    if chain_id not in [
+        StarknetChainId.TESTNET,
+        StarknetChainId.TESTNET2,
+        StarknetChainId.MAINNET,
+    ]:
+        raise ValueError(
```
Arcticae, line 39: [nitpick]: The error message here might depend on the function usage context, so i'd avoid throwing in those kind of methods. I guess for the use-case is fine, just noting here.

starknet_py/net/models/chains.py
```diff
@@ -27,3 +28,16 @@ def chain_from_network(
         raise ValueError("Chain is required when not using predefined networks.")
 
     return chain
+
+
+def default_token_address_for_chain(chain_id: Optional[StarknetChainId] = None) -> str:
```
Arcticae, line 33: i struggle to understand why the chain_id can be `None` here, kind of weird function definition.

starknet_py/net/full_node_client.py
```diff
@@ -65,10 +66,14 @@ def __init__(
...
         self._client = RpcHttpClient(url=node_url, session=session)
+
+        if net is not None:
+            warnings.warn("Parameter net is deprecated.", category=DeprecationWarning)
         self._net = net
 
     @property
-    def net(self) -> Network:
+    def net(self) -> Optional[Network]:
+        warnings.warn("Parameter net is deprecated.", category=DeprecationWarning)
```
Arcticae, line 76: This doesn't tell which class doesn't accept the parameter anymore. Did you check manually how does the trace look when using the library with `net`, and if user can tell where the `warn` has its' origin?

Arcticae (commented):
It looks good from API-construction perspective, i'd circle back on the `cast` though.
Also, some nitpicking but overall looks fine 👍

starknet_py/net/models/chains.py
war-in: It can be None, because 
1. an account can be created without passing the chain
2. if 1 and the user does not provide chain_id in the get_balance => we do not know the chain

starknet_py/net/account/account.py
Arcticae: I actually think that having this kind of implicit logic here makes for confusing code reading experience, did you guys consider changing it?

starknet_py/net/models/chains.py
Arcticae: Can't `StarknetChainId` have this method?

starknet_py/net/account/account.py
drknzz: **chain** cannot be `None` here, the check for it is few lines above:
```
        if chain is None and signer is None:
            raise ValueError("One of chain or signer must be provided.")
```

drknzz: I agree that the code is hard to follow though.
There was some attempt at making it more explicit in [this _closed_ PR](https://github.com/software-mansion/starknet.py/pull/736/files#diff-71fc6b5ba6c214b0ddf79941772b54fef34507113a848e666dbc8e9c1002a0f3).
Maybe it should be implemented in some other PR nonetheless.

war-in: it is fixed now! Check it out

drknzz (approved):

Arcticae (approved):

PR #776, opened 2023-02-20
Add Codecov configuration file

### LINKED ISSUE #777: Make Codecov notify after CI
### Feature Request

Currently Codecov creates a comment (and sends an email) after each action build.
It should wait for all jobs to be completed and only then send a notification with full report.

drknzz:
Closes #777

Add codecov.yml

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,3 @@
+codecov:
+  notify:
+    wait_for_ci: yes
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,3 +1,3 @@
 codecov:
   notify:
-    wait_for_ci: yes
+    after_n_builds: 2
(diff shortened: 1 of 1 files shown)


.github/codecov.yml
```diff
@@ -0,0 +1,3 @@
+codecov:
+  notify:
+    after_n_builds: 2
```
cptartur, line 3: Not sure if `2` is a good choice. If I understand correctly, whole checks run will make 4 codecov uploads. One per package tests per version (2), and one per docs tests per version (2).

drknzz: Right, I didn't take into consideration that the checks run for 2 versions.
However it works - the notification was pushed after all the checks have finished (at least that was how it looked).

drknzz: Actually, now that I've looked, the comment is edited a minute after it is created.

PR #793, opened 2023-02-22
Bump `crypto-cpp-py` and `starknet-devnet`

cptartur:
Closes #

## Introduced changes

- 
- 

##


starknet_py/tests/e2e/client/client_test.py
```diff
@@ -73,7 +73,7 @@ async def test_get_transaction_raises_on_not_received(client):
     with pytest.raises(
         TransactionNotReceivedError, match="Transaction was not received on Starknet."
     ):
-        await client.get_transaction(tx_hash=0x1)
+        await client.get_transaction(tx_hash=0x9999)
```
drknzz, line 76: Will it work on say, integration network?

cptartur: Good question. Would a previous test work though?

drknzz: Good question.

drknzz (approved):

PR #797, opened 2023-02-23
Fix unit tests

drknzz:
Fix a part of unit tests


starknet_py/compile/compiler.py
```diff
@@ -129,7 +129,7 @@ def starknet_compile(
         main_scope=MAIN_SCOPE,
         add_debug_info=False,
         file_contents_for_debug_info=file_contents_for_debug_info,
-        filter_identifiers=False,
```
war-in, line 132: 👀

starknet_py/tests/e2e/client/fixtures/prepare_network.py
```diff
@@ -138,7 +138,7 @@ def fixture_contract_address(prepare_network: Tuple[str, PreparedNetworkData]) -
     return prepared_data.contract_address
 
 
-@pytest.fixture(name="balance_contract", scope="package")
+@pytest.fixture(name="balance_contract", scope="module")
```
war-in, line 141: Why do we need to change the scope?

war-in: Discussed. Can be reverted to `package`

.github/workflows/checks.yml
```diff
@@ -12,8 +12,9 @@ jobs:
     name: Download deps, install poetry
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
war-in, line 15: Does it mean that if one test fails the others will continue running?

drknzz: [Yes](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)

starknet_py/compile/compiler.py
drknzz: It has been changed in cairo-lang.

.github/workflows/checks.yml
drknzz: But I don't think we need the matrix anymore, as only python3.9 is supported

cptartur: I wouldn't include it if we are using only `3.9` for now anyways. I'd address changing the checks in the separate issue.

drknzz: `fail-fast` only applies to a strategy for *matrix* anyway. If set to true, the same jobs (but for different items from matrix) fail, once any job fails.

war-in (approved):

PR #810, opened 2023-02-27
Update `get_transaction` and `get_block`

### LINKED ISSUE #807: Update `get_transaction` and `get_block`
### Feature Request

`get_transaction`

- contract_address is changed to sender_address in invoke transactions
- Declare v2 transaction may be returned

`get_block`

- The transactions field is subject to the same changes in get_transaction

cptartur:
Closes #807 

## Introduced changes

- 
- 

##

### DIFF
--- a/starknet_py/net/schemas/gateway.py
+++ b/starknet_py/net/schemas/gateway.py
@@ -44,7 +44,9 @@ from starknet_py.net.schemas.common import (
     StatusField,
     StorageEntrySchema,
 )
-from starknet_py.net.schemas.utils import replace_sender_address_with_contract_address
+from starknet_py.net.schemas.utils import (
+    replace_invoke_sender_address_with_contract_address,
+)
 
 # pylint: disable=unused-argument, no-self-use
 
@@ -95,7 +97,7 @@ class InvokeTransactionSchema(TransactionSchema):
 
     @post_load
     def make_dataclass(self, data, **kwargs) -> InvokeTransaction:
-        replace_sender_address_with_contract_address(data)
+        replace_invoke_sender_address_with_contract_address(data)
         return InvokeTransaction(**data)
 
 
--- a/starknet_py/net/schemas/gateway.py
+++ b/starknet_py/net/schemas/gateway.py
@@ -115,6 +115,7 @@ class DeclareTransactionSchema(TransactionSchema):
     class_hash = Felt(data_key="class_hash", required=True)
     sender_address = Felt(data_key="sender_address", required=True)
     nonce = Felt(data_key="nonce", load_default=None)
+    compiled_class_hash = Felt(data_key="compiled_class_hash", load_default=None)
 
     @post_load
     def make_dataclass(self, data, **kwargs) -> DeclareTransaction:
--- a/starknet_py/net/schemas/gateway.py
+++ b/starknet_py/net/schemas/gateway.py
@@ -45,7 +45,7 @@ from starknet_py.net.schemas.common import (
     StorageEntrySchema,
 )
 from starknet_py.net.schemas.utils import (
-    replace_invoke_sender_address_with_contract_address,
+    replace_invoke_contract_address_with_sender_address,
 )
 
 # pylint: disable=unused-argument, no-self-use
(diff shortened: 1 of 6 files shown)


cptartur:
Support for Declare v2 is needed for this to be ready.

starknet_py/net/schemas/gateway.py
```diff
@@ -81,13 +82,15 @@ class TransactionSchema(Schema):
 
 
 class InvokeTransactionSchema(TransactionSchema):
-    contract_address = Felt(data_key="contract_address", required=True)
+    contract_address = Felt(data_key="contract_address", load_default=None)
+    sender_address = Felt(data_key="sender_address", load_default=None)
```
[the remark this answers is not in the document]
war-in: Wouldn't old transactions include that field? Or an update will change it everywhere?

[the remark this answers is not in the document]
cptartur: From what I checked, it seems that v0 transactions (still present in old blocks) have `contract_address` field. v1 transactions do have `sender_address` now, so it seems like it was replaced but only in newer versions.

starknet_py/net/schemas/utils.py
```diff
@@ -0,0 +1,11 @@
+def replace_sender_address_with_contract_address(data: dict):
+    data["contract_address"] = data.get("contract_address") or data.get(
+        "sender_address"
+    )
+
+    if data["contract_address"] is None:
+        raise ValueError(
+            "Missing field `contract_address` or `sender_address` for InvokeTransactionSchema."
```
drknzz, line 8: Do we want to be mentioning specific `InvokeTransactionSchema` in a function from other module?

drknzz (approved):

starknet_py/net/schemas/utils.py
cptartur: Good point. I think we do. It is used in both rpc and gateway `InvokeTransactionSchema`.

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -65,6 +65,32 @@ async def test_get_declare_v2_transaction():
...
+            hash=0x722B666CE83EC69C18190AAE6149F79E6AD4B9C051B171CC6C309C9E0C28129,
+            max_fee=0x38D7EA4C68000,
+            signature=[
+                0x6F3070288FB33359289F5995190C1074DE5FF00D181B1A7D6BE87346D9957FE,
+                0x4AB2D251D18A75F8E1AD03ABA2A77BD3D978ABF571DC262C592FB07920DC50D,
+            ],
+            nonce=1,
+            version=2,
+        )
+        in block.transactions
```
drknzz, line 90: I wanted to ask what about the transaction receipts in the block, but it seems we do not have a field for receipts in the block dataclass.

drknzz (approved):

starknet_py/net/schemas/gateway.py
```diff
@@ -85,13 +88,15 @@ class TransactionSchema(Schema):
...
     calldata = fields.List(Felt(), data_key="calldata", required=True)
     entry_point_selector = Felt(data_key="entry_point_selector", load_default=None)
     nonce = Felt(data_key="nonce", load_default=None)
 
     @post_load
     def make_dataclass(self, data, **kwargs) -> InvokeTransaction:
+        replace_invoke_sender_address_with_contract_address(data)
```
war-in, line 99: Shouldn't we replace contract_address with sender_address?

starknet_py/tests/e2e/client/gateway_test.py
war-in: Probably because of RPC specs

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -38,6 +39,58 @@ async def test_get_class_by_hash_sierra_program():
     assert isinstance(contract_class.abi[0], dict)
 
 
+@pytest.mark.asyncio
+async def test_get_declare_v2_transaction():
```
war-in, line 43: This and the next one should be marked as temporary tests

starknet_py/net/schemas/gateway.py
cptartur: It's an mistake in the function name - it does replace contract address with sender.

starknet_py/tests/e2e/client/gateway_test.py
cptartur: Rpc only has `getBlockWithTxHashes` and `getBlockWithTxs` which includes only information similar to this from `get_transaction`. So no way to map these.

war-in (approved):

PR #812, opened 2023-02-28
Update `get_state_update`

### LINKED ISSUE #806: Update `get_state_update`
### Feature Request

- added replaced_classes section
- declared_classes is split into old_declared_classes and declared_classes
- old_declared_classes is a list of class hashes
- declared_classes is a list of (class_hash,compiled_class_hash) pairs

cptartur:
Closes #806 

## Introduced changes

- 
- 

##

- Changed `StateDiff.declared_contract_hashes` from `List[int]` to `List[DeclaredContractHash]` to follow the changes in the gateway and the rpc.


starknet_py/net/client_models.py
```diff
@@ -284,10 +284,17 @@ class ContractsNonce:
...
+
+
 @dataclass
 class StateDiff:
     deployed_contracts: List[DeployedContract]
-    declared_contract_hashes: List[int]
+    deprecated_declared_contract_hashes: List[int]
+    declared_contract_hashes: List[DeclaredContractHash]
     storage_diffs: List[StorageDiffItem]
     nonces: List[ContractsNonce]
```
cptartur, line 299: Gateway's state diff now includes a `replaced_classes` section that is not present in the RPC master branch. Should it be added as an optional field or omitted?

drknzz: I think we shouldn't care for the new RPC compliance right now. It will be taken care of, once a new version comes out.

starknet_py/net/schemas/gateway.py
```diff
@@ -299,17 +300,31 @@ def make_dataclass(self, data, **kwargs):
...
     )
-    declared_contract_hashes = fields.List(
+    deprecated_declared_contract_hashes = fields.List(
         Felt(),
         data_key="declared_contracts",
         required=True,
     )
+    declared_contract_hashes = fields.List(
+        fields.Nested(DeclaredContractHashSchema()),
+        data_key="declared_contracts",
```
drknzz, line 325: This should be `declared_classes`

starknet_py/net/schemas/gateway.py
```diff
@@ -299,17 +300,31 @@ def make_dataclass(self, data, **kwargs):
...
 class StateDiffSchema(Schema):
     deployed_contracts = fields.List(
         fields.Nested(DeployedContractSchema()),
         data_key="deployed_contracts",
         required=True,
     )
-    declared_contract_hashes = fields.List(
+    deprecated_declared_contract_hashes = fields.List(
         Felt(),
         data_key="declared_contracts",
```
drknzz, line 320: This should be `old_declared_classes`

drknzz (changes requested):

starknet_py/net/client_models.py
cptartur: Turns out it will be added to the RPC, will add it to the dataclass as well.

https://github.com/starkware-libs/starknet-specs/pull/76

cptartur: It's a list, so we don't have to make it optional - we can default to an empty list.

starknet_py/net/schemas/gateway.py
```diff
@@ -299,15 +301,38 @@ def make_dataclass(self, data, **kwargs):
...
+    class_hash = Felt(data_key="class_hash", required=True)
+    compiled_class_hash = Felt(data_key="compiled_class_hash", required=True)
+
+    @post_load
+    def make_dataclass(self, data, **kwargs) -> DeclaredContractHash:
+        return DeclaredContractHash(**data)
+
+
+class ReplacedClassSchema(Schema):
+    contract_address = Felt(data_key="address", required=True)
```
drknzz, line 314:
```suggestion
    address = Felt(data_key="address", required=True)
```
Why add the contract_ prefix?

drknzz (approved):

starknet_py/net/schemas/gateway.py
cptartur: It is named like that in the RPC specs. I can change it to just `address` if you think that is better?

drknzz: Hmm, it's a schema for gateway, where the field is `address` (as in data_key). Up to you.

cptartur: I'd prefer we follow RPC specs.

starknet_py/net/client_models.py
```diff
@@ -295,13 +295,20 @@ class DeclaredContractHash:
...
+
+
 @dataclass
 class StateDiff:
     deployed_contracts: List[DeployedContract]
     deprecated_declared_contract_hashes: List[int]
     declared_contract_hashes: List[DeclaredContractHash]
     storage_diffs: List[StorageDiffItem]
     nonces: List[ContractsNonce]
+    replaced_classes: List[ReplacedClass]
```
drknzz, line 311: Adding this as non-optional makes the current version of library not compliant with any rpc spec version.
Do we plan to support rpc spec v0.3.0 in the next release?

starknet_py/tests/e2e/client/gateway_test.py
```diff
@@ -38,6 +39,24 @@ async def test_get_class_by_hash_sierra_program():
...
 
+@pytest.mark.asyncio
+async def test_get_new_state_update():
+    client = GatewayClient(
+        net="https://external.integration.starknet.io"
+    )  # TODO: Replace this with fixture
+
+    state_update = await client.get_state_update(block_number=283364)
+
+    assert state_update.state_diff.replaced_classes == []
```
drknzz, line 50: Would be awesome to find a block with something in `replaced_classes`

starknet_py/net/client_models.py
cptartur: Making this non-optional is not problematic because we can just provide empty list if `replaced_classes` are not present in the response. And it will be added to the RPC spec anyway https://github.com/starkware-libs/starknet-specs/pull/76

starknet_py/tests/e2e/client/gateway_test.py
cptartur: Tried replacing class hashes myself and such block now exists `0x656164aefe7d646927fcbf8c576420475fcd5602bf284c00aa00eb3d1b8151b`.

war-in (approved):

PR #815, opened 2023-03-01
Comment dataclasses

### LINKED ISSUE #774: Remove transaction dataclasses in `client_models.py`
### Feature Request

New transaction dataclasses have been added in #649, leaving us with 2 dataclasses for each transaction:
```
net/models/transaction.py
net/client_models.py
```

We should possibly remove the old transaction classes and replace their usages with the new ones.

drknzz:
Closes #774

## Introduced changes

- Add module comments about the purpose of dataclasses

##

### DIFF
--- a/starknet_py/net/models/transaction.py
+++ b/starknet_py/net/models/transaction.py
@@ -1,5 +1,6 @@
 """
-Dataclasses representing Transactions for internal use.
+Dataclasses representing Transactions for library use, most often
+when sending a transaction to Starknet.
 They should be compliant with the latest Starknet version.
 """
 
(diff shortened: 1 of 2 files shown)


starknet_py/net/models/transaction.py
```diff
@@ -1,3 +1,8 @@
+"""
+Dataclasses representing Transactions for internal use.
```
cptartur, line 2: These are not only for internal use, it is a public API.

cptartur (changes requested):

starknet_py/net/models/transaction.py
drknzz: Would `library use` be better?

cptartur (approved):

PR #817, opened 2023-03-02
Estimate fee skip_validate

### LINKED ISSUE #808: Consider supporting `skip_validate` in gateway `estimate_fee`
### Feature Request

Can now take an additional skip_validate flag in the URL. If set to true, then __validate__ execution will be skipped. This can assist hardware wallets in not having to sign for fee estimations.

drknzz:
Closes #808 

## Introduced changes

- Add keyword parameter `skip_validate` in *GatewayClient.estimate_fee*
- Make `prepare_invoke` public

##


starknet_py/net/client.py
```diff
@@ -187,13 +187,16 @@ async def estimate_fee(
         tx: AccountTransaction,
         block_hash: Optional[Union[Hash, Tag]] = None,
         block_number: Optional[Union[int, Tag]] = None,
+        *,
+        skip_validate: bool = False,
```
cptartur, line 191: I'd first try to get confirmation from starkware that this will be included in the RPC at some point. If not, this argument should not be a part of the interface.

starknet_py/net/utils.py
```diff
@@ -0,0 +1,84 @@
+from collections import OrderedDict
```
cptartur, line 1: Account related utils should be in `/account`, not in top level.

starknet_py/contract.py
```diff
@@ -323,10 +326,16 @@ async def estimate_fee(
             (or "latest" / "pending" for the latest / pending block), default is "pending".
         :return: Estimated amount of Wei executing specified transaction will cost.
         """
-        tx = await self._account.sign_invoke_transaction(calls=self, max_fee=0)
+        nonce = await self._account.get_nonce()
+        tx = await prepare_invoke(
+            calls=self, contract_address=self._account.address, nonce=nonce
+        )
```
cptartur, line 332: Why was this changed?

cptartur (changes requested):

starknet_py/net/utils.py
drknzz: Preparing an invoke transaction is not Account related, is it?

drknzz: (A dummy one, from arguments, without the signature)

starknet_py/net/account/account.py
```diff
@@ -124,18 +122,12 @@ async def _get_max_fee(
 
         return max_fee
 
-    async def _prepare_invoke(
-        self,
-        calls: Calls,
-        max_fee: Optional[int] = None,
-        auto_estimate: bool = False,
-    ) -> Invoke:
+    async def prepare_invoke(self, calls: Calls, max_fee: int = 0) -> Invoke:
```
cptartur, line 125: I'm still unsure about this method. It is a bit weird having method like that which is useful mostly as a helper for the other account methods exposed as public. 

If you think there is a valid reason to expose this I think it should be made into static/completely separate method.

cptartur: cc @war-in

war-in: Agreed, why was it exposed?

drknzz: It's only used in `Account.sign_invoke_transaction` as a 'helper'. It's an independent method, that has its own use cases.

I don't really see why it shouldn't be a part of Account api. The method creates an unsigned invoke transaction from calls. There is no other exposed function in the library to create Invoke from calls apart from `sign_invoke_transaction`, which even from the name does not suggest it does that - it suggests it signs an Invoke transaction.

Also, if it wasn't a part of Account, the version, nonce, contract_address essential fields would need to be filled manually later, after using this method, for the transaction to be sent (its fee to be estimated).

drknzz: So we could either leave `prepare_invoke` exposed as it is, or not support the `skip_validate` flag.
I don't see any other solution that would make sense here.
@war-in

cptartur:
We won't be adding this as of writing this

PR #837, opened 2023-03-08
Add block_number to Contract methods

### LINKED ISSUE #823: Support `block_number` argument in Contract methods
### Feature Request

Argument `block_number` has been a part of Client interface for a good while. Contract methods should be updated to include this argument alongside the `block_hash`.

drknzz:
Closes #823 

It would be best to have both `block_number` and `block_hash` as *keyword only* parameters, however it would be breaking.

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -256,26 +256,32 @@ class PreparedFunctionCall(Call):
     async def call_raw(
         self,
         block_hash: Optional[str] = None,
+        block_number: Optional[Union[int, Tag]] = None,
     ) -> List[int]:
         """
         Calls a method without translating the result into python values.
 
         :param block_hash: Optional block hash.
+        :param block_number: Optional block number.
         :return: list of ints.
         """
-        return await self._client.call_contract(call=self, block_hash=block_hash)
+        return await self._client.call_contract(
+            call=self, block_hash=block_hash, block_number=block_number
+        )
 
     async def call(
         self,
         block_hash: Optional[str] = None,
+        block_number: Optional[Union[int, Tag]] = None,
     ) -> TupleDataclass:
         """
         Calls a method.
 
         :param block_hash: Optional block hash.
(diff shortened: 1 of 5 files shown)


starknet_py/contract.py
```diff
@@ -256,26 +256,32 @@ def _account(self) -> BaseAccount:
...
         """
-        return await self._client.call_contract(call=self, block_hash=block_hash)
+        return await self._client.call_contract(
+            call=self, block_hash=block_hash, block_number=block_number
+        )
 
     async def call(
         self,
         block_hash: Optional[str] = None,
+        block_number: Optional[Union[int, Tag]] = None,
```
cptartur, line 275: What if we made at least `block_number` a keyword only one? 

Ideally there'd be a way to detect if `block_hash` was passed as a positional argument and show a warning, but that might not be possible.

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -167,7 +167,7 @@ async def test_call_contract(client, contract_address):
         calldata=[],
     )
 
-    result = await client.call_contract(call, block_hash="latest")
+    result = await client.call_contract(call, block_number="latest")
```
cptartur, line 170: Why was this changed?

starknet_py/contract.py
drknzz: I'd rather keep both of them simple, non keyword only / both keyword.

starknet_py/tests/e2e/client/client_test.py
drknzz: To avoid confusion, as `latest` / `pending` is not a valid `block_hash`.

cptartur: It's still valid in terms of `Client` interface, but I'm fine with the change.

cptartur (approved):

PR #844, opened 2023-03-09
Remove hardcoded version from `Account.sign_for_fee_estimate`

### LINKED ISSUE #842: [BUG] `Account.sign_for_fee_estimate` uses hardcoded version
### What happened

`Account.sign_for_fee_estimate` uses hardcoded `self.supported_transaction_version + QUERY_VERSION_BASE`, instead it should use `transaction.version + QUERY_VERSION_BASE`.

Current approach is problematic in cases like DeclareV2 where one transaction uses a different version than the rest of the Account.

### Stack trace

n/a

### Steps to reproduce

n/a

### SDK Version

development

### Python version

3.9

Mac

cptartur:
Closes #842 

## Introduced changes

- `Account.sign_for_fee_estimate` now uses version of transaction + 2**128 instead of using hardcoded supported_transaction_version (1) + 2**128

##

### DIFF
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -233,7 +233,7 @@ class Account(BaseAccount):
     async def sign_for_fee_estimate(
         self, transaction: TypeAccountTransaction
     ) -> TypeAccountTransaction:
-        version = self.supported_transaction_version + QUERY_VERSION_BASE
+        version = transaction.version + QUERY_VERSION_BASE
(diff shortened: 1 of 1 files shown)


drknzz (approved):
I suppose there already exists a test for it?

cptartur:
> I suppose there already exists a test for it?

There are some tests for `sign_for_fee_estimate` so it should be covered.

PR #848, opened 2023-03-13
Setup rust on CI

### LINKED ISSUE #845: Consider setting up Cairo compiler in CI for devnet
### Feature Request

Newest devnet version requires a path to Cairo1 compiler. Without it, some tests on CI will always fail.

We might wait to see how it will be supported on devnet's side.

cptartur:
Closes #845 

## Introduced changes

This PR creates an initial rust setup in e2e test checks. 

It does not setup rust for docs test.

Some initial form of caching rust dependencies was also added.

##

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -100,6 +100,36 @@ jobs:
         if: steps.cache-contracts.outputs.cache-hit != 'true'
         run: |
           poetry run poe compile_contracts
+      - name: Install rust
+        run: |
+          curl https://sh.rustup.rs -sSf | sh -s -- -y
+      - name: Clone Cairo1 compiler repository
+        uses: actions/checkout@v3
+        with:
+          repository: starkware-libs/cairo
+          persist-credentials: false
+          ref: v1.0.0-alpha.4
+          path: cairo
+      - name: Cache rust dependencies
+        # Taken from https://github.com/actions/cache/blob/main/examples.md#rust---cargo
+        id: cache-rust
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -100,6 +100,36 @@ jobs:
...
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Build compiler
+        if: steps.cache-rust.outputs.cache-hit != 'true'
+        working-directory: ./cairo
+        run: |
+          cargo build
+      - name: Create manifest file
+        run: |
+          readlink -f cairo/Cargo.toml >> starknet_py/tests/e2e/manifest-path
```
THenry14, line 132: what is the purpose of this?

cptartur: https://github.com/Shard-Labs/starknet-devnet/pull/415 devnet requires passing additional flag `--cairo-compiler-manifest` with the path to the main Cargo.toml of cairo1 compiler. Otherwise Declare v2 transactions won't work in devnet.

I figured out that the easiest solution for now would be to read this path from `manifest-path` file (you can see `manifest-path.template` in the code) and set the flag if it exists.

THenry14: This will be needed in this CI right? if so, you could just export this as env variable, which would be cleaner and less error prone https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

THenry14 (approved):

.github/workflows/checks.yml
cptartur: I went with the env variable approach initially. But handling that in the pytest fixtures to run devnet seemed more complicated than creating a file so went with that instead.

I don't know what the future plans for devnet are. Perhaps they will start bundling cairo1 compiler somehow and we won't have to do it on our side?

PR #858, opened 2023-03-15
Support calculating Casm class hash

### LINKED ISSUE #814: Support calculating new `compiled_class_hash`
### Feature Request

For now w can use one from cairo-lang, but it needs to be reimplemented.

drknzz:
Closes #814


starknet_py/hash/casm_class_hash.py
```diff
@@ -0,0 +1,62 @@
...
+        _entry_points_array(_entry_points.external)
+    )
+    l1_handler_entry_points_hash = poseidon_hash_many(
+        _entry_points_array(_entry_points.l1_handler)
+    )
+    constructor_entry_points_hash = poseidon_hash_many(
+        _entry_points_array(_entry_points.constructor)
+    )
+
+    _input_to_hash = dict(program=casm_contract_class.program)
```
war-in, line 31:
```suggestion
    _program_to_hash = dict(program=casm_contract_class.program)
```

starknet_py/hash/casm_class_hash.py
war-in, line 1: If it is how casm class_hash is computed I'm fine with that.
Is there any documentation about it?

starknet_py/hash/casm_class_hash_test.py
```diff
@@ -0,0 +1,32 @@
+import json
+from typing import cast
+
+import pytest
+from starknet_devnet.contract_class_utils import parse_casm
```
war-in, line 5: 👀 I think this is too much for me

starknet_py/hash/casm_class_hash.py
drknzz: I haven't found any documentation. I mainly looked at starknet-rs / cairo / cairo-lang (cairo-lang uses cairo to calculate the hash).

starknet_py/hash/casm_class_hash_test.py
drknzz: Yeah, it should be a part of the new `cairo-lang`. Let's wait with this PR for that.

cptartur: Worst case, let's commit this file ourselves instead of depending on devnet.

starknet_py/hash/casm_class_hash.py
```diff
@@ -0,0 +1,62 @@
...
+            hinted_casm_class_hash,
+            bytecode_hash,
+        ]
+    )
+
+
+def _entry_points_array(entry_points: List[CasmClassEntryPoint]) -> List[int]:
+    entry_points_array = []
+    for entry_point in entry_points:
+        assert entry_point.builtins is not None
```
cptartur, line 54: I'd prefer an explicit exception instead of assert here.

cptartur (commented):
If the hashes are correct compared to cairo-lang, it's seems fine 😅

starknet_py/hash/casm_class_hash.py
war-in: I knew it!

PR #862, opened 2023-03-16
Rename Deployer.create_deployment_call

### LINKED ISSUE #820: Consider renaming `Deployer.create_deployment_call`
### Feature Request

The name suggests to me that a `Call` object will be returned while actually it returns a `ContractDeployment` named tuple with call and address. I think a better name is needed, perhaps something like `create_deployment` or `create_contract_deployment`.

drknzz:
Closes #820 

## Introduced changes

- deprecate `create_deployment_call`
- deprecate `create_deployment_call_raw`
- add `create_contract_deployment`
- add `create_contract_deployment_raw`

##


war-in (commented):
Changes are good, just wondering if we want to do it now 😅 
There are many breaking changes this time. What about deprecating these two and adding new ones?

cc @cptartur

drknzz:
> Changes are good, just wondering if we want to do it now 😅

I though we might do it now so there's less breaking changes later, but it's whatever to me.

cptartur:
I'd definitely avoid introducing new breaking changes in the upcoming release if possible. I'd follow the standard deprecation flow with these.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -66,20 +66,45 @@ def create_deployment_call(
         """
         Creates deployment call to the UDC contract
 
+         .. deprecated:: 0.15.0
+            Function create_deployment_call is deprecated and will be removed in the future.
```
war-in, line 70: Maybe it is worth mentioning that we just renamed it?

PR #873, opened 2023-03-16
Remove eth deps

cptartur:
Closes #

## Introduced changes

- Removed some leftover dependencies leftover from the L1<>L2 messaging module. 
- Updated the lock.

##


war-in (approved):

pyproject.toml
```diff
@@ -23,13 +23,11 @@ marshmallow = "^3.15.0"
 marshmallow-oneofschema = "^3.0.1"
 typing-extensions = "^4.3.0"
 cairo-lang = { url = "https://github.com/starkware-libs/cairo-lang/releases/download/v0.11.0-pre/cairo-lang-0.11.0a0.zip" }
-eth-utils = "^2.1.0"
```
drknzz, line 26: We're actually using eth-utils directly, for example [here](https://github.com/software-mansion/starknet.py/blob/a52ed29691db998e7b1f90cdcf28121e6383cf9e/starknet_py/hash/utils.py#L5).
Is it not a problem?

war-in: It definitely is! I didn't know we are using it :o

cptartur: Right, didn't realize these come from the eth-utils. Restoring.

drknzz (approved):

cptartur:
Will be easier to address this in the future.

https://github.com/software-mansion/starknet.py/issues/891

PR #878, opened 2023-03-17
Add EthProxyCheck

### LINKED ISSUE #856: Missing a ProxyConfig for the ETH contract
### Feature Request

The [ProxyCheck](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/proxy/proxy_check.py) defines two classes for Argent and OpenZeppelin contracts. However, the standard ETH token uses another convention and requires to call the `implementation` function instead.

A basic fix would be to add 
```
    class EthProxyCheck(ProxyCheck):
        async def implementation_address(
            self, address: Address, client: Client
        ) -> Optional[int]:
            return await self.get_implementation(address, client)

        async def implementation_hash(
            self, address: Address, client: Client
        ) -> Optional[int]:
            return await self.get_implementation(address, client)

        @staticmethod
        async def get_implementation(address: Address, client: Client) -> Optional[int]:
            call = Call(
                to_addr=address,
                selector=get_selector_from_name("implementation"),
                calldata=[],
            )
            (implementation,) = await client.call_contract(call=call)
            return implementation
```

copy/pasted from ArgentProxyCheck with `implementation` instead of `get_implementation`. Another option would be to drop these classes and implement some wild guess rules, such as "if a method contains "implementation" then it's probably a proxy, then try to follow the hash (this kind of things done by starkscan for instance).

Up to you

drknzz:
Closes #856

## Introduced changes

- Add EthProxyCheck and use it by default
- remove deprecated max_steps from ProxyConfig
- rename proxy fixtures

##


starknet_py/tests/e2e/docs/guide/test_resolving_proxies.py
```diff
@@ -32,34 +28,34 @@ async def test_resolving_proxies(
...
+    # It will check if your proxy is OpenZeppelin proxy / ArgentX proxy / proxy of Starknet Eth contract
     contract = await Contract.from_address(
         address=address, client=gateway_client, proxy_config=True
     )
 
     # After that contract can be used as usual
     # docs-1: end
     # docs-2: start
-    # To resolve proxy contract other than OpenZeppelin or ArgentX, a custom ProxyCheck is needed
+    # To resolve proxy contract other than OpenZeppelin / ArgentX / Starknet Eth proxy, a custom ProxyCheck is needed
```
drknzz, line 42: Maybe it should be ERC20 proxy or sth, but I wanted to keep the names consistent with the ProxyCheck name. Would any of StarknetEthProxyCheck / StarknetERC20ProxyCheck / ERC20ProxyCheck be better?

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -33,15 +33,10 @@ class ProxyConfig(TypedDict, total=False):
 
 
 def prepare_proxy_config(proxy_config: ProxyConfig) -> ProxyConfig:
-    if "max_steps" in proxy_config:
-        warnings.warn(
-            "ProxyConfig.max_steps is deprecated. Contract.from_address always makes at most 1 step.",
-            category=DeprecationWarning,
-        )
```
cptartur, line 40: This removal should be included in the migration guide.

starknet_py/proxy/proxy_check.py
```diff
@@ -77,3 +77,25 @@ async def _get_storage_at_or_none(
...
+class EthProxyCheck(ProxyCheck):
+    async def implementation_address(
+        self, address: Address, client: Client
+    ) -> Optional[int]:
+        return await self.implementation(address, client)
+
+    async def implementation_hash(
+        self, address: Address, client: Client
+    ) -> Optional[int]:
+        return await self.implementation(address, client)
```
cptartur, line 91: I think we can verify if it returns implementation address or class hash and explicitly support only one of them (return None in the other case).

starknet_py/tests/e2e/fixtures/proxy.py
```diff
@@ -71,9 +75,26 @@ async def deploy_proxy_to_contract_custom(
...
+        compiled_proxy_name, compiled_contract_name, new_gateway_account_client
+    )
+
+
+@pytest_asyncio.fixture(
+    name="proxy_impl_func",
+    params=[("oz_proxy_impl_func_compiled.json", "map_compiled.json")],
 )
-async def deploy_proxy_to_contract_exposed(
+async def deploy_proxy_to_contract_impl_func(
```
cptartur, line 97: Why was this renamed?

drknzz: In order to show the different `CustomProxyCheck` in docs guide. The old custom proxy check behaved similarly to how the eth proxy check behaves now.

starknet_py/tests/e2e/mock/contracts_compiled/precompiled/eth_proxy_compiled.json
war-in, line 1: Can't it be compiled as usual?

docs/migration_guide.rst
```diff
@@ -20,6 +20,8 @@ Deprecation
 - ``compute_invoke_hash`` is deprecated in favour of ``compute_transaction_hash``
 - ``starknet_py.common.create_contract_class`` is deprecated in favour of ``starknet_py.common.create_compiled_contract``
 
+- Removed deprecated ``max_steps`` in :class:`~starknet_py.proxy.contract_abi_resolver.ProxyConfig`.
```
cptartur, line 23: I think it technically should go to "Breaking changes".

starknet_py/tests/e2e/mock/contracts_compiled/precompiled/eth_proxy_compiled.json
drknzz: No, because it contains `delegate_call` (proxying through address), which is no longer present in current versions

docs/migration_guide.rst
drknzz: Probably, but the breaking changes section is really long this version and this removal still touches the deprecation subject

cptartur: I wouldn't hide the change just because there are already a lot of other breaking changes. It's not the place to address this.

war-in (approved):

cptartur (approved):

PR #879, opened 2023-03-17
Change contract caching

### LINKED ISSUE #875: Investigate contract cache invalidation on CI on starknet update
### Feature Request

.

drknzz:
Closes #875 

## Introduced changes

- Add `poetry.lock` to hash


.github/workflows/checks.yml
```diff
@@ -93,9 +93,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: starknet_py/tests/e2e/mock/contracts_compiled
-          key: ${{ runner.os }}-contracts-${{ hashFiles('starknet_py/tests/e2e/mock/contracts') }}
-          restore-keys: |
-            ${{ runner.os }}-contracts-
+          key: ${{ runner.os }}-contracts-${{ hashFiles('starknet_py/tests/e2e/mock/contracts', 'poetry.lock') }}
```
cptartur, line 96: Quite clever idea, this will increase the checks runtime every re-lock but I think it should be fine.

drknzz: Right, but re-locks are not that common, especially in separate PRs 🚀

PR #913, opened 2023-03-30
Create parser for `Abi V1`

### LINKED ISSUE #825: Support Cairo1 in `Contract`
### Feature Request

They should work like cairo0 (current) contracts.

war-in:
Partially closes #825 

## Introduced changes

Important:
- abi.ebnf
- parser_transformer.py
- test_abi_parser.py

##


starknet_py/abi/v1/parser.py
cptartur, line 1: It's quite complicated to follow what parts of this were copied over from current abi parser and what are the new changes. Can you somehow indicate that?

starknet_py/cairo/data_types.py
```diff
@@ -55,3 +55,14 @@ class StructType(CairoType):
...
+
+@dataclass
+class EnumType(CairoType):
+    name: str
+    variants: OrderedDict[str, CairoType]
+
+
+@dataclass
+class Option(CairoType):
+    value: Optional[CairoType] = None
```
cptartur, line 68: Can we add docstrings to these?

starknet_py/abi/v1/abi.ebnf
```diff
@@ -0,0 +1,21 @@
...
+    | type_unit
+
+core_value: "felt" "252"                                        -> type_felt
+          | "integer" "::" "u" INT                              -> type_uint
+          | "option" "::" "Option" "::" "<" (type | struct) ">" -> type_option
+          | "array" "::" "Array" "::" "<" (type | struct) ">"   -> type_array
+
+type_unit: "()"
+
+struct: (IDENTIFIER "::")+ "<" type ">"
```
cptartur, line 16: Not sure abut this, why is struct defined like that? 

Can't we have a method that accept for example just this `account::account::Call`? How would that be matched?

starknet_py/abi/v1/abi.ebnf
```diff
@@ -0,0 +1,21 @@
+IDENTIFIER: /[a-zA-Z_][a-zA-Z_0-9]*/
+
+start: type
+     | struct
+
+type: "core" "::" core_value
+    | type_unit
+
+core_value: "felt" "252"                                        -> type_felt
```
cptartur, line 9: I'd just match "felt252" here

starknet_py/abi/v1/abi.ebnf
```diff
@@ -0,0 +1,21 @@
...
+          | "integer" "::" "u" INT                              -> type_uint
```
cptartur, line 10: Shouldn't this somehow distinguish between uints like 128, 256 etc?

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,65 @@
...
+        return value
+
+    def type(self, value):
+        return value[0]
+
+    def type_felt(self, _value):
+        return FeltType()
+
+    def type_uint(self, _value):
+        return FeltType()
```
cptartur, line 30: I think treating uint as felt is not a good approach. Larger uints will most likely need to be represented as multiple felts.

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,65 @@
...
+        return None
+
+    def type_option(self, _value):
+        return Option()
+
+    def type_array(self, value):
+        return ArrayType(value[0])
+
+    def struct(self, tokens):
+        return "::".join(token for token in tokens if isinstance(token, str))
```
cptartur, line 42: Can you explain this?

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __default__(self, data: str, children, meta):
+        raise TypeError(f"Unable to parse tree node of type {data}")
+
+    @v_args(inline=True)
+    def start(self, value):
+        return value
+
+    def type(self, value):
+        return value[0]
```
cptartur, line 24: Why index 0?

starknet_py/abi/v1/schemas.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    state_mutability = fields.String(data_key="state_mutability", default=None)
+
+
+class FunctionAbiEntrySchema(FunctionBaseSchema):
+    type = fields.Constant(FUNCTION_ENTRY, data_key="type", required=True)
+
+
+class ConstructorAbiEntrySchema(FunctionBaseSchema):
+    type = fields.Constant(CONSTRUCTOR_ENTRY, data_key="type", required=True)
```
cptartur, line 38: What's the purpose of adding specific type to the constructor function?

starknet_py/cairo/v1/type_parser.py
cptartur, line 1: What are the changes compared to existing type parser?

starknet_py/abi/v1/abi.ebnf
war-in: It won't. I'm adding this based on the examples, so once structs will be included correctly (I believe it just happened) I'll modify them

starknet_py/abi/v1/abi.ebnf
cptartur: I think the type of uint will definitely influence the calldata we are sending so distinguishing between them will be necessary.

starknet_py/abi/v1/parser_transformer.py
war-in: Because value is a list of types

war-in: and it's length is 1 here

starknet_py/abi/v1/schemas.py
war-in: There is no. Leftover from the old parser

starknet_py/cairo/v1/type_parser.py
```diff
@@ -0,0 +1,59 @@
...
+        Inline type is one that can be used inline, for instance as return type. For instance
+        (a: Uint256, b: felt*, c: (felt, felt)). Structure can only be referenced in inline type, can't be defined
+        this way.
+
+        :param type_string: type to parse.
+        """
+        parsed = parse(type_string)
+
+        if isinstance(parsed, str):
+            return self._get_struct(parsed)
```
drknzz, line 52: Doesn't parse return CairoType? If so, why can the result be a string?

starknet_py/tests/e2e/fixtures/abi_v1_structures.py
```diff
@@ -0,0 +1,75 @@
...
+
+my_enum_dict = {
+    "type": "enum",
+    "name": "test::MyEnum::<core::integer::u128>",
+    "variants": [
+        {"name": "a", "type": "core::integer::u256"},
+        {"name": "b", "type": "test::MyStruct::<core::integer::u128>"},
+    ],
+}
+my_enum = StructType(
```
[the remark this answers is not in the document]
war-in: They are built the same way as structs.
Also, it is much easier to keep them as structs since structs and enums can have each other inside (the logic for that is already implemented in the structure type).

starknet_py/cairo/v1/type_parser.py
war-in: Because parsing struct returns it's name

war-in: Maybe I should rethink that approach :hmm:

starknet_py/abi/v1/parser.py
```diff
@@ -0,0 +1,197 @@
...
+            raise AbiParsingError(err) from ValueError
+
+    def _parse_function(self, function: FunctionDict) -> Abi.Function:
+        return Abi.Function(
+            name=function["name"],
+            inputs=self._parse_members(function["inputs"], function["name"]),
+            outputs=list(
+                self.type_parser.parse_inline_type(param["type"])
+                for param in function["outputs"]
+            ),
```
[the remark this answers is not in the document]
war-in: Because `_parse_members` returns OrderedDict and outputs needs a list of CairoTypes.
Returned values are not named

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def type_felt(self, _value):
+        return FeltType()
+
+    def type_uint(self, value):
+        return UintType(int(value[0]))
```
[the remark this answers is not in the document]
war-in: It could but it shouldn't.
I'll add docstrings here to clarify

starknet_py/abi/v1/parser_transformer.py
cptartur: Okay, don't we only want to take the second value that is `"test"`, "MyStruct"` etc. and not the whole token?

starknet_py/abi/v1/parser_transformer.py
cptartur: Maybe let's explicitly check value is of length 1? To make it extra clear.

starknet_py/abi/v1/abi.ebnf
```diff
@@ -2,18 +2,22 @@ IDENTIFIER: /[a-zA-Z_][a-zA-Z_0-9]*/
...
+core_value: "felt252"                                        -> type_felt
           | "integer" "::" "u" INT                              -> type_uint
           | "option" "::" "Option" "::" "<" (type | struct) ">" -> type_option
           | "array" "::" "Array" "::" "<" (type | struct) ">"   -> type_array
+          | "starknet" "::" "contract_address" "::" "ContractAddress" -> type_address
 
 type_unit: "()"
 
-struct: (IDENTIFIER "::")+ "<" type ">"
+struct: (IDENTIFIER | "::")+ ("<" type ">")*
```
cptartur, line 18: In what case can the `("<" type ">")` occur multiple times?

starknet_py/abi/v1/model.py
```diff
@@ -35,10 +35,4 @@ class Event:
         str, StructType
     ]  #: Abi of structures and enums defined by the class.
     functions: Dict[str, Function]  #: Functions defined by the class.
-    constructor: Optional[
-        Function
-    ]  #: Contract's constructor. It is None if class doesn't define one.
-    l1_handler: Optional[
-        Function
-    ]  #: Handler of L1 messages. It is None if class doesn't define one.
```
cptartur: Why was this removed?

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -26,20 +34,27 @@ def type(self, value):
...
-        return FeltType()
+    def type_uint(self, value):
+        return UintType(int(value[0]))
 
     def type_unit(self, _value):
         return None
 
     def type_option(self, _value):
-        return Option()
+        return Option(_value[0])
```
cptartur, line 44: What do you think about naming it `OptionType`? I know it's an enum technically. 

cc @drknzz

starknet_py/cairo/data_types.py
```diff
@@ -59,10 +59,41 @@ class StructType(CairoType):
...
+    """
+    Type representation of Cairo unsigned integers.
+    """
+
+    bits: int
+
+    def check_range(self, value: int):
+        """
+        Utility method checking if the `value` is in range.
+        """
```
cptartur, line 90: How about distinguishing between different types of cairo enums? I think it might be necessary in case of converting python -> starknet callldata.

starknet_py/tests/e2e/fixtures/abi_v1_structures.py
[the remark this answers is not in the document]
cptartur: So why are we even defining `EnumType` in the first place? I'd not mix the logic for these two even if they are similiar.

starknet_py/abi/v1/parser_transformer.py
war-in: If we join a token it will be represented as a string (the second value) automatically

starknet_py/abi/v1/parser_transformer.py
war-in: I don't think `abi.ebnf` lets the value be longer than 1 here, but I can check it for extra clarity

starknet_py/abi/v1/model.py
war-in: There is no more `constructor` or `l1_handler` type.
They're all functions

starknet_py/abi/v1/parser_transformer.py
cptartur: https://github.com/starkware-libs/cairo/blob/v1.0.0-alpha.6/corelib/src/option.cairo#L3

starknet_py/abi/v1/parser.py
```diff
@@ -0,0 +1,212 @@
...
+        abi = [
+            ContractAbiEntrySchema().load(entry, unknown=EXCLUDE) for entry in abi_list
+        ]
+        grouped = defaultdict(list)
+        for entry in abi:
+            grouped[entry["type"]].append(entry)
+
+        self._grouped = grouped
+
+    def parse(self) -> Abi:
```
war-in, line 56: This method was changed.

I've added enums to the `Abi` class

starknet_py/abi/v1/parser.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+    @property
+    def type_parser(self) -> TypeParser:
+        if self._type_parser:
+            return self._type_parser
+
+        raise RuntimeError("Tried to get type_parser before it was set.")
+
+    def _parse_structures_and_enums(
```
war-in, line 96: The same was done here.

starknet_py/cairo/v1/type_parser.py
```diff
@@ -0,0 +1,59 @@
...
+
+        :param type_string: type to parse.
+        """
+        parsed = parse(type_string)
+
+        if isinstance(parsed, TypeIdentifier):
+            return self._get_struct(parsed)
+        return parsed
+
+    def _get_struct(self, identifier: TypeIdentifier):
```
war-in, line 55: Changes: `_get_struct` searches for string in the struct name

starknet_py/abi/v1/parser.py
```diff
@@ -92,20 +93,24 @@ def type_parser(self) -> TypeParser:
 
         raise RuntimeError("Tried to get type_parser before it was set.")
 
-    def _parse_structures(self) -> Dict[str, StructType]:
+    def _parse_structures_and_enums(
```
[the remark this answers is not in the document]
war-in: It would be hard, because I need all of them in the TypeParser (struct can contain enum)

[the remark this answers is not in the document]
war-in: but I'll try to simplify the flow

starknet_py/cairo/v1/type_parser.py
cptartur: This definitely should be handled somehow.

starknet_py/abi/v1/parser.py
```diff
@@ -0,0 +1,212 @@
...
+        for name, struct in structs.items():
+            members = self._parse_members(
+                cast(List[TypedParameterDict], struct_members[name]),
+                f"members of structure '{name}'",
+            )
+            struct.types.update(members)
+        for name, enum in enums.items():
+            members = self._parse_members(
+                cast(List[TypedParameterDict], enum_members[name]),
+                f"members of structure '{name}'",
```
drknzz, line 143:
```suggestion
                f"members of enum '{name}'",
```
should it be?

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,120 @@
...
+        Method all the entries starts from.
+        """
+        return value
+
+    def type(self, value: List[Optional[CairoType]]) -> Optional[CairoType]:
+        """
+        Types starting with "core" or unit type "()".
+
+        Tokens are read bottom-up, so here all of them are parsed and should be just returned.
+        `Optional` is added in case of the unit type.
```
drknzz, line 40: Could you clarify in the docstring this sentence? What's the connection here between Optional and unit type?

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        assert len(value) == 1
+        return value[0]
+
+    def type_felt(self, _value: List[Any]) -> FeltType:
+        """
+        Felt does not contain any additional arguments, so `_value` is just an empty list.
```
drknzz, line 47: :+1:

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def type_uint(self, value: List[Token]) -> UintType:
+        """
+        Uint type contains information about it's size. It is present in the value[0].
+        """
+        return UintType(int(value[0]))
+
+    def type_unit(self, _value: List[Any]) -> None:
+        """
+        () type.
```
drknzz, line 59: So just empty type or?

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def type_array(self, value: List[CairoType]) -> ArrayType:
+        """
+        Array contains values of type under `value[0]`.
+        """
+        return ArrayType(value[0])
+
+    def type_identifier(self, tokens: List[Token]) -> TypeIdentifier:
+        """
+        Structs and enums are defined as follows: (IDENTIFIER | "::")+ ("<" type ">")*
```
drknzz, line 78: Is it true with the current ebnf?
```
type_identifier: (IDENTIFIER | "::")+ ("<" (type | ",")+ ">")?
```

starknet_py/cairo/v1/type_parser.py
drknzz: Regex? 👀

starknet_py/abi/v1/parser_transformer.py
drknzz: I think I understood after reading the rest of docstrings, though not so obvious at first

war-in: I meant `Optional` in typing. `unit` type is returned as `None`

starknet_py/abi/v1/parser_transformer.py
war-in: I think the unit type seems to be `void` in the cairo1.0

starknet_py/abi/v1/parser_transformer.py
war-in: Nope 😢 will have to change the docstring

starknet_py/cairo/v1/type_parser.py
```diff
@@ -0,0 +1,59 @@
...
+            if name != struct.name:
+                raise ValueError(
+                    f"Keys must match name of type, '{name}' != '{struct.name}'."
+                )
+
+    def parse_inline_type(self, type_string: str) -> CairoType:
+        """
+        Inline type is one that can be used inline, for instance as return type. For instance
+        (a: Uint256, b: felt*, c: (felt, felt)). Structure can only be referenced in inline type, can't be defined
+        this way.
```
cptartur, line 45: I think examples in this and other docstring are not valid cairo1 syntax (and were valid in cairo0). This should be updated.

starknet_py/abi/v1/abi.ebnf
```diff
@@ -0,0 +1,25 @@
...
+
+type_unit: "()"
+
+type_identifier: (IDENTIFIER | "::")+ ("<" (type | ",")+ ">")?
+
+tuple: "(" (start | ",")+ ")"
+
+
+%import common.INT
+%import common.WS
```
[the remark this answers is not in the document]
war-in: INT is used to read the `integer` field. it automatically decodes `128, 256, ...` values
WS means whitespace and the next line ignores them (we don't need whitespace)

starknet_py/abi/v1/parser_transformer.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def type_identifier(self, tokens: List[Token]) -> TypeIdentifier:
+        """
+        Structs and enums are defined as follows: (IDENTIFIER | "::")+ [some not important info]
+        where IDENTIFIER is a string.
+
+        Tokens would contain strings and types (if it is present).
+        We are interested only in the strings because a structure (or enum) name can be built from them.
+        """
+        name = "::".join(token for token in tokens if isinstance(token, str)) + "::"
```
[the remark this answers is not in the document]
war-in: It is intentional because it resolves the problem with `test::MyStruct` and `test::MyStruct2`.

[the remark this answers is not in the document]
cptartur: I'll trust you with this one.

starknet_py/cairo/v1/type_parser.py
```diff
@@ -0,0 +1,59 @@
...
+        TypeParser constructor.
+
+        :param defined_types: dictionary containing all defined types. For now, they can only be structures.
+        """
+        self.defined_types = defined_types
+        for name, defined_type in defined_types.items():
+            if name != defined_type.name:
+                raise ValueError(
+                    f"Keys must match name of type, '{name}' != '{defined_type.name}'."
+                )
```
cptartur, line 39: Wasn't this supposed to be renamed?

war-in: which `this`? I've renamed `structure` -> `defined_type`

cptartur: Okay, right. It's fine.

cptartur (approved):
Do we want to target the development branch already?

PR #918, opened 2023-03-31
Remove `enhancement` tag

drknzz:


war-in:
Any reason why?

cptartur:
> Any reason why?

I think it's because 90% of the issues gets the `enhancement` because of that and it stops being useful.

cptartur:
But I wouldn't remove this tag entirely from possible tags. Only from the template.

drknzz:
> But I wouldn't remove this tag entirely from possible tags. Only from the template.

Yes, the tag is still available.

cptartur (approved):

PR #947, opened 2023-04-11
Remove usage of starknet `pedersen_hash`

tomek0123456789:
## Introduced changes

- removed usage of starknet implementation of `pedersen_hash`

##

- `starknet.py` now only uses `crypto-cpp-py` implementation of pedersen hash


poetry.lock
war-in, line 1: Why poetry.lock has changed without changes in pyproject.toml?

starknet_py/hash/utils.py
```diff
@@ -23,9 +21,7 @@ def pedersen_hash(left: int, right: int) -> int:
     """
     One of two hash functions (along with _starknet_keccak) used throughout StarkNet.
     """
-    if use_cpp_variant():
-        return cpp_hash(left, right)
```
cptartur, line 27: `use_cpp_variant` function should be removed entirely, as I'm assuming it's no longer used anywhere.

cptartur (commented):
This may be a big breaking change, though I'm not sure. I'd definitely at least include this in the migration guide once we release.

starknet_py/hash/utils.py
```diff
@@ -3,11 +3,9 @@
 
 from crypto_cpp_py.cpp_bindings import ECSignature, cpp_hash
 from eth_utils.crypto import keccak
-from starkware.cairo.lang.vm.crypto import pedersen_hash as default_hash
 from starkware.crypto.signature.signature import sign
 
 from starknet_py.common import int_from_bytes
-from starknet_py.utils.crypto.facade import use_cpp_variant
```
war-in, line 10: If we decide to remove usages of `default_hash` we should do the following:
- modify the docs (I remember there is something about disabling cpp extension)
- remove `use_cpp_variant`

starknet_py/hash/utils.py
```diff
@@ -23,9 +21,7 @@ def pedersen_hash(left: int, right: int) -> int:
     """
     One of two hash functions (along with _starknet_keccak) used throughout StarkNet.
     """
-    if use_cpp_variant():
-        return cpp_hash(left, right)
-    return default_hash(left, right)
+    return cpp_hash(left, right)
```
war-in, line 24: Shouldn't we deprecate it first?

cc @cptartur @drknzz

war-in: Totally agree. Let's deprecate it and necessarily write about it in the docs

starknet_py/hash/utils.py
tomek0123456789: Left it, because I added deprecation to the pedersen hash.

starknet_py/utils/crypto/crypto_test.py
```diff
@@ -2,24 +2,14 @@
...
     msg_hash, priv_key = args
-    monkeypatch.setenv("DISABLE_CRYPTO_C_EXTENSION", "")
-    assert use_cpp_variant()
-    signature_1 = message_signature(*args)
-
-    monkeypatch.setenv("DISABLE_CRYPTO_C_EXTENSION", "true")
-    assert not use_cpp_variant()
-    signature_2 = message_signature(*args)
-
-    assert signature_1 == signature_2
```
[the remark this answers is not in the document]
cptartur: Let's keep these then.

docs/migration_guide.rst
```diff
@@ -1,6 +1,16 @@
...
 ===============
 
+**********************
+0.16.0 Migration guide
+**********************
+
+Deprecations
+------------
+
+- Python implementation of pedersen hash in :meth:`~starknet_py.hash.utils.pedersen_hash` function is removed in favour of `crypto-cpp-py` library implementation.
```
war-in, line 11: It's not removed 😅 The implementation has changed and it always uses crypto-cpp

war-in (approved):

PR #949, opened 2023-04-12
Wrap missing pythonic hints error

### LINKED ISSUE #948: Mention `--add-pythonic-hints` in docs
### Feature Request

It's not mentioned and users are confused

drknzz:
Closes #948


starknet_py/common.py
```diff
@@ -97,7 +97,15 @@ def create_casm_class(compiled_contract: str) -> CasmClass:
     :param compiled_contract: contract compiled using starknet-sierra-compile.
     :return: CasmClass instance.
     """
-    return cast(CasmClass, CasmClassSchema().loads(compiled_contract))
+    try:
+        return cast(CasmClass, CasmClassSchema().loads(compiled_contract))
+    except ValidationError as err:
+        if err.messages == {"pythonic_hints": ["Missing data for required field."]}:
```
war-in, line 103: How about adding `Error: Libfunc alloc_local is not allowed in the libfuncs list` too?

drknzz: I think it's not necessary, as this is included in the error message:
```
Run with '--allowed-libfuncs-list-name experimental_v0.1.0' to allow all libfuncs.
```

war-in (approved):

PR #950, opened 2023-04-12
Replace sign, verify, private_to_stark_key

### LINKED ISSUE #930: Implement `private_to_stark_key`, `sign`, `verify`
### Feature Request

Functions are from `starkware.crypto.signature.signature`.

Used in `StarkCurveSigner`, message signatures.

drknzz:
Closes #930

## Introduced changes

- use verify from crypto-cpp-py instead of account calling

##


starknet_py/net/account/base_account.py
```diff
@@ -188,7 +188,7 @@ def sign_message(self, typed_data: TypedData) -> List[int]:
         """
 
     @abstractmethod
-    async def verify_message(self, typed_data: TypedData, signature: List[int]) -> bool:
+    def verify_message(self, typed_data: TypedData, signature: List[int]) -> bool:
```
drknzz, line 191: This is a breaking change, needs a mention in docs

docs/migration_guide.rst
```diff
@@ -8,7 +8,18 @@ Migration guide
 Breaking changes
 ----------------
 
-- :meth:`~starknet_py.hash.utils.pedersen_hash` function implementation has been changed, now uses only `crypto-cpp-py` library implementation.
+.. currentmodule:: starknet_py.net.account.base_account
```
war-in, line 11: I'm not sure if we should use it here. This could lead to hard-to-find problems in the future. 
wdyt @cptartur?

war-in (approved):

docs/migration_guide.rst
drknzz: Migration guide file is written from the bottom-up, so I think there shouldn't be any problems.

war-in: Ok, fair, but what are the advantages of this approach?

drknzz: BaseAccount.verify_message is linked to api section

war-in: And we could do `starknet_py.net.account.base_account.BaseAccount.verify_message` as always

drknzz: Then it's either the long path with all the package names you listed or just `verify_message` if you precede the path with a tilde `~`.

Current solution leaves only `BaseAccount.verify_message` with a reference.

PR #959, opened 2023-04-13
Add note about cairo1 contracts being strings in docs

cptartur:
Closes #

## Introduced changes

- Update comments in "Declaring Cairo1 contracts" to indicate contracts are passed as strings. Current description could suggest that they were paths.

##


drknzz (commented):
Maybe we should just create some `read_contract` function in public api?
Then we could use it in the test.

starknet_py/tests/e2e/docs/guide/test_cairo1_contract.py
```diff
@@ -26,7 +26,7 @@ async def test_cairo1_contract(
     from starknet_py.common import create_casm_class
     from starknet_py.hash.casm_class_hash import compute_casm_class_hash
 
-    # contract_compiled_casm is the output of the starknet-sierra-compile (.casm file)
+    # contract_compiled_casm is a string with the output of the starknet-sierra-compile (.casm file)
```
drknzz, line 29:
```suggestion
    # contract_compiled_casm is the content of the starknet-sierra-compile output (.casm file)
```
Would something like this be clearer maybe?

cptartur:
> Maybe we should just create some `read_contract` function in public api? Then we could use it in the test.

Pathlib already provides `.read_text` which is simple enough. Don't see any benefit exposing something like that on our side.

starknet_py/tests/e2e/docs/guide/test_cairo1_contract.py
cptartur: Sounds nicer, but I wanted to put emphasis on the fact that it is a `string` so someone doesn't try to use a path or something else.

drknzz: > > Maybe we should just create some `read_contract` function in public api? Then we could use it in the test.
> 
> Pathlib already provides `.read_text` which is simple enough. Don't see any benefit exposing something like that on our side.

Then I think we should just use the function from pathlib directly in the test (and make it visible in the docs of course).

drknzz:
> > Maybe we should just create some `read_contract` function in public api? Then we could use it in the test.
> 
> Pathlib already provides `.read_text` which is simple enough. Don't see any benefit exposing something like that on our side.

Then I think we should just use the function from pathlib directly in the test (and make it visible in the docs of course).

cptartur:
> > > Maybe we should just create some `read_contract` function in public api? Then we could use it in the test.
> > 
> > 
> > Pathlib already provides `.read_text` which is simple enough. Don't see any benefit exposing something like that on our side.
> 
> Then I think we should just use the function from pathlib directly in the test (and make it visible in the docs of course).

So to have something like

```py
contract_compiled_casm = Path(contract_compiled_path).read_text("utf-8")
casm_class = create_casm_class()
```

I'm worried this will reduce clarity of tests.

drknzz:
> > > > Maybe we should just create some `read_contract` function in public api? Then we could use it in the test.
> > > 
> > > 
> > > Pathlib already provides `.read_text` which is simple enough. Don't see any benefit exposing something like that on our side.
> > 
> > 
> > Then I think we should just use the function from pathlib directly in the test (and make it visible in the docs of course).
> 
> So to have something like
> 
> ```python
> contract_compiled_casm = Path(contract_compiled_path).read_text("utf-8")
> casm_class = create_casm_class()
> ```
> 
> I'm worried this will reduce clarity of tests.

Then let's put a new line there 👀

cptartur:
> > > > > Maybe we should just create some `read_contract` function in public api? Then we could use it in the test.
> > > > 
> > > > 
> > > > Pathlib already provides `.read_text` which is simple enough. Don't see any benefit exposing something like that on our side.
> > > 
> > > 
> > > Then I think we should just use the function from pathlib directly in the test (and make it visible in the docs of course).
> > 
> > 
> > So to have something like
> > ```python
> > contract_compiled_casm = Path(contract_compiled_path).read_text("utf-8")
> > casm_class = create_casm_class()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm worried this will reduce clarity of tests.
> 
> Then let's put a new line there 👀

👁️ 👁️  Not sure that will help in this case

starknet_py/tests/e2e/docs/guide/test_cairo1_contract.py
war-in, line 29:
```suggestion
    # contract_compiled_casm is a string containing the content of the starknet-sierra-compile (.casm file)
```

drknzz (approved):

war-in (approved):

starknet_py/tests/e2e/docs/guide/test_cairo1_contract.py
```diff
@@ -39,7 +39,7 @@ async def test_cairo1_contract(
 
     # Create Declare v2 transaction
     declare_v2_transaction = await account.sign_declare_v2_transaction(
-        # compiled_contract is the output of the starknet-compile (.json file)
+        # contract_compiled is a string containing the content of the starknet-compile (.json file)
```
drknzz, line 42:
```suggestion
        # compiled_contract is a string containing the content of the starknet-compile (.json file)
```

drknzz (approved):
👍 👍 👍 👍

PR #970, opened 2023-04-19
Remove requirements.txt

### LINKED ISSUE #941: Remove optional dependencies from requirements.txt
### Feature Request

When installing with pip all extras are downloaded. We should remove optional dependencies from requirements.txt and maybe create an optional_requirements.txt file for extras.

war-in:
Closes #941 

# Please update your hooks after this PR is merged

## Introduced changes

- `docs-requirements.txt` was added
- optional dependencies are no longer present in `requirements.txt`
- unused `sphinx-rtd-theme` was removed

##


.github/workflows/checks.yml
```diff
@@ -67,9 +67,10 @@ jobs:
       - name: Typechecking
         run: |
           poetry run poe typecheck
-      - name: Check if requirements.txt is up-to-date
+      - name: Check if requirements.txt and docs-requirements.txt are up-to-date
         run: |
-          poetry export -f requirements.txt --without-hashes --extras docs | cmp - requirements.txt
+          poetry export -f requirements.txt --without-hashes | cmp - requirements.txt
+          poetry export -f requirements.txt --without-hashes --extras docs | cmp - docs-requirements.txt
```
drknzz, line 73: Is `docs-requirements` okay? Isn't `optional-requirements` kind of conventional?

.readthedocs.yml
```diff
@@ -13,6 +13,6 @@ sphinx:
 
 python:
   install:
-    - requirements: requirements.txt
+    - requirements: docs-requirements.txt
```
drknzz, line 16: Can `starknet.py` be installed with all the optional dependencies from pypi?

drknzz (commented):
112 -> 68 lines in requirements.txt 👌

.readthedocs.yml
cptartur: Is there any point of installing starknet.py from pypi that way? I

cptartur (commented):
I started to wonder if there's any point of having `requirements` defined at all. I think for pypi dependencies are somehow defined in the setup.py or other file - they were just there for docs I believe.

.readthedocs.yml
drknzz: I don't know, just curious though. 😄

.github/workflows/checks.yml
war-in: In our case, those requirements are only for the docs, but to be more future proof I can change the name to `optional`

war-in:
> I started to wonder if there's any point of having requirements defined at all. I think for pypi dependencies are somehow defined in the setup.py or other file - they were just there for docs I believe.

Unfortunately, I don't think there is an easy way to do so.
I've tried to upload the package without requirements.txt to test-pypi

https://test.pypi.org/project/starknet-py/

And I can't install it :/

cptartur:
> > I started to wonder if there's any point of having requirements defined at all. I think for pypi dependencies are somehow defined in the setup.py or other file - they were just there for docs I believe.
> 
> Unfortunately, I don't think there is an easy way to do so. I've tried to upload the package without requirements.txt to test-pypi
> 
> https://test.pypi.org/project/starknet-py/
> 
> And I can't install it :/

If they are necessary, let's keep them.

PR #977, opened 2023-04-20
Make cairo-lang dev dependency

drknzz:


starknet_py/hash/casm_class_hash_test.py
```diff
@@ -1,10 +1,5 @@
+# fmt: off
```
war-in, line 1: Why? 😢

starknet_py/hash/transaction_test.py
```diff
@@ -21,7 +14,7 @@
     compute_transaction_hash,
 )
 from starknet_py.net.models import StarknetChainId
-from starknet_py.net.schemas.gateway import SierraCompiledContractSchema
+from starknet_py.tests.e2e.fixtures.constants import CONTRACTS_COMPILED_V1_DIR
```
war-in, line 17: Nope, not here 🙏

starknet_py/hash/casm_class_hash_test.py
war-in: Couldn't it be just above @parametrize?
And below

requirements.txt
war-in, line 1: 👀

PR #983, opened 2023-04-24
Adjust serializers

war-in:
Closes #

## Introduced changes

- Interacting with V1 Contracts 🎉 
- `cairo_version` argument in some methods to enable V1 parsing
- Serializing uints

##


starknet_py/tests/e2e/fixtures/devnet.py
```diff
@@ -47,6 +47,8 @@ def start_devnet():
         str(1),
         "--seed",  # generates same accounts each time
         str(1),
+        "--timeout",
+        "1000",
```
tomek0123456789, line 51: Why is such a long timeout needed?

starknet_py/serialization/data_serializers/output_serializer.py
```diff
@@ -0,0 +1,27 @@
...
+        for serializer in self.serializers:
+            with context.push_entity("output"):
+                result.append(serializer.deserialize_with_context(context))
+
+        return tuple(result)
+
+    def serialize_with_context(
+        self, context: SerializationContext, value: Dict
+    ) -> Generator[int, None, None]:
+        yield 0
```
drknzz, line 27: Any explanation about this?

starknet_py/serialization/data_serializers/uint_serializer.py
```diff
@@ -0,0 +1,75 @@
...
+
+from starknet_py.cairo.felt import uint256_range_check
+from starknet_py.serialization import CairoDataSerializer
+from starknet_py.serialization._context import (
+    Context,
+    DeserializationContext,
+    SerializationContext,
+)
+
+U128_UPPER_BOUND = 2**128
```
drknzz, line 12: Here it does not matter but you could always do `1 << 128` which is much faster. (again here it does not matter 🗿)

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
```diff
@@ -0,0 +1,58 @@
+import pytest
+
+from starknet_py.cairo.felt import decode_shortstring, encode_shortstring
+from starknet_py.tests.e2e.fixtures.constants import MAX_FEE
+from starknet_py.tests.e2e.fixtures.contracts import deploy_v1_contract
+
+
+@pytest.mark.asyncio
+async def test_general_v1_interaction(gateway_account):
```
drknzz, line 9: A concurrent PR with replacing gateway with full node in tests exists, so it's something that probably would have to be taken care of before merging into development (for now, in the tracking branch, it does not matter really)

starknet_py/tests/e2e/fixtures/devnet.py
drknzz: I suppose it's from local testing. I think we could get rid of it. Is this change necessary?

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
```diff
@@ -0,0 +1,58 @@
...
+        account=gateway_account,
+        contract_file_name="token_bridge",
+        calldata={"governor_address": gateway_account.address},
+    )
+
+    await (
+        await bridge.functions["set_l1_bridge"].invoke(
+            l1_bridge_address={"address": 0x11}, max_fee=MAX_FEE
+        )
+    ).wait_for_acceptance()
```
drknzz, line 58: Could the test also check the `l1_bridge` after?

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
```diff
@@ -0,0 +1,58 @@
...
+        await erc20.functions["transfer"].invoke(
+            recipient=0x11, amount=10, max_fee=MAX_FEE
+        )
+    ).wait_for_acceptance()
+
+    (after_transfer_balance,) = await erc20.functions["balance_of"].call(
+        account=gateway_account.address
+    )
+
+    assert "erc20_basic" in decoded_name
```
drknzz, line 39: Could it be `==`?

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
```diff
@@ -0,0 +1,58 @@
...
+    assert decimals == 10
+    assert supply == 12345
+    assert account_balance == 12345
+    assert after_transfer_balance == 12345 - 10
```
drknzz, line 43: You could create a calldata dict, pass it to `deploy_v1_contract` and later make asserts with `== calldata.param`

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
```diff
@@ -0,0 +1,58 @@
...
+        calldata={
+            "name_": encode_shortstring("erc20_basic"),
+            "symbol_": encode_shortstring("ERC20B"),
+            "decimals_": 10,
+            "initial_supply": 12345,
+            "recipient": gateway_account.address,
+        },
+    )
+
+    decoded_name = decode_shortstring((await erc20.functions["get_name"].call())[0])
```
drknzz, line 22:
```suggestion
    (name,) = await erc20.functions["get_name"].call()
    decoded_name = decode_shortstring(name)
```

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
war-in: I would have to add a custom function to `token_bridge` :/

starknet_py/serialization/data_serializers/uint_serializer.py
cptartur: I might be confused: should uint128 upper bound be just `2 ** 128` or `(2 ** 128) - 1`?

starknet_py/serialization/data_serializers/uint_serializer.py
```diff
@@ -0,0 +1,75 @@
...
+
+        [low, high] = context.reader.read(2)
+
+        # Checking if resulting value is in [0, 2**256) range is not enough. Uint256 should be made of two uint128.
+        with context.push_entity("low"):
+            self._ensure_valid_uint128(low, context)
+        with context.push_entity("high"):
+            self._ensure_valid_uint128(high, context)
+
+        return (high << 128) + low
```
cptartur, line 40: What in the case someone uses other kind of uint (are these possible in cairo1)?

starknet_py/serialization/data_serializers/uint_serializer.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def serialize_with_context(
+        self, context: SerializationContext, value: Union[int, Uint256Dict]
+    ) -> Generator[int, None, None]:
+        context.ensure_valid_type(value, isinstance(value, (int, dict)), "int or dict")
+        if isinstance(value, int):
+            yield from self._serialize_from_int(value, self.bits)
+        else:
+            yield from self._serialize_from_dict(context, value)
```
cptartur, line 49: Is it only possible for this branch to execute if uint is 256 bit?

starknet_py/serialization/function_serialization_adapter.py
```diff
@@ -90,3 +93,16 @@ def _ensure_no_missing_args(expected_args: Set[str], provided_args: Set[str]):
...
+class FunctionSerializationAdapterV1(FunctionSerializationAdapter):
+    outputs_deserializer: OutputSerializer
+
+    def deserialize(self, data: List[int]) -> Tuple:
+        """
+        Deserializes data into TupleDataclass containing python representations.
+
+        :return: cairo data.
+        """
+        return self.outputs_deserializer.deserialize(data)
```
cptartur, line 108: Why do we have to introduce an adapter for this?

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
drknzz: Okay, let's not then. I'd rather not modify the contracts from cairo repo ourselves, so we can adjust easily to new versions.

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
drknzz: Okay 👌 seems weird though, should it be like this?

starknet_py/utils/contructor_args_translator.py
```diff
@@ -1,17 +1,31 @@
...
+
+        serializer = serializer_for_function_v1(constructor)
+    else:
+        parsed = AbiParser(abi).parse()
+
+        # Constructor might not accept any arguments
+        if not parsed.constructor or not parsed.constructor.inputs:
+            return []
+
+        serializer = serializer_for_function(parsed.constructor)
```
cptartur, line 28: Let's split these into two helper methods.

starknet_py/tests/e2e/fixtures/contracts.py
```diff
@@ -93,6 +95,54 @@ async def deploy_contract(account: BaseAccount, class_hash: int, abi: List) -> C
     return deployment_result.deployed_contract
 
 
+async def deploy_v1_contract(
+    account: BaseAccount, contract_file_name: str, calldata: Dict[str, Any]
```
cptartur, line 99: Nitpick, but perhaps passing whole paths to this method would be a better and more explicit approach.

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
cptartur: I think you don't need parentheses around `name,`.

starknet_py/serialization/data_serializers/uint_serializer.py
drknzz: Depends on how it is used, but the name suggests it should be `-1`. Good catch, let's double check on that.

starknet_py/net/udc_deployer/deployer.py
```diff
@@ -86,6 +86,7 @@ def create_contract_deployment(
         *,
         salt: Optional[int] = None,
         abi: Optional[List] = None,
+        cairo_version: int = 0,
```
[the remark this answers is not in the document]
drknzz: I don't mind it here, but it might be better in the constructor

starknet_py/serialization/data_serializers/uint_serializer.py
war-in: It is used like this `0 <= value < U128_UPPER_BOUND`, so the value is fine

war-in: I would keep it as it is because this form is much more readable

starknet_py/net/udc_deployer/deployer.py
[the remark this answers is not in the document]
war-in: I think it might be better to keep it there because what if someone would like to deploy two contracts (0 and 1) with the same deployer?

starknet_py/serialization/data_serializers/uint_serializer.py
war-in: Yes, there are a few uints in cairo1 (u8, u32, ...).
If self.bits is lower than 256, the first part of the code will be used

starknet_py/serialization/data_serializers/uint_serializer.py
war-in: Yes

starknet_py/tests/e2e/fixtures/contracts.py
war-in: I'm not sure, since we want to keep cairo1 contracts in the one repository (although for now)

starknet_py/serialization/data_serializers/uint_serializer.py
cptartur: Perhaps using `_ensure_valid_uint128` may be a bit confusing then? Though I guess it won't matter considering how python handles integers.

starknet_py/net/udc_deployer/deployer.py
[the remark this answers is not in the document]
tomek0123456789: Good argument, I'd stay then with current implementation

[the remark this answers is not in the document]
cptartur: I think either are acceptable solution

drknzz (commented):
💯

starknet_py/serialization/data_serializers/unit_serializer.py
```diff
@@ -0,0 +1,29 @@
...
+    Serializer for unit type.
+    Can't serialize anything.
+    Deserializes data to None.
+
+    Example:
+        [] => None
+    """
+
+    def deserialize_with_context(self, context: DeserializationContext) -> None:
+        return None
```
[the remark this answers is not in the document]
war-in: context does it by himself, we do not have to check it

starknet_py/serialization/data_serializers/uint_serializer_test.py
```diff
@@ -0,0 +1,118 @@
...
+    with pytest.raises(InvalidValueException, match=low_error_message):
+        u256_serializer.deserialize([-1, 0])
+
+    high_error_message = re.escape(
+        "Error at path 'high': expected value in range [0;2**128)"
+    )
+    with pytest.raises(InvalidValueException, match=high_error_message):
+        u256_serializer.deserialize([0, MAX_U128 + 1])
+    with pytest.raises(InvalidValueException, match=high_error_message):
+        u256_serializer.deserialize([0, -1])
```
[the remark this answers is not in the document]
war-in: Nice one :o

starknet_py/serialization/data_serializers/unit_serializer_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+from starknet_py.serialization.data_serializers.unit_serializer import UnitSerializer
+
+serializer = UnitSerializer()
+
+
+def test_deserialize_unit():
+    deserialized = serializer.deserialize([])
+
+    assert deserialized is None
```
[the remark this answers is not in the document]
drknzz: What about this? Could we have a test for it?

drknzz (approved):
👌

PR #989, opened 2023-04-26
Do not test `GatewayClient` on integration

### LINKED ISSUE #974: Replace integration tests in `gateway_test.py`
Some tests in `gateway_test.py` run on integration network - they have a comment

```
# Temporary test to be replaced after devnet supports new starknet
```

war-in:
Closes #974 

## Introduced changes

- 
- 

##


starknet_py/tests/e2e/client/fixtures/transactions.py
```diff
@@ -53,3 +60,73 @@ def block_with_deploy_account_number(
...
+    await deploy_result.wait_for_acceptance()
+
+    contract = deploy_result.deployed_contract
+
+    resp = await contract.functions["replace_implementation"].invoke(
+        new_class=map_class_hash, max_fee=MAX_FEE
+    )
+    await resp.wait_for_acceptance()
+
+    return resp.block_number, contract.address, map_class_hash
```
drknzz, line 132: Are we testing anywhere the class replacement? Maybe let's add a test for the new `replace_class.cairo` to make sure it's working properly

war-in: I don't think it's necessary. The test is working properly

starknet_py/tests/e2e/client/fixtures/transactions.py
```diff
@@ -1,20 +1,27 @@
+# pylint: disable=redefined-outer-name
```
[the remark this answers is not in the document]
war-in: to the pytest_asyncio extension
image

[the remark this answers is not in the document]
drknzz: Could we add a comment / bring it closer to the issue?

starknet_py/tests/e2e/client/fixtures/transactions.py
drknzz: I thought about some test that would use the contract after replacing the class

war-in: But I'm not sure if we should. That would not test the starknet.py feature, just cairo contract

starknet_py/tests/e2e/client/fixtures/transactions.py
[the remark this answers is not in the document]
war-in: We could but it is already present in many files with fixtures
image

drknzz (approved):
👍

starknet_py/tests/e2e/client/fixtures/transactions.py
war-in: Checked it locally and it works 🎉

PR #1017, opened 2023-05-15
Allow using cairo1 in abi resolver

### LINKED ISSUE #1012: Consider better error handling for `get_abi_for_address` and `resolve_abi` methods in `ContractAbiResolver`
TODO comments:
- [no. 1](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/proxy/contract_abi_resolver.py#L97)
- [no. 2](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/proxy/contract_abi_resolver.py#L126)

war-in:
Closes #1012 

## Introduced changes

- 
- 

##


tomek0123456789 (approved):

starknet_py/proxy/contract_abi_resolver.py
```diff
@@ -85,26 +87,27 @@ async def resolve(self) -> AbiDictList:
...
         if contract_class.abi is None:
             raise AbiNotFoundError()
-        return contract_class.abi
 
-    async def resolve_abi(self) -> AbiDictList:
+        if isinstance(contract_class, SierraContractClass):
+            assert isinstance(contract_class.abi, str)
+            return json.loads(contract_class.abi), 1
+
+        return cast(AbiDictList, contract_class.abi), 0
```
drknzz, line 106: Why only cast to the type when the version is 0?

starknet_py/contract.py
```diff
@@ -557,16 +556,16 @@ async def from_address(
...
 
-        try:
-            abi = await ContractAbiResolver(
-                address=address, client=client, proxy_config=proxy_config
-            ).resolve()
-        except UnsupportedAbiError as err:
-            raise ValueError(
-                "Provided address of Cairo1 contract which is currently not supported in Contract."
-            ) from err
+        abi, cairo_version = await ContractAbiResolver(
```
drknzz, line 559: Would it be possible to not change the signature and logic, and instead write a static method in the Resolver that would return the cairo_version, given an abi?

Nothing would change then, except Contract constructor below would take 
```
ContractAbiResolver._cairo_version_from_abi(abi)
```

starknet_py/proxy/contract_abi_resolver.py
war-in: Because json.loads already returns a `List[Dict]`

PR #1060, opened 2023-06-14
Remove `compute_invoke_hash`

### LINKED ISSUE #768: Remove deprecated compute_invoke_hash
_Originally posted by @cptartur in https://github.com/software-mansion/starknet.py/pull/592#discussion_r1073292122_

drknzz:
Closes #768 

## Introduced changes

- Remove `compute_invoke_hash`
- Add `compute_invoke_transaction_hash` to `hash.transaction` module
- Refactor tests to make them more isolated from context

##

### DIFF
--- a/starknet_py/hash/transaction.py
+++ b/starknet_py/hash/transaction.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from typing import Sequence
+from typing import Optional, Sequence
 
 from starknet_py.common import int_from_bytes
 from starknet_py.constants import DEFAULT_ENTRY_POINT_SELECTOR
@@ -30,7 +30,7 @@ def compute_transaction_hash(
     calldata: Sequence[int],
     max_fee: int,
     chain_id: int,
-    additional_data: Sequence[int],
+    additional_data: Optional[Sequence[int]] = None,
 ) -> int:
     """
     Calculates the transaction hash in the Starknet network - a unique identifier of the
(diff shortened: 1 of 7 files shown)


starknet_py/hash/transaction.py
```diff
@@ -76,6 +78,38 @@ def compute_transaction_hash(
     )
 
 
+def compute_invoke_transaction_hash(
+    *,
```
tomek0123456789, line 82: Do we want this function to have keyword-only arguments? It's a bit chaotic in this file.

starknet_py/hash/transaction.py
```diff
@@ -30,7 +30,7 @@ def compute_transaction_hash(
     calldata: Sequence[int],
     max_fee: int,
     chain_id: int,
-    additional_data: Sequence[int],
+    additional_data: Optional[Sequence[int]] = None,
```
tomek0123456789, line 33: Why is it optional now?

drknzz: Just for the ease of use. If it is not passed, an empty list will be assigned (see below).

starknet_py/hash/transaction.py
drknzz: Ideally everything should be keyword only 😃 

I decided to go with keyword-only as the order of the parameters is different in many of these functions.
Starknet.js has them ordered similiarly (the order in which these arguments are, actually) in each of their functions, but unfortunately we cannot change them without making it a breaking change.
Also, keyword only functions are less error prone.

As for the file, it contains just the methods that are for computing a hash of a transaction.
Since all functions but for Invoke were present here, I don't think it should be somewhere else.

PR #1078, opened 2023-06-22
Throw an error when `FullNodeClient` uses `wait_for_tx` without `wait_for_acceptance=True`

### LINKED ISSUE #1067: Raise when `FullNodeClient.wait_for_tx` used without `wait_for_accept=True`
### Feature Request

Full nodes are not able to detect whether a transaction does not exist or its status is still `PENDING`.

`FullNodeClient.wait_for_tx` should not allow to be used without waiting for the block to be `ACCEPTED_ON_L2`.

tomek0123456789:
Closes #1067 

## Introduced changes

- as in title

##

- as in title


starknet_py/net/client.py
```diff
@@ -182,6 +184,10 @@ async def wait_for_tx(
                 await asyncio.sleep(check_interval)
         except asyncio.CancelledError as exc:
             raise TransactionNotReceivedError from exc
+        except ClientError as exc:
+            raise ClientError(
```
drknzz, line 188: Let's wrap the exception only if the tx hash hasn't been found

drknzz: Let's add a test for this

drknzz (approved):

PR #1094, opened 2023-06-28
Add missing RPC methods

### LINKED ISSUE #1093: Add missing RPC methods
### Feature Request

- `starknet_blockNumber`
- `starknet_blockHashAndNumber`
- `starknet_chainId`
- `starknet_syncing`

drknzz:
Closes #1093

## Introduced changes

- `starknet_blockNumber`
- `starknet_blockHashAndNumber`
- `starknet_chainId`
- `starknet_syncing`

##


starknet_py/net/http_client.py
```diff
@@ -85,7 +85,7 @@ async def handle_request_error(self, request: ClientResponse):
 
 
 class RpcHttpClient(HttpClient):
-    async def call(self, method_name: str, params: dict) -> dict:
```
tomek0123456789, line 88: Was it needed because of the `bool` in `syncing` method?

starknet_py/net/full_node_client.py
```diff
@@ -315,6 +319,26 @@ async def estimate_fee(
...
+    async def get_chain_id(self) -> int:
+        """Return the currently configured Starknet chain id"""
+        return await self._client.call(method_name="chainId", params={})
+
+    async def get_syncing_status(self) -> Union[bool, SyncStatus]:
+        """Returns an object about the sync status, or false if the node is not syncing"""
+        sync_status = await self._client.call(method_name="syncing", params={})
+        if isinstance(sync_status, bool):
+            return sync_status
+        return cast(SyncStatus, SyncStatusSchema().load(sync_status))
```
tomek0123456789, line 340: A test checking this line would be nice

starknet_py/net/http_client.py
drknzz: The typing was wrong, the result is not always a dict, some methods return just a number (chain id), syncing just a bool, etc.

starknet_py/net/full_node_client.py
drknzz: Devnet always return false, so can't do it there. I've tested it on testnet though.

tomek0123456789: What about a mock? 👀

PR #1102, opened 2023-07-11
Deprecate `wait_for_accept` parameter 

tomek0123456789:
## Introduced changes

- `wait_for_accept` parameter in `Client.wait_for_tx` and `Contract.wait_for_acceptance` methods is now deprecated
- added `retries` parameter

##

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,8 @@ compile_contracts_v1 = "bash starknet_py/tests/e2e/mock/compile_contracts_v1.sh"
 compile_contracts_v2 = "bash starknet_py/tests/e2e/mock/compile_contracts_v2.sh"
 circular_imports_check.shell = "poetry run pytest circular.py"
 ci = ["lint", "format_check", "typecheck", "test_ci"]
+precommit.sequence = ["format", "lint", "typecheck"]
+precommit.ignore_fail = true
 
 [tool.poetry.build]
 generate-setup-file = true
--- a/starknet_py/net/client.py
+++ b/starknet_py/net/client.py
@@ -136,7 +136,8 @@ class Client(ABC):
         self,
         tx_hash: Hash,
         wait_for_accept: Optional[bool] = None,  # pylint: disable=unused-argument
-        check_interval=5,
+        check_interval: int = 5,
+        retries: int = 200
     ) -> Tuple[int, TransactionStatus]:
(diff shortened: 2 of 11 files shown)


starknet_py/net/client.py
```diff
@@ -135,23 +135,33 @@ async def get_transaction_receipt(
...
         # pylint: disable=too-many-branches
         """
         Awaits for transaction to get accepted or at least pending by polling its status.
 
         :param tx_hash: Transaction's hash.
-        :param wait_for_accept: If true waits for at least ACCEPTED_ON_L2 status, otherwise waits for at least PENDING.
-            Defaults to false
+        :param wait_for_accept:
+            .. deprecated:: 0.17.0
+                Parameter `wait_for_accept` and `PENDING` status have been deprecated - if a transaction is accepted,
```
drknzz, line 149: Same here

starknet_py/net/client.py
```diff
@@ -135,23 +135,33 @@ async def get_transaction_receipt(
...
+        if wait_for_accept is not None:
+            warnings.warn(
+                "Parameter `wait_for_accept` and `PENDING` status have been deprecated - if a transaction is accepted, "
+                "it goes straight into ACCEPTED_ON_L2 status."
+            )
 
-        first_run = True
         try:
             while True:
                 result = await self.get_transaction_receipt(tx_hash=tx_hash)
```
drknzz, line 167: We would probably need to catch failed receipt fetch, and retry

pyproject.toml
```diff
@@ -81,6 +81,8 @@ compile_contracts_v1 = "bash starknet_py/tests/e2e/mock/compile_contracts_v1.sh"
 compile_contracts_v2 = "bash starknet_py/tests/e2e/mock/compile_contracts_v2.sh"
 circular_imports_check.shell = "poetry run pytest circular.py"
 ci = ["lint", "format_check", "typecheck", "test_ci"]
+precommit.sequence = ["format", "lint", "typecheck"]
+precommit.ignore_fail = true
```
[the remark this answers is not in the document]
tomek0123456789: Yeah, I chose the name poorly, i mostly use those three commands to check what should be changed before a commit - and if `lint` fails, `typecheck` will not be run. I don't know if you'd use it though.

drknzz (approved):

PR #1121, opened 2023-07-18
Speed up CI

### LINKED ISSUE #1106: Speed up CI
### Feature Request

Current approach runs
```
poe test_ci_full_node_v1
poe test_ci_full_node_v2
poe test_ci_gateway_v1
poe test_ci_gateway_v2
```
Refactor would be needed to avoid running some tests multiple times smarter.

drknzz:
Closes #1106 

## Introduced changes

- Split running tests into:
   - `test_local` - tests that do not require devnet
   - `test_ci`
   - `test_ci_docs`

Add logic to `read_contract` that skips tests depending on passed contract directory argument

##


starknet_py/net/signer/test_stark_curve_signer.py
```diff
@@ -57,3 +43,32 @@ def test_sign_transaction(transaction):
     assert len(signature) > 0
     assert all(isinstance(i, int) for i in signature)
     assert all(i != 0 for i in signature)
+
+
+# separate test due to contract_class using functions
```
tomek0123456789, line 48: How does this affect the CI?

starknet_py/tests/e2e/fixtures/contracts.py
```diff
@@ -44,8 +44,8 @@ def sierra_minimal_compiled_contract_and_class_hash() -> Tuple[str, int]:
     """
     Returns minimal contract compiled to sierra and its compiled class hash.
     """
-    compiled_contract = read_contract("minimal_contract_compiled.json")
-    compiled_contract_casm = read_contract("minimal_contract_compiled.casm")
+    compiled_contract = read_contract("minimal_contract_compiled.json", v1_v2=True)
```
tomek0123456789, line 47: Shouldn't it be in `contracts_v1.py` file since it is compiled to sierra?

starknet_py/net/signer/test_stark_curve_signer.py
drknzz: After many changes in `read_contract` and using `pytest.skip()` inside it, **pytest** complained about the `read_contract` not being called from a test case.

starknet_py/tests/e2e/fixtures/contracts.py
drknzz: Probably.
I saw this before, was just too lazy to change it 🐵

tomek0123456789 (approved):
👍 😎 👍

PR #1127, opened 2023-07-21
`lstrip` in `decode_shortstring`

drknzz:
Closes #

## Introduced changes

- strip empty bytes from the front in decode_shortstring

##


tomek0123456789 (commented):
💯 
I think a note in the migration guide would be useful, too.

drknzz:
> 💯
> I think a note in the migration guide would be useful, too.

Right, I was hoping I could do another PR with the changes after your PR with new 0.18 migration guide 👀

PR #1130, opened 2023-07-21
Add missing fields to dataclasses

### LINKED ISSUE #1119: Compare schemas and dataclasses with other sdks
### Feature Request

In case some field is missing / has some invalid type etc.

tomek0123456789:
Closes #1119 

## Introduced changes

- add missing fields to dataclasses
- add `get_full_contract` method to GatewayClient

##


starknet_py/net/client_models.py
```diff
@@ -173,6 +176,24 @@ class TransactionStatus(Enum):
...
+    poseidon_builtin: Optional[int] = None
+
+
+@dataclass
+class ExecutionResources:
+    n_steps: int
+    n_memory_holes: int
+    builtin_instance_counter: BuiltinInstanceCounter
+
+
```
drknzz, line 196: Let's not type this ourselves. It's a lot of internal stuff that could change and probably would be sufficient to make it just a dictionary.

starknet_py/net/schemas/gateway.py
```diff
@@ -188,6 +226,10 @@ class TransactionReceiptSchema(Schema):
         data_key="l2_to_l1_messages",
         load_default=[],
     )
+    transaction_index = fields.Integer(data_key="transaction_index", required=True)
+    execution_resources = fields.Nested(
```
drknzz, line 230: Dict would probably be enough

starknet_py/net/schemas/rpc.py
```diff
@@ -51,6 +52,7 @@
 # pylint: disable=unused-argument, no-self-use
 
 
+# TODO (#1119): what is this for, it isn't even used anywhere
```
drknzz, line 55: If it's unused, let's just remove it

starknet_py/net/client_models.py
```diff
@@ -192,7 +195,11 @@ class TransactionReceipt:
 
     events: List[Event] = field(default_factory=list)
     l2_to_l1_messages: List[L2toL1Message] = field(default_factory=list)
+
+    # gateway only
     l1_to_l2_consumed_message: Optional[L1toL2Message] = None
+    execution_resources: Optional[dict] = None
+    transaction_index: Optional[int] = None
```
drknzz, line 202: Do we need both `GatewayBlockTransactionReceipt` and this?

starknet_py/net/schemas/rpc.py
```diff
@@ -83,6 +76,8 @@ def make_dataclass(self, data, **kwargs):
 
 
 class L1toL2MessageSchema(Schema):
+    nonce = Felt(data_key="nonce", required=True)
+    selector = Felt(data_key="selector", required=True)
```
drknzz, line 80: Is it okay for these to be marked as required?
Won't it break v0.3?

starknet_py/net/client_models.py
tomek0123456789: We could, however more fields in `TransactionReceipt` would have to be marked as Optional

starknet_py/net/schemas/rpc.py
tomek0123456789: This schema could be as well removed, because the field connected to it isn't returned in FullNode, but is in Gateway

starknet_py/net/client_models.py
```diff
@@ -268,7 +298,9 @@ class GatewayBlock(StarknetBlockCommon):
     gas_price: int
     status: BlockStatus
     transactions: List[Transaction]
+    transaction_receipts: List[TransactionReceipt]
```
drknzz, line 301: Shouldn't this be `List[GatewayBlockTransactionReceipt]` now?

starknet_py/net/full_node_client.py
```diff
@@ -107,20 +111,25 @@ async def get_block(
             method_name="getBlockWithTxs",
             params=block_identifier,
         )
-        return cast(StarknetBlock, StarknetBlockSchema().load(res, unknown=EXCLUDE))
+        if "new_root" in res:
+            return cast(StarknetBlock, StarknetBlockSchema().load(res, unknown=EXCLUDE))
```
drknzz, line 115: In theory I think the if condition maybe should be something like 
```
if block_identifier == {"blockNumber": "pending"}:
```
but this also works I guess. wdyt 🤔

starknet_py/net/full_node_client.py
```diff
@@ -129,9 +138,15 @@ async def get_block_with_tx_hashes(
...
+        if "status" in res:
+            return cast(
+                StarknetBlockWithTxHashes,
+                StarknetBlockWithTxHashesSchema().load(res, unknown=EXCLUDE),
+            )
         return cast(
-            StarknetBlockWithTxHashes,
-            StarknetBlockWithTxHashesSchema().load(res, unknown=EXCLUDE),
+            PendingStarknetBlockWithTxHashes,
+            PendingStarknetBlockWithTxHashesSchema().load(res, unknown=EXCLUDE),
```
drknzz, line 149: Similar situation here

starknet_py/net/schemas/gateway.py
```diff
@@ -213,6 +219,26 @@ def make_dataclass(self, data, **kwargs):
...
+    events = fields.List(fields.Nested(EventSchema()), data_key="events", required=True)
+    execution_resources = fields.Dict(data_key="execution_resources", load_default=None)
+    actual_fee = Felt(data_key="actual_fee", required=True)
+
+    @post_load
+    def make_dataclass(self, data, **kwargs):
+        return GatewayBlockTransactionReceipt(**data)
+
+
 class StarknetBlockSchema(Schema):
```
drknzz, line 242: Who named it `StarknetBlock` smh 💀

starknet_py/net/schemas/rpc.py
```diff
@@ -214,6 +195,21 @@ class TypesOfTransactionsSchema(OneOfSchema):
...
 
+class PendingStarknetBlockSchema(Schema):
+    parent_hash = Felt(data_key="parent_hash", required=True)
+    sequencer_address = Felt(data_key="sequencer_address", required=True)
+    transactions = fields.List(
+        fields.Nested(TypesOfTransactionsSchema(unknown=EXCLUDE)),
+        data_key="transactions",
+        required=True,
+    )
+    timestamp = fields.Integer(data_key="timestamp", required=True)
```
drknzz, line 206: Same here, only transaction is marked as required.

drknzz (commented):
👍

starknet_py/net/full_node_client.py
tomek0123456789: Your version is more readable, I think

drknzz (approved):
lgtm

PR #1152, opened 2023-08-18
Starknet v0.12.2 update

### LINKED ISSUE #1149: Starknet v0.12.2 update
https://community.starknet.io/t/introducing-p2p-authentication-and-mismatch-resolution-in-v0-12-2/97993

tomek0123456789:
Closes #1149 

## Introduced changes

- added two methods to `GatewayClient` - `get_public_key` and `get_signature`
- `get_state_update` now accepts additional parameter - `include_block`
- changed `old_root`, `new_root` (`BlockStateUpdateSchema`) and `root` (`state_root` in `StarknetBlockSchema`) to be Felts

##


starknet_py/net/gateway_client.py
```diff
@@ -143,21 +147,32 @@ async def get_state_update(
         self,
         block_hash: Optional[Union[Hash, Tag]] = None,
         block_number: Optional[Union[int, Tag]] = None,
-    ) -> BlockStateUpdate:
+        include_block: Optional[bool] = None,
```
drknzz, line 150: What's the default behaviour when sending the query?
Maybe just set it to false as default?

starknet_py/net/gateway_client.py
```diff
@@ -143,21 +147,32 @@ async def get_state_update(
...
         """
-        Get the information about the result of executing the requested block
+        Get the information about the result of executing the requested block.
 
-        :param block_hash: Block's hash
-        :param block_number: Block's number (default "pending")
-        :return: BlockStateUpdate object representing changes in the requested block
+        :param block_hash: Block's hash.
+        :param block_number: Block's number (default "pending").
+        :param include_block: Flag deciding whether to include the queried block.
```
drknzz, line 157: Also mention the default value here then

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -658,5 +694,5 @@ async def test_get_new_state_update(
 
     assert (
         ReplacedClass(contract_address=contract_address, class_hash=class_hash)
-        in state_update.state_diff.replaced_classes
+        in state_update.state_diff.replaced_classes  # pyright: ignore
```
drknzz, line 697: what's the error here? is it okay to silence it?

starknet_py/tests/e2e/integration_tests/client_test.py
```diff
@@ -414,3 +416,45 @@ async def test_get_block(full_node_client_integration):
...
+
+@pytest.mark.skipif(
+    condition="--client=full_node" in sys.argv,
+    reason="Separate FullNode tests from Gateway ones.",
+)
+@pytest.mark.asyncio
+async def test_get_public_key(gateway_client_integration):
+    public_key = await gateway_client_integration.get_public_key()
+
+    assert isinstance(public_key, str)
```
drknzz, line 429: Could this be tested some better way than just checking the type? should this public_key be equal to some value we can check?

starknet_py/tests/e2e/integration_tests/client_test.py
```diff
@@ -414,3 +416,45 @@ async def test_get_block(full_node_client_integration):
...
+    reason="Separate FullNode tests from Gateway ones.",
+)
+@pytest.mark.asyncio
+async def test_get_signature(gateway_client_integration):
+    signature = await gateway_client_integration.get_signature(block_number="latest")
+
+    assert isinstance(signature, SignatureOnStateDiff)
+    assert signature.block_number is not None
+    assert len(signature.signature) == 2
+    assert signature.signature_input is not None
```
drknzz, line 443: Same here

drknzz (approved):
Green checks in tests is what I like 😎 👍 
The codecov got hit a bit though, would be good to do something about it before merging

starknet_py/net/gateway_client.py
tomek0123456789: I thought devnet would not like any additional arguments, but it seems that they are omitted. `return` condition would need to be changed a bit, from 
```python
        if include_block is not None:
```
to 
```python
        if res.get("block", None) is not None:
```
to allow compatibility with devnet tests.

starknet_py/tests/e2e/client/client_test.py
tomek0123456789: The error is here because `get_state_update` now has return type `Union[BlockStateUpdate, StateUpdateWithBlock]`. The latter one doesn't have a field `state_diff`, because it is in `state_update.state_diff`. However, this test doesn't call the method with `include_block=True` argument, hence the ignore is okay, at least in my opinion.

starknet_py/tests/e2e/integration_tests/client_test.py
tomek0123456789: We can, but it would have to be changed every year or in case it was leaked, wdyt?

drknzz: what's supposed to be leaked? the public key?

PR #1175, opened 2023-09-06
Trace API

### LINKED ISSUE #981: RPC Trace Api
### Feature Request

RPC Trace Api is not supported at this moment.

It probably should be implemented at some point in the future.

Blocked by devnet not supporting trace api.

tomek0123456789:
Closes #981

## Introduced changes

- added Trace API
- changed devnet to 0.6.2 and cairo-lang to 0.12.2

##


starknet_py/net/full_node_client.py
```diff
@@ -659,6 +657,110 @@ async def get_contract_nonce(
...
+        )
+        return cast(
+            TransactionTrace, TransactionTraceSchema().load(res, unknown=EXCLUDE)
+        )
+
+    async def simulate_transactions(
+        self,
+        transactions: List[AccountTransaction],
+        skip_validate: bool = False,
+        skip_fee_charge: bool = False,
```
DelevoXDG, line 686: Are you sure you don't want to pass a Set of `SimulationFlag` instead?
It just seems to me it'd be less obvious to use and understand from documentation, since you don't have a direct connection between `SimulationFlag` Enum which clearly documents all the available flags and the method signature.
Also, if the number of flags used grows, the current approach might quickly get out of hand.

The approach with using set would also make refactoring easier, you would only need to modify a enum without changing function signature and body every time.

starknet_py/net/full_node_client.py
```diff
@@ -659,6 +657,110 @@ async def get_contract_nonce(
...
+        """
+        block_identifier = get_block_identifier(
+            block_hash=block_hash, block_number=block_number
+        )
+
+        simulation_flags = []
+        if skip_validate:
+            simulation_flags.append("SKIP_VALIDATE")
+        if skip_fee_charge:
+            simulation_flags.append("SKIP_FEE_CHARGE")
```
DelevoXDG, line 711: Are you sure these shouldn't be object of `SimulationFlag` enum instead?

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,171 @@ class PendingTransactionsSchema(Schema):
...
+        return {"revert_reason": data["revert_reason"]}
+
+
+class InvokeTransactionTraceSchema(Schema):
+    validate_invocation = fields.Nested(
+        FunctionInvocationSchema(), data_key="validate_invocation", load_default=None
+    )
+    execute_invocation = fields.Nested(
+        ExecuteInvocationSchema(), data_key="execute_invocation", required=True
+    )
```
DelevoXDG, line 596: Not required?

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,171 @@ class PendingTransactionsSchema(Schema):
...
+        return DeclareTransactionTrace(**data)
+
+
+class DeployAccountTransactionTraceSchema(Schema):
+    validate_invocation = fields.Nested(
+        FunctionInvocationSchema(), data_key="validate_invocation", load_default=None
+    )
+    constructor_invocation = fields.Nested(
+        FunctionInvocationSchema(), data_key="constructor_invocation", required=True
+    )
```
DelevoXDG, line 629: Not required?

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,171 @@ class PendingTransactionsSchema(Schema):
...
+
+    @post_load
+    def make_dataclass(self, data, **kwargs) -> DeployAccountTransactionTrace:
+        return DeployAccountTransactionTrace(**data)
+
+
+class L1HandlerTransactionTraceSchema(Schema):
+    function_invocation = fields.Nested(
+        FunctionInvocationSchema(), data_key="function_invocation", required=True
+    )
```
DelevoXDG, line 644: Not required?

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,171 @@ class PendingTransactionsSchema(Schema):
...
+    def get_data_type(self, data):
+        # All possible types of transaction trace (loaded by sketchy logic),
+        # it's possible that more are added later in the future, and it needs to be reformatted
+        if "function_invocation" in data:
+            return "L1_HANDLER"
+        if "constructor_invocation" in data:
+            return "DEPLOY_ACCOUNT"
+        if "execute_invocation" in data:
+            return "INVOKE"
+        return "DECLARE"
```
DelevoXDG, line 668: Discussed offline previously, but it should be mentioned that until [changes from this PR](https://github.com/starkware-libs/starknet-specs/pull/131) are added to nodes, this is unstable and might or might not fly depending on the node, since none of the properties are actually required in the spec. 

I'm not sure there's a good solution for this aside from waiting, unless someone has an idea.

Either way, we should definitely rework this once 0.5 drops, so maybe we should add an issue for that?

starknet_py/tests/e2e/fixtures/constants.py
```diff
@@ -22,6 +22,8 @@
 
 INTEGRATION_GATEWAY_URL = "https://external.integration.starknet.io"
 
+TESTNET_NODE_URL = "http://188.34.188.184:6060"
+
```
DelevoXDG, line 26: Is this URL of our node? 👀

starknet_py/net/full_node_client.py
```diff
@@ -659,6 +657,110 @@ async def get_contract_nonce(
...
+        block_hash: Optional[Union[Hash, Tag]] = None,
+        block_number: Optional[Union[int, Tag]] = None,
+    ) -> List[SimulatedTransaction]:
+        # pylint: disable=too-many-arguments
+        """
+        Simulates a given sequence of transactions on the requested state, and generates the execution traces.
+        If one of the transactions is reverted, raises CONTRACT_ERROR.
+
+        :param transactions: Transactions to be traced.
+        :param skip_validate: Flag checking whether there are enough funds in the account.
```
DelevoXDG, line 696: Where do you have info on exact purpose of `skip_validate`? This is not a source of truth and might be a devnet specific thing, but [in starknet-jvm ](https://github.com/software-mansion/starknet-jvm/blob/f551c0827935aa79bda1b29898668a21f6af870f/lib/src/test/kotlin/starknet/account/StandardAccountTest.kt#L777) we actually test this flag for transactions with removed signature.

DelevoXDG (commented):
Nice work! 👍🏻  Added some comments, some of them are subjective, so someone please TAL.

starknet_py/net/client_models.py
```diff
@@ -401,6 +401,8 @@ class BlockSingleTransactionTrace:
     validate_invocation: Optional[dict] = None
     fee_transfer_invocation: Optional[dict] = None
     constructor_invocation: Optional[dict] = None
+    # Gateway-only field, revert_error in RPC spec is returned in
```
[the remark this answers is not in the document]
tomek0123456789: woops, didn't finish that, thanks!

poetry.lock
```diff
@@ -2524,6 +2540,26 @@ pytest = ">=5.0"
...
+]
+
+[package.dependencies]
+gprof2dot = "*"
+pytest = "*"
+six = "*"
+
+[package.extras]
+tests = ["pytest-virtualenv"]
+
```
drknzz, line 2562: Has this been added manually?

starknet_py/net/full_node_client.py
```diff
@@ -659,6 +657,110 @@ async def get_contract_nonce(
...
+                    _create_broadcasted_txn(transaction=t) for t in transactions
+                ],
+            },
+        )
+        return cast(
+            List[SimulatedTransaction],
+            SimulatedTransactionSchema().load(res, unknown=EXCLUDE, many=True),
+        )
+
+    async def get_block_traces(
```
drknzz, line 728: Let's change it to `starknet_traceBlockTransactions` in snake case as in the spec

starknet_py/net/schemas/gateway.py
```diff
@@ -661,3 +650,18 @@ class SignatureOnStateDiffSchema(Schema):
...
     def make_dataclass(self, data, **kwargs) -> SignatureOnStateDiff:
         return SignatureOnStateDiff(**data)
+
+
+# Trace API schemas
+
+
+class BlockTransactionTracesSchema(Schema):
+    traces = fields.List(
+        fields.Nested(BlockSingleTransactionTraceSchema(unknown=EXCLUDE)),
```
drknzz, line 660: Do we need the unknown=EXCLUDE?

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,171 @@ class PendingTransactionsSchema(Schema):
...
+        ),
+        data_key="calls",
+        required=True,
+    )
+    events = fields.List(
+        fields.Nested(EventContentSchema()), data_key="events", required=True
+    )
+    messages = fields.List(
+        fields.Nested(L2toL1MessageSchema()), data_key="messages", required=True
+    )
```
drknzz, line 561: Are all of these required?

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,171 @@ class PendingTransactionsSchema(Schema):
...
+    function_invocation = fields.Dict(data_key="function_invocation", load_default=None)
+
+    @pre_load
+    def alter_data(self, data, **kwargs):
+        # This method does data preprocessing before it is loaded into the fields above. If tx is not reverted, data
+        # gets loaded into 'function_invocation' data by adding a dict layer, otherwise gets loaded into "revert_reason"
+        assert isinstance(data, dict)
+        if data.get("revert_reason", None) is None:
+            data = {"function_invocation": data}
+        return data
```
drknzz, line 579: Could it not be if "revert_reason" in data?

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,171 @@ class PendingTransactionsSchema(Schema):
...
+
+    @post_load
+    def make_dataclass_or_string(
+        self, data, **kwargs
+    ) -> Union[FunctionInvocation, dict]:
+        if data.get("revert_reason", None) is None:
```
drknzz, line 585: Same here

poetry.lock
tomek0123456789: I guess those are dependencies to either `cairo-lang` or `devnet`, they happened to install after updating those two packages' versions and doing `poetry lock --no-update` and then `poetry install`.

starknet_py/net/schemas/gateway.py
tomek0123456789: That's the problem with gateway - we don't know (the code was already here, I just moved it to the bottom of the file 😅)

starknet_py/net/schemas/rpc.py
tomek0123456789: In spec those are specified as [`"allOf"`](https://github.com/starkware-libs/starknet-specs/blob/v0.5.0/api/starknet_trace_api_openrpc.json#L290)

starknet_py/net/schemas/rpc.py
tomek0123456789: Could be, thanks!

starknet_py/net/client_models.py
```diff
@@ -702,3 +704,146 @@ class SignatureOnStateDiff:
...
+@dataclass
+class InvokeTransactionTrace:
+    """
+    Dataclass representing a transaction trace of an INVOKE transaction.
+    """
+
+    validate_invocation: Optional[FunctionInvocation]
+    execute_invocation: Union[
+        FunctionInvocation, dict
+    ]  # dict in case "revert_reason" is returned
```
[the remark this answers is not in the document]
tomek0123456789: I think it just wasn't specified in previous versions of RPC that it is required (it happened a few times). In v0.5.0 it is required, though.

starknet_py/net/client_models.py
```diff
@@ -702,3 +704,146 @@ class SignatureOnStateDiff:
...
+
+
+@dataclass
+class DeployAccountTransactionTrace:
+    """
+    Dataclass representing a transaction trace of an DEPLOY_ACCOUNT transaction.
+    """
+
+    validate_invocation: Optional[FunctionInvocation]
+    constructor_invocation: FunctionInvocation
```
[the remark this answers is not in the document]
tomek0123456789: as in comment above

starknet_py/net/client_models.py
```diff
@@ -702,3 +704,146 @@ class SignatureOnStateDiff:
...
+    fee_transfer_invocation: Optional[FunctionInvocation]
+
+
+@dataclass
+class L1HandlerTransactionTrace:
+    """
+    Dataclass representing a transaction trace of an L1_HANDLER transaction.
+    """
+
+    function_invocation: FunctionInvocation
```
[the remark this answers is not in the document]
tomek0123456789: Yup, will target 0.5.0 with possibly optional `type` field to make it backwards compatible for a while

starknet_py/net/client_models.py
```diff
@@ -702,3 +704,146 @@ class SignatureOnStateDiff:
...
+
+
+@dataclass
+class TransactionTraceAndHash:
+    """
+    Dataclass representing a single pair of transaction hash and a corresponding trace.
+    """
+
+    transaction_hash: int
+    trace_root: TransactionTrace
```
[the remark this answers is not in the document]
tomek0123456789: Good catch, it was supposed to be a class for `traceBlockTransactions`, but I forgot about it and created one with more valid name 💀

starknet_py/net/client_models.py
```diff
@@ -702,3 +704,146 @@ class SignatureOnStateDiff:
...
+
+# TODO (#981): consider changing Client interface return type to match both implementations
+@dataclass
+class BlockTransactionTrace:
+    """
+    Dataclass representing a single transaction trace in a block.
+    """
+
+    transaction_hash: int
+    trace_root: TransactionTrace
```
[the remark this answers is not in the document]
tomek0123456789: Exactly 😅

starknet_py/net/full_node_client.py
tomek0123456789: Yup, you're right ((I stole the idea from starknet.js)), but currently if there are only two flags (and I suspect there won't be more added), the interface is more user-friendly. Passing a bool argument is (at least imo) better than importing an enum just to pass it as an element of an array.  cc @drknzz

starknet_py/net/full_node_client.py
tomek0123456789: Yup, those would have to be serializable and passing a string is simply easier. If more flags are added, then it would need to be changed.

starknet_py/net/full_node_client.py
drknzz: Agree with the set approach, however we shouldn't expect any more flags here. Current approach is better ui 👍

starknet_py/tests/e2e/fixtures/constants.py
tomek0123456789: 🤭

starknet_py/net/schemas/rpc.py
tomek0123456789: Yup, it will need a refactor, but until actual 0.5.0 drop, it doesn't need to be touched. Issue [here](https://github.com/software-mansion/starknet.py/issues/1177)

starknet_py/net/schemas/rpc.py
drknzz: Does that imply that all of them are required?

tomek0123456789: Hard to say, sometimes `required` is specified inside `allOf`, sometimes not. It's not specified in v0.5.0 either, I'll ask a question under the PR regarding that

starknet_py/net/client_models.py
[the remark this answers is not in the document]
DelevoXDG: `allOf` in the spec contains two objects. First object has their fields set as required, second one does not.

starknet_py/net/client_models.py
```diff
@@ -702,3 +704,146 @@ class SignatureOnStateDiff:
...
+    entry_point_selector: int
+    calldata: List[int]
+    caller_address: int
+    class_hash: int
+    entry_point_type: EntryPointType
+    call_type: CallType
+    result: List[int]
+    calls: List["FunctionInvocation"]
+    events: List[Event]
+    messages: List[L2toL1Message]
```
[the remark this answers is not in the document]
DelevoXDG: `allOf` in the spec contains two objects. First object has their fields set as required, second one does not.

starknet_py/net/full_node_client.py
DelevoXDG: I don't really mind, but would it problematic to make them serializable? Raw strings can cause unexpected problems in the future 😅

starknet_py/tests/e2e/fixtures/constants.py
DelevoXDG: Would it be hard to retrieve those from env like we do in [jvm sdk](https://github.com/software-mansion/starknet-jvm/pull/281)?

I don't really think we should be keeping secrets out in the open.

starknet_py/net/schemas/rpc.py
DelevoXDG: Not all of them, mentioned here https://github.com/software-mansion/starknet.py/pull/1175/files/089c91c2a334930384bc2cffc11d238c92e2e51a#r1320093740

docs/migration_guide.rst
```diff
@@ -1,6 +1,83 @@
 Migration guide
 ===============
 
+**********************
+0.18.2 Migration guide
+**********************
+
+Version 0.18.2 of **starknet.py** comes with support of RPC v0.4.0 Trace API!
```
drknzz, line 8: Let's link the rpc

docs/migration_guide.rst
```diff
@@ -1,6 +1,83 @@
...
+
+
+0.18.2 Minor changes
+--------------------
+
+1. :meth:`Client.trace_block_transactions` return type has been changed from ``BlockTransactionTraces`` to ``Union[BlockTransactionTraces, List[BlockTransactionTrace]]``.
+
+.. currentmodule:: starknet_py.net.gateway_client
+
+2. Reverted error while using ``include_block`` in :meth:`GatewayClient.get_state_update` parameter on gateway mainnet.
```
drknzz, line 26: Maybe reword to sound out that *include block works on gateway mainnet*

docs/migration_guide.rst
```diff
@@ -1,6 +1,83 @@
...
+|
+
+**********************
+0.18.1 Migration guide
+**********************
+
+.. currentmodule:: starknet_py.net.gateway_client
+
+This version contains a quick fix to :meth:`GatewayClient.get_state_update` method (mainnet wasn't updated to 0.12.2 then).
+Additionally, accounts in Cairo1 are now supported!
```
drknzz, line 60: What does that mean?
edit; it'd be good to have the `cairo_version` change up here or sth, for it to be more visible

starknet_py/tests/e2e/test-variables.env.template
```diff
@@ -0,0 +1,6 @@
+INTEGRATION_NODE_URL=URL
+TESTNET_NODE_URL=URL
+INTEGRATION_ACCOUNT_PRIVATE_KEY=0x1234567890
+INTEGRATION_ACCOUNT_ADDRESS=0x9876543210
+TESTNET_ACCOUNT_PRIVATE_KEY=0xabcdef
+TESTNET_ACCOUNT_ADDRESS=0xfedcba
```
drknzz, line 6: So running the tests will fail because these are not set?
Maybe it would be better to not make this a template, but an actual file with some dummy values?

starknet_py/tests/e2e/fixtures/constants.py
```diff
@@ -1,24 +1,45 @@
...
-TESTNET_ACCOUNT_PRIVATE_KEY = (
-    "0x61910356c5adf66efb65ec3df5d07a6e5e6e7c8b59f15a13eda7a34c8d1ecc4"
-)
-TESTNET_ACCOUNT_ADDRESS = (
-    "0x59083382aadec25d7616a7f48942d72d469b0ac581f2e935ec26b68f66bd600"
-)
+if (TESTNET_ACCOUNT_ADDRESS := os.getenv("TESTNET_ACCOUNT_ADDRESS")) is not None:  # pyright: ignore
+    TESTNET_ACCOUNT_ADDRESS: str = TESTNET_ACCOUNT_ADDRESS
+else:
+    raise ValueError("TESTNET_ACCOUNT_ADDRESS env variable is not set.")
```
drknzz: :exploding_head:

.github/workflows/checks.yml
```diff
@@ -169,6 +169,13 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [ "3.8", "3.9", "3.10", "3.11" ]
+    env:
+      INTEGRATION_NODE_URL: ${{ secrets.INTEGRATION_NODE_URL }}
+      TESTNET_NODE_URL: ${{ secrets.TESTNET_NODE_URL }}
+      INTEGRATION_ACCOUNT_PRIVATE_KEY: ${{ secrets.INTEGRATION_ACCOUNT_PRIVATE_KEY }}
+      INTEGRATION_ACCOUNT_ADDRESS: ${{ secrets.INTEGRATION_ACCOUNT_ADDRESS }}
+      TESTNET_ACCOUNT_PRIVATE_KEY: ${{ secrets.TESTNET_ACCOUNT_PRIVATE_KEY }}
+      TESTNET_ACCOUNT_ADDRESS: ${{ secrets.TESTNET_ACCOUNT_ADDRESS }}
```
drknzz, line 178: any way we could remove the duplication in each job?

drknzz (commented):
good job 🔥 
what do you think of writing in the github release the info of targeted versions, so something like: starknet.py 0.x.y, rpc 0.4.0, starknet 0.12.2

.github/workflows/checks.yml
tomek0123456789: Probably after refactoring the logic behind those envs it will be possible

.github/workflows/checks.yml
```diff
@@ -390,6 +411,13 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [ "3.8", "3.9", "3.10", "3.11" ]
+    env:
+      INTEGRATION_NODE_URL: ${{ secrets.INTEGRATION_NODE_URL }}
+      TESTNET_NODE_URL: ${{ secrets.TESTNET_NODE_URL }}
```
DelevoXDG, line 416: Maybe we should use the same naming in starknet.py and starknet-jvm, to avoid confusion in case someone uses both?
It's `INTEGRATION_RPC_URL` in jvm, we could decide on either one and use it in both SDKs.

starknet_py/net/client_models.py
```diff
@@ -702,3 +704,140 @@ class SignatureOnStateDiff:
...
+@dataclass
+class InvokeTransactionTrace:
+    """
+    Dataclass representing a transaction trace of an INVOKE transaction.
+    """
+
+    validate_invocation: Optional[FunctionInvocation]
+    execute_invocation: Union[
+        FunctionInvocation, dict
+    ]  # dict in case "revert_reason" is returned
```
DelevoXDG, line 784: What do you think of creating a dedicated dataclass like `RevertedExecuteInvocation` or `RevertedFunctionInvocation` instead of using dict? It would have a single field `revert_reason`, but the purpose of return type would be clearer and it would be easier to expand error response should it be changed.

starknet_py/net/schemas/rpc.py
```diff
@@ -510,3 +522,172 @@ class PendingTransactionsSchema(Schema):
...
+            data = {"function_invocation": data}
+        return data
+
+    @post_load
+    def make_dataclass_or_string(
+        self, data, **kwargs
+    ) -> Union[FunctionInvocation, dict]:
+        if "revert_reason" not in data:
+            return FunctionInvocation(**data["function_invocation"])
+        return {"revert_reason": data["revert_reason"]}
```
DelevoXDG, line 587: Like i mentioned before, maybe it could get simpler if a dataclass with `revert_reason` is used instead.

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -383,15 +374,22 @@ async def test_deploy_account_raises_on_incorrect_address(
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    "call_contract",
+    "call_contract, corresponding_client",
```
DelevoXDG, line 377: this is not important, but maybe just `client`? 😄

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -2,21 +2,34 @@
...
+    SimulatedTransaction,
     SyncStatus,
     TransactionType,
 )
 from starknet_py.net.full_node_client import _to_rpc_felt
 from starknet_py.net.models import StarknetChainId
+from starknet_py.tests.e2e.fixtures.constants import (
+    CONTRACTS_COMPILED_DIR,
+    CONTRACTS_COMPILED_V1_DIR,
+)
```
DelevoXDG, line 31: I know it was there before, but maybe change this to `CONTRACTS_COMPILED_V0_DIR`.

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -400,3 +413,179 @@ async def test_get_syncing_status(full_node_client):
...
+    )
+    invoke_tx = await full_node_account.sign_invoke_transaction(
+        calls=call, auto_estimate=True
+    )
+
+    simulated_txs = await full_node_account.client.simulate_transactions(
+        transactions=[invoke_tx], skip_validate=True, block_number="latest"
+    )
+
+    assert simulated_txs[0].transaction_trace.validate_invocation is None
```
DelevoXDG, line 439: Maybe instead test this with transaction that has no signature like we do in jvm and swift sdks? 
Otherwise it doesn't seem to differ much from other tests that don't have `skip_validate`.

DelevoXDG (commented):
Left few suggestions, otherwise looks good

starknet_py/tests/e2e/account/account_test.py
tomek0123456789: Thought it would break the test since we have a fixture called `client`, but seems to be working!

DelevoXDG (approved):
lgtm 🔥 
@drknzz could you please TAL?

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -400,3 +414,184 @@ async def test_get_syncing_status(full_node_client):
...
+        selector=get_selector_from_name("increase_balance"),
+        calldata=[0x10],
+    )
+    invoke_tx = await full_node_account.sign_invoke_transaction(
+        calls=call, auto_estimate=True
+    )
+    invoke_tx = dataclasses.replace(invoke_tx, signature=[])
+
+    simulated_txs = await full_node_account.client.simulate_transactions(
+        transactions=[invoke_tx], skip_validate=True, block_number="latest"
```
drknzz, line 438: So the same tx without skip_validate would be rejected?

drknzz: Ok, i see below 👍

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -400,3 +414,184 @@ async def test_get_syncing_status(full_node_client):
...
+    )
+    assert simulated_txs[0].transaction_trace.validate_invocation is None
```
drknzz, line 440: Can we check the contents of the validate_invocation or there is no need since it's specifically VALIDATE_invocation?

starknet_py/tests/e2e/fixtures/accounts.py
```diff
@@ -218,6 +221,11 @@ def pre_deployed_account_with_validate_deploy(
     net = pytestconfig.getoption("--net")
     address, private_key = address_and_priv_key[net]
 
+    if net != "devnet":
+        # because TESTNET and INTEGRATION fixtures are lambdas
```
drknzz, line 225: it this comment correct?

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -400,3 +414,184 @@ async def test_get_syncing_status(full_node_client):
...
+        calls=call, auto_estimate=True
+    )
+    simulated_txs = await full_node_account.client.simulate_transactions(
+        transactions=[invoke_tx], block_number="latest"
+    )
+
+    assert isinstance(simulated_txs[0], SimulatedTransaction)
+    assert isinstance(simulated_txs[0].transaction_trace, InvokeTransactionTrace)
+    assert simulated_txs[0].fee_estimation.overall_fee > 0
+    assert simulated_txs[0].transaction_trace.validate_invocation is not None
```
drknzz, line 490: Would it make sense to check the contents of this?

starknet_py/tests/e2e/client/full_node_test.py
```diff
@@ -400,3 +414,184 @@ async def test_get_syncing_status(full_node_client):
...
+    invoke_tx = await full_node_account.sign_invoke_transaction(
+        calls=[call, call], auto_estimate=True
+    )
+    simulated_txs = await full_node_account.client.simulate_transactions(
+        transactions=[invoke_tx], block_number="latest"
+    )
+
+    assert isinstance(simulated_txs[0].transaction_trace, InvokeTransactionTrace)
+    assert simulated_txs[0].fee_estimation.overall_fee > 0
+    assert simulated_txs[0].transaction_trace.validate_invocation is not None
```
drknzz, line 501: Would it make sense to assert other fields of transaction_trace?

starknet_py/tests/e2e/client/full_node_test.py
drknzz: Nevermind, missread. It's None because of the skip_validate flag?

.github/workflows/checks.yml
```diff
@@ -227,15 +227,22 @@ jobs:
         uses: codecov/codecov-action@v3
 
   # ---------------------------------------------------------- #
-  # ..................RUN-TESTS-INTEGRATION................... #
+  # ..................RUN-TESTS-ON-NETWORKS................... #
   # ---------------------------------------------------------- #
 
   run-tests-integration:
```
drknzz, line 233:
```suggestion
  run-tests-on-networks:
```

drknzz (approved):
giga job 💯

starknet_py/net/client.py
```diff
@@ -68,11 +69,11 @@ async def get_block(
         """
 
     @abstractmethod
-    async def get_block_traces(
+    async def trace_block_transactions(
         self,
         block_hash: Optional[Union[Hash, Tag]] = None,
         block_number: Optional[Union[int, Tag]] = None,
-    ) -> BlockTransactionTraces:
+    ) -> Union[BlockTransactionTraces, List[BlockTransactionTrace]]:
```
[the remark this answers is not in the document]
tomek0123456789: `BlockTransactionTraces` is a """legacy""" class for gateway (everything there is a dict), `List[BlockTransactionTrace]` is for RPC with exact classes as in spec

starknet_py/tests/e2e/client/full_node_test.py
tomek0123456789: Yup

starknet_py/tests/e2e/fixtures/accounts.py
tomek0123456789: uhh, no, thanks 😅

starknet_py/tests/e2e/client/full_node_test.py
tomek0123456789: it is a problem with devnet, because it doesn't return `fee_transfer_invocation` at all. I can add `execute_invocation` though

starknet_py/tests/e2e/client/full_node_test.py
tomek0123456789: Contents of `validate_invocation`?

DelevoXDG: None of them are optional, so if something would be missing probably there would be an error anyway?

PR #1183, opened 2023-09-14
Change `cairo_version` into Account property

### LINKED ISSUE #1182: Add `cairo_version` parameter to `PreparedFunctionCall.invoke`

tomek0123456789:
Closes #1182

## Introduced changes

- add `cairo_version` parameter to `Account` constructor and `Account.deploy_account` method
- add `cairo_version` property

##


starknet_py/tests/e2e/fixtures/accounts.py
```diff
@@ -172,9 +172,13 @@ class AccountToBeDeployedDetailsFactory:
     class_hash: int
     fee_contract: Contract
 
-    async def get(self) -> AccountToBeDeployedDetails:
+    async def get(
+        self, *, class_hash: Optional[int] = None, argent_calldata: bool = False
```
drknzz, line 176: could it be class_hash = self.class_hash?

drknzz (approved):
Looks good :+1:
Let's just have a second look around the library to make sure no place with account was omitted.

starknet_py/tests/e2e/fixtures/accounts.py
tomek0123456789: Nope, `self` cannot be accessed before the method body

starknet_py/net/account/account.py
```diff
@@ -96,6 +99,10 @@ def __init__(
     def address(self) -> int:
         return self._address
 
+    @property
+    def cairo_version(self) -> int:
+        return self._cairo_version
```
ClementWalter, line 104: here you could do instead something like

```
@property
async def cairo_version(self) -> int:
    if self._cairo_version is None:
        # Get the class hash of self.address
        class_ = await self._client.get_class_at(self.address)
        self.cairo_version = 1 if isinstance(class, Sierra) else 0
    return self._cairo_version
```

ClementWalter (commented):
I think that you should get the cairo_version completely internal, and set it async based on the `address` in the fixture, though the current code works as well

starknet_py/net/account/account.py
tomek0123456789: I'm not sure if an async property would be a good UI. However, we've got a sync interface for that, so something like this would work without `await`ing:
```py
@property
def cairo_version(self) -> int:
    if self._cairo_version is None:
        if isinstance(self._client, GatewayClient):
            contract_class = self._client.get_full_contract_sync(contract_address=self._address)
        else:
            contract_class = self.client.get_class_at_sync(contract_address=self._address)
        self._cairo_version = 1 if isinstance(contract_class, SierraContractClass) else 0
    return self._cairo_version
```
wdyt?

ClementWalter: well, just see this comment now. Async property you're right may not have been the best thing, but with sync definitely as It's quite common to use property to cache a given computed/fetched data

there is no information regarding the internal implementation for the outside world, use you still use `self.cairo_version`

PR #1201, opened 2023-10-18
RPC v0.5.0

tomek0123456789:
Closes #1186
Closes #1177
Closes #1169 

## Introduced changes

- added RPC 0.5.0

##

- [x] This PR contains breaking changes

- a few, find them in migration guide

PR #1206, opened 2023-10-31
Fix reverted error message in wait_for_tx

### LINKED ISSUE #1203: Investigate wrong error messages when creating `ClientError`
https://discord.com/channels/793094838509764618/1162147867859894333

There seems to be a problem with parsing errors from blastapi public node (`https://starknet-mainnet.public.blastapi.io/`) - AFAIK it hosts a Pathfinder node, so we can test it locally on our one.

edit: possible that the issue is related to [`data` field in RPC spec](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3383).

ddoktorski:
Closes #1203

## Introduced changes

- Provide correct error message for both clients in `TransactionRevertedError` in function `wait_for_tx`

Some tests from `tests_ci_on_network` expected to fail at this point. It should be fixed once #1201 is merged.

##

### DIFF
--- a/starknet_py/net/client.py
+++ b/starknet_py/net/client.py
@@ -179,7 +179,10 @@ class Client(ABC):
                     raise TransactionRejectedError(message=tx_receipt.rejection_reason)
 
                 if execution_status == TransactionExecutionStatus.REVERTED:
-                    raise TransactionRevertedError(message=tx_receipt.revert_error)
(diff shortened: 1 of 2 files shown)


tomek0123456789 (approved):
LGTM!

PR #1217, opened 2023-11-07
Add method to get gas price

### LINKED ISSUE #1140: Investigate how to implement `get_gas_price` method
There were some questions about such a method on Discord. We should investigate if it is possible and how to implement it.

ddoktorski:
Closes #1140 

## Introduced changes

- Function to get gas price for a given block
- If `block_hash` and `block_number` are not specified then it returns a price for current `pending` block

Note: For now `price_in_strk` is always `None`

##


ddoktorski:
As discussed with @tomek0123456789 we decided not to add this method given that it is easy to retrieve `gas_price` value from a block and additional wrapper seems unnecessary.

Issue #1219, opened 2023-11-08
Refactor tests


ddoktorski:
- Redesign general tests structure
- Move all test files to `tests` directory in subdirectories `tests/e2e` and `tests/unit`
- Consider refactoring fixtures e.g. limit scope for some fixtures to module instead of package

It should be done as part of #1047

PR #1225, opened 2023-11-16
Remove GatewayClient

### LINKED ISSUE #1047: Remove GatewayClient
### Feature Request

After it is removed from Starknet, it should be removedfrom Starknet.py

Also, models refactoring is related to this issue - should be done as in RPC spec (consider splitting into different files in `net.models` module).

ddoktorski:
Related to #1047, #1219

## Introduced changes

- Remove `GatewayClient`
- Remove `--client` flag from tests
- Remove `/tests/e2e/core`
- Modify `pyproject.toml` and `checks.yml` to include above changes

Changes that will be addressed in separate PRs (in the scope to close #1047):
- Remove `GatewayClient` from documentation
- Redesign `client_models` to have it aligned with RPC spec, it includes schema changes - `/net/schemas/gateway.py` file is only slightly changed in this PR

It is also worth considering to remove `Client` class.

Please note that some failing tests on integration network are expected and it is not caused by the changes in this PR.

edit: PR with updated docs #1226 

##


starknet_py/tests/e2e/docs/code_examples/test_full_node_client.py
```diff
@@ -56,42 +56,40 @@ async def test_get_storage_at(full_node_client, map_contract):
...
-    )
+    transaction_receipt = await client.get_transaction_receipt(tx_hash=transaction_hash)
     # docs-end: get_transaction_receipt
 
 
 @pytest.mark.asyncio
-async def test_estimate_fee(full_node_account, deploy_account_transaction):
-    full_node_client = full_node_account.client
+async def test_estimate_fee(account, deploy_account_transaction):
+    client = account.client
```
tomek0123456789, line 79: `client` fixture would be sufficient here.

starknet_py/tests/e2e/docs/guide/test_cairo1_contract.py
```diff
@@ -85,7 +85,5 @@ async def test_cairo1_contract(
     assert isinstance(contract_deployment.address, int)
     assert contract_deployment.address != 0
 
-    compiled_class = await gateway_client.get_compiled_class_by_class_hash(
-        class_hash=sierra_class_hash
-    )
-    assert isinstance(compiled_class, CasmClass)
+    compiled_class = await client.get_class_by_hash(class_hash=sierra_class_hash)
```
tomek0123456789, line 88: `account.client` can be used here and `client` fixture can be removed.

starknet_py/tests/e2e/tests_on_networks/client_test.py
```diff
@@ -65,11 +54,6 @@ async def test_wait_for_tx_reverted_full_node(full_node_account_integration):
         await account.client.wait_for_tx(tx_hash=invoke.transaction_hash)
 
 
-# Same here as in comment above 'test_wait_for_tx_reverted_full_node'
-@pytest.mark.skipif(
-    condition="--client=gateway" in sys.argv,
-    reason="Gateway client has been disabled on integration network.",
-)
 @pytest.mark.asyncio
 async def test_wait_for_tx_full_node_accepted(full_node_account_integration):
```
tomek0123456789, line 58: In order to stay consistent (my bad here 😅), renaming to suggestion below (or renaming the test above to match   name of this test) would be great:

```suggestion
async def test_wait_for_tx_accepted_full_node(full_node_account_integration):
```

starknet_py/net/client.py
```diff
@@ -179,10 +179,7 @@ async def wait_for_tx(
                     raise TransactionRejectedError(message=tx_receipt.rejection_reason)
 
                 if execution_status == TransactionExecutionStatus.REVERTED:
-                    # TODO (#1047): message should be always revert_reason once GatewayClient is deprecated
-                    raise TransactionRevertedError(
-                        message=(tx_receipt.revert_reason or tx_receipt.revert_error)
-                    )
+                    raise TransactionRevertedError(message=tx_receipt.revert_reason)
```
tomek0123456789, line 182: This is not covered by codecov because of old devnet, right?

tomek0123456789 (commented):
image

starknet_py/net/client.py
ddoktorski: Currently it is covered in `tests_on_networks/client_test.py` [here](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/tests/e2e/tests_on_networks/client_test.py#L64), but in this PR those tests failed. It seems to be some general issue with integration network, that needs to be investigated further.

starknet_py/tests/e2e/client/fixtures/transactions.py
```diff
@@ -35,6 +36,7 @@ async def deploy_account_transaction(
         class_hash=class_hash,
         salt=salt,
         network=network,
```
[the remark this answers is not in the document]
ddoktorski: Actually I think it's good idea to remove `network` from `get_deploy_account_transaction` in this PR

starknet_py/tests/e2e/tests_on_networks/client_test.py
ddoktorski: I guess we can also remove `_full_node` from the function name

starknet_py/tests/e2e/client/client_test.py
```diff
@@ -258,24 +249,9 @@ async def test_get_class_by_hash(client, class_hash):
...
-async def test_wait_for_tx_accepted(client, get_tx_receipt, request):
-    get_tx_receipt = request.getfixturevalue(get_tx_receipt)
-
+async def test_wait_for_tx_accepted(client, get_tx_receipt_path):
     with patch(
-        get_tx_receipt,
+        get_tx_receipt_path,
         AsyncMock(),
     ) as mocked_receipt:
         mocked_receipt.return_value = TransactionReceipt(
```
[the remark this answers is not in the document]
ddoktorski: Thanks, good point!

DelevoXDG (approved):
lgtm

starknet_py/net/client.py
tomek0123456789: Sure, but later we'll still need to create new tests on integration, since it is the first network to receive changes (if we want to be up-to-date with RPC updates).

DelevoXDG: >  we'll still need to create new tests on integration, since it is the first network to receive changes

Sure, but if we merge testnet and integration tests, we could still pass integration-only data to parametrized test, until the changes are not available on testnet.

ddoktorski: Thanks! Added this point in #1219 to consider this during tests refactoring.

DelevoXDG (approved):
lgtm 🔥

PR #1226, opened 2023-11-20
Remove Gateway from docs and update client models

ddoktorski:
Related to #1047 

## Introduced changes

- Remove `GatewayClient` from documentation and gateway classes from `client_models`
- Remove `net` property
- Change type of `execution_resources` in `TransactionReceipt` from `dict` to `ExecutionResources` class

##

### DIFF
--- a/starknet_py/net/client.py
+++ b/starknet_py/net/client.py
@@ -9,7 +9,6 @@ from starknet_py.net.client_errors import ClientError
 from starknet_py.net.client_models import (
     BlockStateUpdate,
     BlockTransactionTrace,
-    BlockTransactionTraces,
     Call,
     ContractClass,
     DeclareTransactionResponse,
@@ -73,7 +72,7 @@ class Client(ABC):
         self,
         block_hash: Optional[Union[Hash, Tag]] = None,
         block_number: Optional[Union[int, Tag]] = None,
-    ) -> Union[BlockTransactionTraces, List[BlockTransactionTrace]]:
+    ) -> List[BlockTransactionTrace]:
         """
         Receive the traces of all the transactions within specified block
 
--- a/starknet_py/net/client.py
+++ b/starknet_py/net/client.py
@@ -32,7 +32,6 @@ from starknet_py.net.models.transaction import (
     DeployAccount,
     Invoke,
 )
-from starknet_py.net.networks import Network
 from starknet_py.transaction_errors import (
     TransactionNotReceivedError,
     TransactionRejectedError,
@@ -43,16 +42,6 @@ from starknet_py.utils.sync import add_sync_methods
 
 @add_sync_methods
 class Client(ABC):
-    @property
-    @abstractmethod
-    def net(self) -> Network:
-        """
-        Network of the client.
-
-         .. deprecated:: 0.15.0
-            Property net of the Client interface is deprecated.
-        """
-
     @abstractmethod
     async def get_block(
         self,
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -6,7 +6,7 @@ They need to stay backwards compatible for old transactions/blocks to be fetchab
 from abc import ABC
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Any, Dict, Iterable, List, Optional, Union
+from typing import Any, Iterable, List, Optional, Union
 
 from typing_extensions import Literal
 
@@ -49,19 +49,6 @@ class EventsChunk:
 
     events: List[Event]
     continuation_token: Optional[str] = None
-
-
-@dataclass
-class L1toL2Message:
-    """
-    Dataclass representing a L1->L2 message.
-    """
-
-    payload: List[int]
-    nonce: int
-    selector: int
-    l1_address: int
-    l2_address: int
 
 
 @dataclass
@@ -158,7 +145,6 @@ class DeployTransaction(Transaction):
     Dataclass representing deploy transaction.
     """
 
-    contract_address: Optional[int]  # Gateway-only field, hence Optional
     contract_address_salt: int
     constructor_calldata: List[int]
     class_hash: int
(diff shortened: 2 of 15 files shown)


starknet_py/net/client.py
```diff
@@ -44,16 +42,6 @@
...
-    @property
-    @abstractmethod
-    def net(self) -> Network:
-        """
-        Network of the client.
-
-         .. deprecated:: 0.15.0
-            Property net of the Client interface is deprecated.
-        """
-
```
tomek0123456789, line 56: Needs information in migration guide about that removal

tomek0123456789 (commented):
image

starknet_py/net/client_models.py
```diff
@@ -226,7 +212,26 @@ class TransactionFinalityStatus(Enum):
...
+    pedersen_builtin_applications: int
+    poseidon_builtin_applications: int
+    ec_op_builtin_applications: int
+    ecdsa_builtin_applications: int
+    bitwise_builtin_applications: int
+    keccak_builtin_applications: int
+    memory_holes: Optional[int] = None
+
+
+# TODO (#1179): split into PendingTransactionReceipt and TransactionReceipt
```
DelevoXDG, line 234: Are you sure this is correct issue no? 👀

starknet_py/net/client_models.py
```diff
@@ -239,10 +244,9 @@ class TransactionReceipt:
     events: List[Event] = field(default_factory=list)
     l2_to_l1_messages: List[L2toL1Message] = field(default_factory=list)
```
DelevoXDG, line 245: Maybe change this to `messages_sent` so it matches the RPC spec?

starknet_py/net/client_models.py
```diff
@@ -239,10 +244,9 @@ class TransactionReceipt:
...
 
-    execution_status: Optional[
-        TransactionExecutionStatus
-    ] = None  # gateway/pending receipt field
+    execution_status: Optional[TransactionExecutionStatus] = None
     finality_status: Optional[TransactionFinalityStatus] = None
+    # TODO (#1179): remove status field once devnet is updated
```
DelevoXDG, line 249: Since you are already making changes to the receipt and removing gateway fields, maybe we could make execution and finality statuses non-optional as per the spec, or is there something blocking this?

starknet_py/net/client_models.py
```diff
@@ -253,20 +257,13 @@ class TransactionReceipt:
     block_number: Optional[int] = None
     block_hash: Optional[int] = None
     actual_fee: int = 0
-    # TODO (#1047): change that into ExecutionResources class after gateway removal
-    #  (values of course differ for each client)
     # TODO (#1179): this field should be required
-    execution_resources: Optional[dict] = field(default_factory=dict)
+    execution_resources: Optional[ExecutionResources] = None
```
DelevoXDG, line 261: Same as above

starknet_py/net/client_models.py
```diff
@@ -253,20 +257,13 @@ class TransactionReceipt:
...
 
     message_hash: Optional[int] = None  # L1_HANDLER_TXN_RECEIPT-only
 
     rejection_reason: Optional[str] = None
```
DelevoXDG, line 265: Since we are removing gateway-only `revert_error`, maybe remove `rejection_reason` as well, or do we have a reason to keep it?

starknet_py/net/client_models.py
```diff
@@ -742,56 +634,6 @@ class TransactionStatusResponse:
...
-    signature_input: SignatureInput
-
-
-class DAMode(Enum):
-    """
-    Enum specifying a storage domain in Starknet. Each domain has different gurantess regarding availability.
-    """
-
-    L1 = "L1"
-    L2 = "L2"
```
DelevoXDG, line 772: Might need this for RPC 0.6.0, although it can probably be added back when needed 😅

docs/migration_guide.rst
```diff
@@ -1,11 +1,29 @@
...
+**********************
+
+0.19.0 Breaking changes
+-----------------------
+
+.. currentmodule:: starknet_py.net.client_models
+
+1. :class:`GatewayClient` and all related classes have been removed.
+2. Client ``net`` property has been removed.
+3. :class:`TransactionReceipt` field ``execution_resources`` has been changed from ``dict`` to :class:`ExecutionResources`.
```
DelevoXDG, line 15: Maybe we should additionally have info about fields removed from `TransactionReceipt`?

Alternatively: 
> 1. :class:`GatewayClient` and all related classes and fields have been removed.

starknet_py/net/client_models.py
ddoktorski: Good point! I am fine to remove it for now though as it not a problem to add it back later if necessary.

starknet_py/net/client_models.py
ddoktorski: That was the plan at the beginning, but since we are still using the old devnet with RPC v0.3.0 it would require some additional work to make those changes now. @Radinyn is already working on updating devnet and once it is done we will do a proper refactoring that will include changes that you mentioned. Apologies for not making this clear in this PR.

starknet_py/net/client_models.py
ddoktorski: Same response as [here](https://github.com/software-mansion/starknet.py/pull/1226/files#r1405971792)

starknet_py/net/client_models.py
ddoktorski: Same response as [here](https://github.com/software-mansion/starknet.py/pull/1226/files#r1405971792)

starknet_py/net/client_models.py
ddoktorski: I decided to put this issue here as these changes will be easier to do after updating devnet. Maybe we should consider creating a separate issue though.

docs/migration_guide.rst
ddoktorski: Changed to your suggestion. There will be more information added to the migration guide after models refactoring that will take place before next release.

DelevoXDG (approved):
image

PR #1239, opened 2023-12-18
Update `starknet-devnet`

### LINKED ISSUE #1179: Update to `devnet-rs`
- moving and refactoring tests from file `/tests/e2e/tests_on_networks/` to proper other files (as those won't need nodes anymore, just devnet)

ddoktorski:
Closes #1179 
Related to #1219 

## Introduced changes

- Use `starknet-devnet-rs` with RPC v0.5.1
- Skip some failing tests / move them to integration networks
- Slight refactoring of models (mostly related to `TransactionReceipt`) and some tests
- Change logic in `wait_for_tx` function

Failing tests that require further investigation:
- `contract_interaction/v1_interaction_test.py` fail for cairo 1 contracts compiled with compiler version 1.x. It seems to be a problem on devnet side
- `tests_on_networks/account_test.py` are successful only when run one by one manually, they are moved to integration only because they fail on devnet
- ~Other failing tests on network, not sure if this is related to our node or to some recent changes on integration/testnet~

##


ddoktorski:
Windows tests checked in my fork repository -> https://github.com/ddoktorski/starknet.py/actions/runs/7246444653

.github/workflows/checks.yml
```diff
@@ -1,5 +1,10 @@
 name: Checks
 
+env:
+  STARKNET_VERSION: "0.12.3"
+  RPC_SPEC_VERSION: "0.5.1"
+  DEVNET_SHA: "78527de"
+
```
THenry14, line 7: how do we set this locally, if someone wants to install devnet?

.github/workflows/checks.yml
```diff
@@ -206,6 +171,14 @@ jobs:
         run: |
           poetry install --without py39-dev
 
+      # ====================== SETUP DEVNET ====================== #
+
+      - name: Install devnet
+        run: |
+          cargo install --locked \
+          --git https://github.com/0xSpaceShard/starknet-devnet-rs.git \
+          --rev ${{ env.DEVNET_SHA }}
```
THenry14, line 180: It may be a good idea to set `--root` to better control where the devnet is installed

starknet_py/net/client.py
```diff
@@ -159,20 +157,13 @@ async def wait_for_tx(
             try:
                 tx_receipt = await self.get_transaction_receipt(tx_hash=tx_hash)
```
THenry14, line 158: maybe we could use `getTransactionStatus` here instead?

THenry14 (commented):
generally this looks good to me

.github/workflows/checks.yml
ddoktorski: At this moment all the installation is on the developer side, tests will use `starknet-devnet` that is added to `PATH` (or fail if it is not found)

THenry14: yup, but how should the developer know which hash/rpc/starknet version to use? is it docummented somewhere? (sorry if it is, I couldn't find it)

.github/workflows/checks.yml
ddoktorski: I was initially doing it this way, but given that it works very well without `--root` I would leave it as it is. If we want to change it at any point, it would be a quick fix.

THenry14: sure, this is a nitpick, but please remember that this will mean that if someone modifies cwds at any point in the future, the tests may break

starknet_py/net/client.py
ddoktorski: The difference is that for `getTransactionStatus` field `execution_status` is optional, but for `getTransactionReceipt` both `execution_status` and `finality_status` are required. I think it might change the logic how this function works, so I would prefer not to change it.

THenry14: I think this is because `status` can be also used in not received/rejected txs that probably won't have execution_status (but don't quote me on that - let's double check). If that's the case, this is another reason to use status endpoint instead

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
```diff
@@ -6,6 +6,7 @@
 from starknet_py.tests.e2e.fixtures.contracts import deploy_v1_contract
```
abulenok, line 6: > contract_interaction/v1_interaction_test.py fail for cairo 1 contracts compiled with compiler version 1.x. It seems to be a problem on devnet side

Did you verify this on integration node?

ddoktorski: I haven't tried to run these tests on integration, but the exact error is related to `kind` field in Sierra compiled files: ``Invalid declare transaction v2: missing field `kind`.``
 
Indeed, cairo 1 contracts compiled with compiler v1 do not have this field, but they can be declared using DeclareV2 transaction.

starknet_py/net/http_client.py
```diff
@@ -63,10 +63,12 @@ async def call(self, method_name: str, params: dict):
         payload = {
             "jsonrpc": "2.0",
             "method": f"starknet_{method_name}",
-            "params": params,
             "id": 0,
         }
 
+        if params:
+            payload["params"] = params
+
```
DelevoXDG, line 71: Not sure about this change. Yes, it can be omitted from the json-rpc point of view. 
In the starknet spec, however, if a method doesn't require parameters, an empty array is explicitly included in the "params" field (I wasn't able to find an exception):
```json
"params": [],
```
And here, if the params dict is empty, nothing will be passed. Maybe nodes accept that currently, but I think it would be better to pass empty array if params dict is empty just in case, wdyt?

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
DelevoXDG: Can you please explain what part of starknet.py fails in these tests exactly? Just to make sure the problem is on devnet side?

starknet_py/net/http_client.py
ddoktorski: Thanks! Sounds good to me.

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -447,45 +443,8 @@ async def test_deploy_account_uses_custom_calldata(
...
-    with pytest.raises(ClientError):
-        await account.client.declare(estimate_fee_transaction)
-
-    # Verify that original transaction can be sent
-    result = await account.client.declare(transaction)
-    await account.client.wait_for_tx(result.transaction_hash)
-
-
+# TODO (#1219): investigate why this test fails
+@pytest.mark.skip
```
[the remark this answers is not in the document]
ddoktorski: The problem with this test is that it doesn't throw any error. We are trying to deploy account using a signed transaction for fee estimation and it should throw an error.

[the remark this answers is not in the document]
DelevoXDG: Strange, for me it's the other way around, it's throwing error when estimating fee with estimate version 😄

starknet_py/tests/e2e/contract_interaction/v1_interaction_test.py
ddoktorski: `addDeclareTransaction` fails for Cairo 1 contract compiled with compiler v1 with the error:
```
{'error': {'code': -32602, 'message': 'Invalid declare transaction v2: missing field `kind`'}, 'id': 0, 'jsonrpc': '2.0'}
```
@DelevoXDG  Have you noticed anything similar in jvm?

ddoktorski: Actually some of those tests pass and it also depends on the specific contract. For some contracts without `kind` field it throws the error and for other it does not :confused: In the current version I added skip condition and reason only for the tests that fail.

abulenok: Is it possible to execute these tests on integration net instead of devnet?

DelevoXDG: > Have you noticed anything similar in jvm?

Not really, declare with cairo 1 contract code seems to work fine on devnet with rpc 0.6.0 (`85495efb71a37ad3921c8986474b7e78a9a9f5fc`) in jvm tests.
Can you send the request the full `addDeclareTransaction ` request that gets sent maybe?

starknet_py/net/client_models.py
```diff
@@ -241,28 +235,22 @@ class TransactionReceipt:
...
     messages_sent: List[L2toL1Message] = field(default_factory=list)
 
-    execution_status: Optional[TransactionExecutionStatus] = None
-    finality_status: Optional[TransactionFinalityStatus] = None
-    # TODO (#1179): remove status field once devnet is updated
-    status: Optional[
-        TransactionStatus
-    ] = None  # replaced by execution and finality status in RPC v0.4.0-rc1
-
     type: Optional[TransactionType] = None
```
THenry14, line 245: nit: I believe `type` is now always reqiured for tx receipts, but I guess this is out of scope for this PR. let's remember to re-check fields in time for 0.6.0 update

starknet_py/net/client.py
```diff
@@ -155,30 +166,37 @@ async def wait_for_tx(
...
+                    if (
+                        tx_receipt.execution_status
+                        == TransactionExecutionStatus.SUCCEEDED
+                        or tx_receipt.finality_status
+                        in (
+                            TransactionFinalityStatus.ACCEPTED_ON_L2,
+                            TransactionFinalityStatus.ACCEPTED_ON_L1,
+                        )
+                    ):
+                        return tx_receipt
```
DelevoXDG, line 198: `finality_status` is a mandatory property in `tx_receipt` with only two possible values: `ACCEPTED_ON_L2` and `ACCEPTED_ON_L1`. As a result, the given condition will always evaluate to true.

Since you raise the error if the transaction is `REVERTED` above, I don't really see the need for this check:
```suggestion
                    else:
                        return tx_receipt
```
Additionally, you could introduce `isAccepted`/`isSuccessful`/`isReverted` method in `tx_receipt` to encapsulate the condition that makes transaction "unsuccessful".

starknet_py/net/client_models.py
DelevoXDG: I think it would make sense to add this change now, since changing `tx_receipt` is already in the scope of this PR and we made every other required property non-optional.

But, if this would take much time and stale this PR, I don't mind pushing it back, so it's up to you 😄

starknet_py/net/client.py
```diff
@@ -155,30 +166,37 @@ async def wait_for_tx(
                 " block have status ACCEPTED_ON_L2."
             )
 
+        transaction_received = False
         while True:
+            retries -= 1
```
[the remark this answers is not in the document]
ddoktorski: I see your point and agree. However, I think we might still need another loop to get transaction receipt to be sure that it won't break. Will keep this in mind for a future refactoring, but not a top priority for now.

DelevoXDG (approved):
image

THenry14 (approved):
💪

PR #1242, opened 2023-12-26
feat: add block_number to Event and EventSchema

JuArce:
## Introduced changes
- Add block_number to Event
- Add block_number to EventSchema

### DIFF
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -39,6 +39,7 @@ class Event:
     from_address: int
     keys: List[int]
     data: List[int]
+    block_number: int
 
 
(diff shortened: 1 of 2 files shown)


ddoktorski:
Hi @JuArce Thank you for your contribution. This change will require a little bit more work:
- Revise the changes to better reflect the [RPC spec](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_api_openrpc.json#L859) (e.g. also include `transaction_hash`)
- Addition of some tests to verify that the new fields are correctly returned would be beneficial

Are you open to continue working on these changes?

PR #1165, opened 2023-08-23
add cairo_version to execute to support cairo1 calls encoding

CremaFR:
Closes #1164

## Introduced changes

- add cairo_version to execute to support cairo1 calls encoding

##

- [ ] This PR contains breaking changes

PR #1243, opened 2024-01-03
Support RPC v0.6.0

### LINKED ISSUE #1235: Update RPC to v0.6.0
https://github.com/starkware-libs/starknet-specs/releases/tag/v0.6.0

ddoktorski:
Closes #1235 

## Introduced changes

- Classes for V3 transactions, currently we are using separate classes for input and output to the RPC provider. The modifications in this PR follow this approach to avoid extensive changes unrelated to RPC v0.6.0. However, this aspect should be addressed during the next refactoring
- Classes added to `client_models` to mirror the new structs added to the RPC spec
    - `ResourceBoundsMapping`, `PriceUnit`, `FeePayment`, `DAMode`
    - Abstract class `TransactionV3` with common V3 fields
- Schemas for new transaction in `rpc.py` that uses `OneOfSchema`, this approach aligns more closely with RPC spec structures, as opposed to the earlier use of `Optional` fields
- Methods for calculating V3 transactions hashes and helper class `CommonTransactionV3Fields`
- `Account`
    -  Methods for signing v3 transactions and `execute_v3` method
- `StarkCurveSigner`
    - In `sign_transaction` remove specific logic for hash calculation for every transaction type, use abstract method `calculate_hash` from `AccountTransaction`
- `FullNodeClient`
    - `estimate_fee` method now includes `skip_validate`
    - `send_transaction`, `deploy_account` and `declare` have been adapted to alight with V3 transactions
    - Helper methods for creating broadcasted transactions V3 <- Note: Since only the declare transaction has a specific broadcasted structure, it suggests potential for simplifying these methods
    - Some utility methods have been moved to `client_utils.py`

##

### DIFF
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -16,13 +16,22 @@ Module containing base models and functions to operate on them.
 .. autoclass:: DeployAccountV1
     :exclude-members: __init__, __new__
 
+.. autoclass:: DeployAccountV3
+    :exclude-members: __init__, __new__
+
 .. autoclass:: DeclareV1
     :exclude-members: __init__, __new__
 
 .. autoclass:: DeclareV2
     :exclude-members: __init__, __new__
 
+.. autoclass:: DeclareV3
+    :exclude-members: __init__, __new__
+
 .. autoclass:: InvokeV1
+    :exclude-members: __init__, __new__
+
+.. autoclass:: InvokeV3
     :exclude-members: __init__, __new__
 
 .. autoenum:: StarknetChainId
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -13,16 +13,16 @@ Module containing base models and functions to operate on them.
     :exclude-members: __init__, __new__
     :members:
 
-.. autoclass:: DeployAccount
+.. autoclass:: DeployAccountV1
     :exclude-members: __init__, __new__
 
-.. autoclass:: Declare
+.. autoclass:: DeclareV1
     :exclude-members: __init__, __new__
 
 .. autoclass:: DeclareV2
     :exclude-members: __init__, __new__
 
-.. autoclass:: Invoke
+.. autoclass:: InvokeV1
     :exclude-members: __init__, __new__
 
 .. autoenum:: StarknetChainId
--- a/starknet_py/hash/transaction.py
+++ b/starknet_py/hash/transaction.py
@@ -1,12 +1,24 @@
+from dataclasses import dataclass
 from enum import IntEnum
-from typing import Optional, Sequence
+from typing import List, Optional, Sequence
 
+from poseidon_py.poseidon_hash import poseidon_hash_many
+
+from starknet_py.cairo.felt import encode_shortstring
 from starknet_py.common import int_from_bytes
 from starknet_py.constants import DEFAULT_ENTRY_POINT_SELECTOR
 from starknet_py.hash.class_hash import compute_class_hash
 from starknet_py.hash.sierra_class_hash import compute_sierra_class_hash
 from starknet_py.hash.utils import compute_hash_on_elements
-from starknet_py.net.client_models import ContractClass, SierraContractClass
+from starknet_py.net.client_models import (
+    ContractClass,
+    DAMode,
+    ResourceBoundsMapping,
+    SierraContractClass,
+)
+
+L1_GAS_ENCODED = encode_shortstring("L1_GAS")
+l2_GAS_ENCODED = encode_shortstring("L2_GAS")
 
 
 class TransactionHashPrefix(IntEnum):
@@ -19,6 +31,54 @@ class TransactionHashPrefix(IntEnum):
     DEPLOY_ACCOUNT = int_from_bytes(b"deploy_account")
     INVOKE = int_from_bytes(b"invoke")
     L1_HANDLER = int_from_bytes(b"l1_handler")
+
+
+@dataclass
+class CommonTransactionV3Fields:
+    # pylint: disable=too-many-instance-attributes
+
+    tx_prefix: TransactionHashPrefix
+    version: int
+    address: int
+    tip: int
+    resource_bounds: ResourceBoundsMapping
+    paymaster_data: List[int]
+    chain_id: int
+    nonce: int
+    nonce_data_availability_mode: DAMode
+    fee_data_availability_mode: DAMode
+
+    def get_common_tx_fields(self):
+        return [
+            self.tx_prefix,
+            self.version,
+            self.address,
+            self.compute_tip_resource_bounds_hash(),
+            poseidon_hash_many(self.paymaster_data),
+            self.chain_id,
+            self.nonce,
+            self.get_data_availability_modes(),
+        ]
+
+    def compute_tip_resource_bounds_hash(self) -> int:
+        l1_gas_bounds = (
+            (L1_GAS_ENCODED << (128 + 64))
+            + (self.resource_bounds.l1_gas.max_amount << 128)
+            + self.resource_bounds.l1_gas.max_price_per_unit
+        )
+
+        l2_gas_bounds = (
+            (l2_GAS_ENCODED << (128 + 64))
+            + (self.resource_bounds.l2_gas.max_amount << 128)
+            + self.resource_bounds.l2_gas.max_price_per_unit
+        )
+
+        return poseidon_hash_many([self.tip, l1_gas_bounds, l2_gas_bounds])
+
+    def get_data_availability_modes(self) -> int:
+        return (
+            self.nonce_data_availability_mode.value << 32
+        ) + self.fee_data_availability_mode.value
 
 
 # pylint: disable=too-many-arguments
@@ -110,6 +170,21 @@ def compute_invoke_transaction_hash(
     )
 
 
+def compute_invoke_v3_transaction_hash(
+    *,
+    account_deployment_data: List[int],
+    calldata: List[int],
+    common_fields: CommonTransactionV3Fields,
+) -> int:
+    return poseidon_hash_many(
+        [
+            *common_fields.get_common_tx_fields(),
+            poseidon_hash_many(account_deployment_data),
+            poseidon_hash_many(calldata),
+        ]
+    )
+
+
 def compute_deploy_account_transaction_hash(
     version: int,
     contract_address: int,
@@ -142,6 +217,23 @@ def compute_deploy_account_transaction_hash(
         max_fee=max_fee,
         chain_id=chain_id,
         additional_data=[nonce],
+    )
+
+
+def compute_deploy_account_v3_transaction_hash(
+    *,
+    class_hash: int,
+    constructor_calldata: List[int],
+    contract_address_salt: int,
+    common_fields: CommonTransactionV3Fields,
+) -> int:
+    return poseidon_hash_many(
+        [
+            *common_fields.get_common_tx_fields(),
+            poseidon_hash_many(constructor_calldata),
+            class_hash,
+            contract_address_salt,
+        ]
     )
 
 
@@ -220,4 +312,27 @@ def compute_declare_v2_transaction_hash(
         max_fee=max_fee,
         chain_id=chain_id,
         additional_data=[nonce, compiled_class_hash],
+    )
+
+
+def compute_declare_v3_transaction_hash(
+    *,
+    contract_class: Optional[SierraContractClass] = None,
+    class_hash: Optional[int] = None,
+    account_deployment_data: List[int],
+    compiled_class_hash: int,
+    common_fields: CommonTransactionV3Fields,
+) -> int:
+    if class_hash is None:
+        if contract_class is None:
+            raise ValueError("Either contract_class or class_hash is required.")
+        class_hash = compute_sierra_class_hash(contract_class)
+
+    return poseidon_hash_many(
+        [
+            *common_fields.get_common_tx_fields(),
+            poseidon_hash_many(account_deployment_data),
+            class_hash,
+            compiled_class_hash,
+        ]
     )
--- a/starknet_py/hash/transaction.py
+++ b/starknet_py/hash/transaction.py
@@ -312,13 +312,9 @@ def compute_declare_v2_transaction_hash(
     :param nonce: Nonce of the transaction.
     :return: Hash of the transaction.
     """
-    if contract_class is None and class_hash is None:
-        raise ValueError("Either contract_class or class_hash is required.")
-    if contract_class is not None and class_hash is not None:
-        raise ValueError("Both contract_class and class_hash passed.")
-
     if class_hash is None:
-        assert contract_class is not None
+        if contract_class is None:
+            raise ValueError("Either contract_class or class_hash is required.")
         class_hash = compute_sierra_class_hash(contract_class)
 
     return compute_transaction_hash(
--- a/starknet_py/hash/transaction.py
+++ b/starknet_py/hash/transaction.py
@@ -53,7 +53,7 @@ class CommonTransactionV3Fields:
             self.tx_prefix,
             self.version,
             self.address,
-            poseidon_hash_many(tip, self.compute_resource_bounds_for_fee()),
+            poseidon_hash_many([self.tip, *self.compute_resource_bounds_for_fee()]),
             poseidon_hash_many(self.paymaster_data),
             self.chain_id,
             self.nonce,
@@ -178,7 +178,7 @@ def compute_invoke_v3_transaction_hash(
 ) -> int:
     return poseidon_hash_many(
         [
-            *common_fields.get_common_tx_fields(),
+            *common_fields.compute_common_tx_fields(),
             poseidon_hash_many(account_deployment_data),
             poseidon_hash_many(calldata),
         ]
@@ -229,7 +229,7 @@ def compute_deploy_account_v3_transaction_hash(
 ) -> int:
     return poseidon_hash_many(
         [
-            *common_fields.get_common_tx_fields(),
+            *common_fields.compute_common_tx_fields(),
             poseidon_hash_many(constructor_calldata),
             class_hash,
             contract_address_salt,
@@ -330,7 +330,7 @@ def compute_declare_v3_transaction_hash(
 
     return poseidon_hash_many(
         [
-            *common_fields.get_common_tx_fields(),
+            *common_fields.compute_common_tx_fields(),
             poseidon_hash_many(account_deployment_data),
             class_hash,
             compiled_class_hash,
--- a/starknet_py/hash/transaction.py
+++ b/starknet_py/hash/transaction.py
@@ -53,14 +53,14 @@ class CommonTransactionV3Fields:
             self.tx_prefix,
             self.version,
             self.address,
-            self.compute_tip_resource_bounds_hash(),
+            poseidon_hash_many(tip, self.compute_resource_bounds_for_fee()),
             poseidon_hash_many(self.paymaster_data),
             self.chain_id,
             self.nonce,
             self.get_data_availability_modes(),
         ]
 
-    def compute_tip_resource_bounds_hash(self) -> int:
+    def compute_resource_bounds_for_fee(self) -> List[int]:
         l1_gas_bounds = (
             (L1_GAS_ENCODED << (128 + 64))
             + (self.resource_bounds.l1_gas.max_amount << 128)
@@ -73,7 +73,7 @@ class CommonTransactionV3Fields:
             + self.resource_bounds.l2_gas.max_price_per_unit
         )
 
-        return poseidon_hash_many([self.tip, l1_gas_bounds, l2_gas_bounds])
+        return [l1_gas_bounds, l2_gas_bounds]
 
     def get_data_availability_modes(self) -> int:
         return (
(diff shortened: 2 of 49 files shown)


ddoktorski:
TODO:
- Add `auto_estimate` parameter to `Account` signing methods
- Decide on approach for V3 transaction in `Contract` class and for `deploy_account` static method in `Account`
- Update documentation

starknet_py/net/full_node_client.py
```diff
@@ -703,7 +719,11 @@ async def simulate_transactions(
         # pylint: disable=too-many-arguments
         """
         Simulates a given sequence of transactions on the requested state, and generates the execution traces.
-        If one of the transactions is reverted, raises CONTRACT_ERROR.
+        Note that some of the transactions may revert, in which case no error is thrown, but revert details can be seen
+        on the returned trace object.
+        Note that some of the transactions may revert, this will be reflected by the revert_error property in the trace.
+        Other types of failures (e.g. unexpected error or failure in the validation phase) will result
+        in TRANSACTION_EXECUTION_ERROR.
```
THenry14, line 726: could be consolidated probably?

THenry14 (commented):
partial review.

overall logic-wise this looks good. there are couple of things I did not mention in this part of review that could be done differently, but can easily be refactored at lated date.

Did not look into tests yet, will try to do it first thing tomorrow.

starknet_py/net/full_node_client.py
```diff
@@ -849,74 +871,100 @@ def _create_broadcasted_declare_v2_properties(transaction: DeclareV2) -> dict:
...
-    return hashed_key
-
-
-def _to_rpc_felt(value: Hash) -> str:
-    """
-    Convert the value to RPC felt matching a ``^0x(0|[a-fA-F1-9]{1}[a-fA-F0-9]{0,62})$`` pattern.
+    if _extract_tx_version(transaction.version) < 3 and hasattr(transaction, "max_fee"):
+        broadcasted_txn_common_properties["max_fee"] = _to_rpc_felt(
+            transaction.max_fee  # pyright: ignore
+        )
```
DelevoXDG, line 950: Not necessarily in this PR, but I think it could and should be refactored, as the logic seems very unintuitive to me.
1. This should apply all common properties, but in reality it applies only intersection of v1 and v3 common properties + if needed the entirety of v1 common properties.
2. At the same time, all v3 common properties are actually applied by type-specific functions.
3. I think logic in `_create_broadcasted_txn` should be refactored:
```python
    txn_map = {
        TransactionType.DECLARE: _create_broadcasted_declare_properties,
        TransactionType.INVOKE: _create_broadcasted_invoke_properties,
        TransactionType.DEPLOY_ACCOUNT: _create_broadcasted_deploy_account_properties,
    }

    common_properties = _create_broadcasted_txn_common_properties(transaction)
    transaction_specific_properties = txn_map[transaction.type](transaction)
```
First, we should create all common properties by version, regardless of the type.
Then, we should apply the rest of the properties based both on type and version.

starknet_py/net/models/transaction.py
```diff
@@ -76,13 +86,54 @@ class AccountTransaction(Transaction, ABC):
 TypeAccountTransaction = TypeVar("TypeAccountTransaction", bound=AccountTransaction)
 
 
+@dataclass(frozen=True)
+class DeclareV3(AccountTransaction):
```
DelevoXDG, line 90: Could this dataclass be manually instantiated by a user? If it is, we should probably limit the values that could be passed to a constructor.

starknet_py/net/schemas/rpc.py
```diff
@@ -180,22 +200,54 @@ def make_dataclass(self, data, **kwargs) -> TransactionStatusResponse:
...
+    fee_data_availability_mode = DAModeField(
+        data_key="fee_data_availability_mode", load_default=DAMode.L1
+    )
+    paymaster_data = fields.List(Felt(), data_key="paymaster_data", load_default=[])
+    resource_bounds = fields.Nested(
+        ResourceBoundsMappingSchema(), data_key="resource_bounds", required=True
+    )
+
+
+class InvokeTransactionDeprecatedSchema(TransactionSchema):
```
DelevoXDG, line 249: I think `DeprecatedInvokeTransactionSchema` or `InvokeTransactionV1Schema` would be preferable. As before, adding intermediate `DeprecatedTransactionSchema` would make sense.

starknet_py/tests/e2e/client/fixtures/transactions.py
```diff
@@ -22,13 +22,18 @@
 
 @pytest_asyncio.fixture(scope="package")
 async def deploy_account_transaction(
-    account_with_validate_deploy_class_hash: int, fee_contract: Contract, network: str
+    account_with_validate_deploy_class_hash: int,
+    fee_contract: Contract,
```
DelevoXDG, line 26:
```suggestion
    eth_fee_contract: Contract,
```

starknet_py/tests/e2e/fixtures/contracts.py
```diff
@@ -215,11 +234,34 @@ async def deploy_erc20_contract(
 
 
 @pytest.fixture(scope="package")
-def fee_contract(account: BaseAccount) -> Contract:
+def fee_contract(account: BaseAccount, fee_contract_abi) -> Contract:
```
DelevoXDG, line 237:
```suggestion
def eth_fee_contract(account: BaseAccount, fee_contract_abi) -> Contract:
```

DelevoXDG (commented):
Good logic overall, most suggestions are nitpicks. 
I would refactor logic of creating broadcasted transactions, but it's probably not urgent if it works. Before releasing this, I think it would be nice to get naming in line: add versioning and add intermediate dataclasses for deprecated transactions.

starknet_py/net/client_models.py
```diff
@@ -104,39 +153,79 @@ class Transaction(ABC):
...
+    max_fee: int
     class_hash: int  # Responses to getBlock and getTransaction include the class hash
     sender_address: int
     compiled_class_hash: Optional[int] = None  # only in DeclareV2, hence Optional
     nonce: Optional[int] = None
-    l1_gas: Optional[ResourceLimits] = None  # DECLARE_V3-only field, hence Optional
+
+
+@dataclass
+class DeclareTransactionV3(TransactionV3):
```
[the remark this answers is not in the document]
ddoktorski: Good catch, thanks!

starknet_py/net/full_node_client.py
ddoktorski: Totally agree, thank you for your comment! 

I believe we can also leverage `TypesOfTransactionsSchema().dump(transaction)` here to get broadcasted dict. Only `Declare` may require some additional work. I've created an issue so it won't get lost: https://github.com/software-mansion/starknet.py/issues/1246

starknet_py/tests/e2e/client/fixtures/transactions.py
ddoktorski: Thanks! I will commit it separately as it requires modifications in several other places.

starknet_py/net/client_models.py
```diff
@@ -238,16 +338,16 @@ class TransactionReceipt:
     execution_status: TransactionExecutionStatus
     finality_status: TransactionFinalityStatus
     execution_resources: ExecutionResources
-    type: TransactionType
+    actual_fee: FeePayment
 
     events: List[Event] = field(default_factory=list)
     messages_sent: List[L2toL1Message] = field(default_factory=list)
 
+    type: Optional[TransactionType] = None
```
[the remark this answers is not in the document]
ddoktorski: Thanks! Not sure what happened here, will be fixed.

starknet_py/net/full_node_client.py
ddoktorski: Basically this is copied from [the spec](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_trace_api_openrpc.json#L42), but I can try to make it more concise.

starknet_py/net/account/account.py
```diff
@@ -535,6 +709,26 @@ def _add_max_fee_to_transaction(
...
 
+def _add_resource_bounds_to_transaction(
+    tx: TypeAccountTransaction, resource_bounds: ResourceBoundsMapping
+) -> TypeAccountTransaction:
+    return dataclasses.replace(tx, resource_bounds=resource_bounds)
+
+
+def _parse_calls(cairo_version: int, calls: Calls) -> List[int]:
+    if cairo_version == 1:
+        parsed_calls = _parse_calls_v2(ensure_iterable(calls))
```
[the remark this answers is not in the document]
ddoktorski: I have no idea, to be honest. This was added by an external contributor, #1165 for reference.

starknet_py/hash/transaction.py
```diff
@@ -110,6 +170,21 @@ def compute_invoke_transaction_hash(
     )
 
 
+def compute_invoke_v3_transaction_hash(
+    *,
+    account_deployment_data: List[int],
+    calldata: List[int],
+    common_fields: CommonTransactionV3Fields,
+) -> int:
+    return poseidon_hash_many(
```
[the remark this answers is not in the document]
ddoktorski: Added in this PR.

starknet_py/hash/transaction.py
```diff
@@ -221,3 +313,26 @@ def compute_declare_v2_transaction_hash(
...
+    contract_class: Optional[SierraContractClass] = None,
+    class_hash: Optional[int] = None,
+    account_deployment_data: List[int],
+    compiled_class_hash: int,
+    common_fields: CommonTransactionV3Fields,
+) -> int:
+    if class_hash is None:
+        if contract_class is None:
+            raise ValueError("Either contract_class or class_hash is required.")
+        class_hash = compute_sierra_class_hash(contract_class)
```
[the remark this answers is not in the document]
ddoktorski: Done

starknet_py/net/account/account.py
```diff
@@ -408,6 +568,20 @@ async def execute(
         )
         return await self._client.send_transaction(execute_transaction)
 
+    async def execute_v3(
```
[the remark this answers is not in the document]
ddoktorski: Done

starknet_py/net/schemas/rpc.py
ddoktorski: Added `DeprecatedTransactionSchema` and renamed other deprecated schemas to have format `Deprecated<NAME>TransactionSchema`

starknet_py/net/models/transaction.py
ddoktorski: Added `field(init=False)` to fields that shouldn't be instantiated by a user. I have also created internal class `_AccountTransactionV3` with a helper function `get_common_fields()`

starknet_py/net/client_models.py
```diff
@@ -104,39 +153,79 @@ class Transaction(ABC):
...
+    nonce: int
+
+
 @dataclass
 class DeclareTransaction(Transaction):
     """
     Dataclass representing declare transaction.
     """
 
+    max_fee: int
```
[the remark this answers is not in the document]
ddoktorski: Added `DeprecatedTransaction`

starknet_py/net/schemas/rpc.py
```diff
@@ -244,7 +322,54 @@ def make_dataclass(self, data, **kwargs) -> DeployAccountTransaction:
...
+    nonce = Felt(data_key="nonce", required=True)
+
+    @post_load
+    def make_dataclass(self, data, **kwargs) -> DeployAccountTransactionV3:
+        return DeployAccountTransactionV3(**data)
+
+
+class DeclareTransactionSchema(OneOfSchema):
+    type_schemas = {
+        0: DeclareTransactionDeprecatedSchema,
```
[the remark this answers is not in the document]
ddoktorski: Issue to track this: #1249

docs/api/models.rst
```diff
@@ -13,16 +13,16 @@ Module containing base models and functions to operate on them.
...
 
-.. autoclass:: Declare
+.. autoclass:: DeclareV1
     :exclude-members: __init__, __new__
 
 .. autoclass:: DeclareV2
     :exclude-members: __init__, __new__
 
-.. autoclass:: Invoke
+.. autoclass:: InvokeV1
```
THenry14, line 25: Other (new) models will be added in separate PR, right?

starknet_py/net/full_node_client.py
THenry14: Let's, this reads quite badly imo

starknet_py/net/account/account.py
```diff
@@ -27,8 +29,11 @@
     AccountTransaction,
     Declare,
```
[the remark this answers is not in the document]
THenry14: Imo yes, just to conform to the current approach, but I'd suggest more general interfaces overhaul, as I find it too complex currently (eg, we, as users, should be able to just `sign` a tx and not care about its version. the sign function should be smart enough). ofc this is an issue that should rather be addresses in another issue/PR

docs/api/models.rst
ddoktorski: Correct, it will be done with other documentation updates related to RPC v0.6.0.

starknet_py/net/full_node_client.py
```diff
@@ -849,74 +867,100 @@ def _create_broadcasted_declare_v2_properties(transaction: DeclareV2) -> dict:
...
         ],
         "class_hash": _to_rpc_felt(transaction.class_hash),
     }
+
+    if isinstance(transaction, DeployAccountV3):
+        return {
+            **_create_broadcasted_txn_v3_common_properties(transaction),
+            **deploy_account_txn_properties,
+        }
+
```
DelevoXDG, line 931: Nit: From future refactoring pov, I think using different approaches in all three functions that do a similar thing is a bit sub-optimal.

starknet_py/net/models/transaction.py
```diff
@@ -76,13 +87,69 @@ class AccountTransaction(Transaction, ABC):
 TypeAccountTransaction = TypeVar("TypeAccountTransaction", bound=AccountTransaction)
 
 
+@dataclass(frozen=True)
+class _AccountTransactionV3(AccountTransaction, ABC):
```
DelevoXDG, line 91: Why do we start the class name with an underscore?
If it means it's some sort of internal class, I assume `AccountTransaction` should also be `_AccountTransaction`?

starknet_py/net/models/transaction.py
```diff
@@ -150,18 +218,49 @@ def calculate_hash(self, chain_id: StarknetChainId) -> int:
...
+
+
+@dataclass(frozen=True)
+class DeployAccountV1(AccountTransaction):
     """
     Represents a transaction in the Starknet network that is a deployment of a Starknet account
     contract.
     """
 
+    max_fee: int = field(metadata={"marshmallow_field": Felt()})
```
DelevoXDG, line 258: Could we do the same thing as in other places, create intermediate `DeprecatedAccountTransaction` and move `max_fee` there?

starknet_py/net/models/transaction.py
```diff
@@ -220,10 +343,14 @@ def calculate_hash(self, chain_id: StarknetChainId) -> int:
...
-DeclareSchema = marshmallow_dataclass.class_schema(Declare)
+Declare = Union[DeclareV1, DeclareV2, DeclareV3]
+DeployAccount = Union[DeployAccountV1, DeployAccountV3]
+Invoke = Union[InvokeV1, InvokeV3]
+
+InvokeSchema = marshmallow_dataclass.class_schema(InvokeV1)
+DeclareSchema = marshmallow_dataclass.class_schema(DeclareV1)
 DeclareV2Schema = marshmallow_dataclass.class_schema(DeclareV2)
-DeployAccountSchema = marshmallow_dataclass.class_schema(DeployAccount)
+DeployAccountSchema = marshmallow_dataclass.class_schema(DeployAccountV1)
```
DelevoXDG, line 353: Why not add V1 to those?

starknet_py/net/schemas/rpc.py
```diff
@@ -180,22 +200,57 @@ def make_dataclass(self, data, **kwargs) -> TransactionStatusResponse:
...
+    fee_data_availability_mode = DAModeField(
+        data_key="fee_data_availability_mode", load_default=DAMode.L1
+    )
+    paymaster_data = fields.List(Felt(), data_key="paymaster_data", load_default=[])
+    resource_bounds = fields.Nested(
+        ResourceBoundsMappingSchema(), data_key="resource_bounds", required=True
+    )
+
+
+class DeprecatedInvokeTransactionSchema(DeprecatedTransactionSchema):
```
DelevoXDG, line 253: I still think `InvokeTransactionSchemaV1` is a better name as it's more precise, is there a particular reason why you opted for `Deprecated` prefix?

starknet_py/net/models/transaction.py
ddoktorski: I agree that there is some inconsistency. However, `AccountTransaction` is utilisied in `Client` methods for all types of transaction, while `_AccountTransactionV3` is a helper class added to reduce redundant code and I don't want to expose it to users. I propose either renaming `_AccountTransactionV3` or introducing another internal class `_DeprecatedAccountTransaction` or `_AccountTransactionV1` (not sure how to handle `v2` then, since  `_AccountTransactionV2` would be identical to `_AccountTransactionV1`).

starknet_py/net/client_models.py
```diff
@@ -69,18 +69,67 @@ class ResourcePrice:
     """
 
     price_in_wei: int
-    price_in_strk: Optional[int]
+    price_in_fri: Optional[int]
```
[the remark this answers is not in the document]
ddoktorski: That's true, thanks!

starknet_py/net/models/transaction.py
ddoktorski: First I saw [this comment](https://github.com/software-mansion/starknet.py/pull/1243/files#r1446337115). It sounds good to me.

starknet_py/net/schemas/rpc.py
ddoktorski: I agree, but we would need to split this into `InvokeTransactionV0Schema` and `InvokeTransactionV1Schema`. I am fine with doing this change, also for other deprecated schemas, but can it be addressed in #1249?

starknet_py/net/client_models.py
```diff
@@ -114,7 +161,37 @@ def __post_init__(self):
...
+    nonce_data_availability_mode: DAMode
+    fee_data_availability_mode: DAMode
+
+    def __post_init__(self):
+        if self.__class__ == TransactionV3:
+            raise TypeError("Cannot instantiate abstract TransactionV3 class.")
+
+
+@dataclass
+class InvokeTransaction(DeprecatedTransaction):
```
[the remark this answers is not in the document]
ddoktorski: The same comment as [here](https://github.com/software-mansion/starknet.py/pull/1243/files#r1447128899).

starknet_py/net/schemas/rpc.py
DelevoXDG: Makes sense, you can do this in a separate PR if you prefer not to clutter this one

starknet_py/net/models/transaction.py
DelevoXDG: > I agree that there is some inconsistency. However, `AccountTransaction` is utilisied in `Client` methods for all types of transaction, while `_AccountTransactionV3` is a helper class added to reduce redundant code and I don't want to expose it to users.

Yes, the inconsistency is what alerts me a bit, I didn't find any other case where class is named that way. With that said, you argument makes sense. I don't know if the class is necessarily something that should be "hidden" from the users, so it's up to you. 

> introducing another internal class `_DeprecatedAccountTransaction` or `_AccountTransactionV1` (not sure how to handle `v2` then, since `_AccountTransactionV2` would be identical to `_AccountTransactionV1`).

`_DeprecatedAccountTransaction` should be good for both v1 and v2.

starknet_py/net/account/account.py
```diff
@@ -408,6 +568,20 @@ async def execute(
         )
         return await self._client.send_transaction(execute_transaction)
 
+    async def execute_v3(
+        self,
+        calls: Calls,
+        resource_bounds: ResourceBoundsMapping,
```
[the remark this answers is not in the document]
ddoktorski: Done

THenry14 (approved):
lgtm

obraz

starknet_py/net/account/base_account.py
```diff
@@ -197,24 +197,24 @@ async def sign_declare_v3_transaction(
...
-        Create and sign declare transaction using sierra contract.
+        Create and sign declare transaction version 3 using sierra contract.
 
         :param compiled_contract: string containing a compiled Starknet contract.
             Supports new contracts (compiled to sierra).
         :param compiled_class_hash: a class hash of the sierra compiled contract used in the declare transaction.
             Computed from casm compiled contract.
         :param nonce: Nonce of the transaction.
-        :param resource_bounds: Max amount of Wei or Fri to be paid when executing transaction.
+        :param l1_resource_bounds: Max amount and max price per unit of L1 gas used in this transaction.
```
DelevoXDG, line 211: Nit: it could also be just `of Wei` as for now.

starknet_py/net/full_node_client.py
```diff
@@ -46,7 +46,7 @@
 from starknet_py.net.models.transaction import (
     AccountTransaction,
     Declare,
-    DeclareSchema,
+    DeclareV1Schema,
```
DelevoXDG, line 49: Sorry, but it gets confusing, why do we add v1 to declare schema and not deploy account and invoke?

DelevoXDG (approved):
lgtm 🔥

PR #1247, opened 2024-01-05
Add `Uint64` and `Uint128` classes

ddoktorski:
Related to #1235  and #1243 

## Introduced changes

- Introduce classes `Uint64` and `Uint128` used for correct serialization/deserialization of RPC `u64` and `u128` types
- Add `NumberAsHex` class, from which the `Felt`, `Uint64` and `Uint128` classes can inherit. As a result, regex pattern checking has now been added to the `Felt` type
- Use `TransactionV3Schema` to create `dict` of common fields for broadcasted transaction v3, ideally we should implement similar approach for all transactions

##

### DIFF
--- a/starknet_py/net/schemas/common.py
+++ b/starknet_py/net/schemas/common.py
@@ -30,8 +30,15 @@ class Felt(fields.Field):
     Field that serializes int to felt (hex encoded string)
     """
 
+    REGEX_PATTERN = r"^0x(0|[a-fA-F1-9]{1}[a-fA-F0-9]{0,62})"
+
     def _serialize(self, value: Any, attr: str, obj: Any, **kwargs):
-        return hex(value)
+        if isinstance(value, int):
+            serialized = hex(value)
+            if re.fullmatch(self.REGEX_PATTERN, serialized) is not None:
+                return serialized
+
+        raise ValidationError(f"Invalid value provided for Felt: {value}.")
 
     def _deserialize(
         self,
@@ -44,12 +51,12 @@ class Felt(fields.Field):
             return value
 
         if not isinstance(value, str) or not value.startswith("0x"):
-            raise ValidationError(f"Invalid value provided for felt: {value}.")
+            raise ValidationError(f"Invalid value provided for Felt: {value}.")
 
         try:
             return int(value, 16)
         except ValueError as error:
-            raise ValidationError("Invalid felt.") from error
+            raise ValidationError("Invalid Felt.") from error
 
 
 class Uint64(fields.Field):
--- a/starknet_py/net/schemas/common.py
+++ b/starknet_py/net/schemas/common.py
@@ -1,4 +1,5 @@
 import re
+import sys
 from typing import Any, Mapping, Optional, Union
 
 from marshmallow import Schema, ValidationError, fields, post_load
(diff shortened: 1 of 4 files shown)


starknet_py/net/schemas/common.py
```diff
@@ -44,12 +51,67 @@ def _deserialize(
...
+        return isinstance(value, int) and 0 <= value < self.MAX_VALUE
+
+    def _is_str_and_valid_pattern(self, value: Any) -> bool:
+        return (
+            isinstance(value, str)
+            and re.fullmatch(self.REGEX_PATTERN, value) is not None
+        )
+
+
+class Uint128(Uint64):
```
THenry14, line 108: Maybe it would make sense to generalize what's in Uint64 class, and have Uint64 and Uint128 classes directly inherit from this new class? I guess it might be a nitpick thoug

starknet_py/net/schemas/common_test.py
```diff
@@ -1,55 +1,170 @@
...
-    class SchemaWithFelt(Schema):
-        value1 = Felt(data_key="value1")
+class SchemaWithDAModeField(Schema):
+    value = DAModeField(data_key="value")
 
-    data = {"value1": None}
-    with pytest.raises(TypeError):
+
+def test_serialize_felt():
+    data = {"value": 2137}
```
THenry14, line 35: 😅

THenry14 (commented):
lgtm, very minor comments

starknet_py/net/schemas/common.py
```diff
@@ -30,8 +30,15 @@ class Felt(fields.Field):
...
     """
 
+    REGEX_PATTERN = r"^0x(0|[a-fA-F1-9]{1}[a-fA-F0-9]{0,62})"
+
     def _serialize(self, value: Any, attr: str, obj: Any, **kwargs):
-        return hex(value)
+        if isinstance(value, int):
+            serialized = hex(value)
+            if re.fullmatch(self.REGEX_PATTERN, serialized) is not None:
+                return serialized
```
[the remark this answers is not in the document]
ddoktorski: The primary intent behind this was to avoid double conversion to hex and perform this conversion only when we know that the value is `int`. We can indeed implement the same logic as follows:
```
if isinstance(value, int) and re.fullmatch(self.REGEX_PATTERN, hex(value)):
    return hex(value)
```
Nevertheless, in such cases I personally prefer to explicitly check `is not None`. This is because omitting this condition might suggest that `fullmatch` returns `bool`, which isn't the case.

starknet_py/net/schemas/common.py
ddoktorski: Maybe it would even make sense to have a common abstract class for `Felt`, `Uint64` and `Uint128`?

starknet_py/net/schemas/common.py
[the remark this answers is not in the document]
THenry14: Sure, makes sense, though this is not a part of any external interface, right? So I wouldn't be too bothered about this check, and in this context it does not really matter what fullmatch returns - as long as we can easily verify if there was a hit or not. I get the double conversion bit, this is not ideal.

This is a minor thing, I leave it up to you to decide.

starknet_py/net/schemas/common.py
THenry14: If the interface is the same - yes

PR #1255, opened 2024-01-16
Support V3 transactions in the `Contract` class

ddoktorski:
Related to #1235 

## Introduced changes

- Add parameters `tx_version` and `l1_resource_bounds` to:
  - `ContractFunction.invoke()`, `ContractFunction.prepare()`
  - `PreparedFunctionCall`, `PreparedFunctionCall.invoke()`, `PreparedFunctionCall.estimate_fee()` (only `tx_version`)
  - `Contract.declare()`, `Contract.deploy_contract()`
  - `DeclareResult.deploy()`
- Add optional parameters `salt` and `unique` to static method `Contract.deploy_contract()` and use this function in `DeclareResult.deploy()`
- Add `declare_transaction` field to `DeclareResult`
- Increase Pylint `max-args` from 5 to 6 to avoid writing `pylint: disable=too-many-arguments` in certain functions. It is likely that we will need to further increase this value when we update Pylint. It seems that newer versions of the tool report errors in some areas that are currently overlooked

##

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -338,7 +338,6 @@ class PreparedFunctionCall(Call):
         result = await self.call_raw(block_hash=block_hash, block_number=block_number)
         return self._payload_transformer.deserialize(result)
 
-    # TODO (#1182): add `cairo_version` parameter
     async def invoke(
         self,
         max_fee: Optional[int] = None,
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -35,7 +35,7 @@ from starknet_py.net.account.base_account import BaseAccount
 from starknet_py.net.client import Client
 from starknet_py.net.client_models import Call, EstimatedFee, Hash, ResourceBounds, Tag
 from starknet_py.net.models import AddressRepresentation, parse_address
-from starknet_py.net.models.transaction import Invoke
+from starknet_py.net.models.transaction import Declare, Invoke
 from starknet_py.net.udc_deployer.deployer import Deployer
 from starknet_py.proxy.contract_abi_resolver import (
     ContractAbiResolver,
@@ -162,9 +162,6 @@ class InvokeResult(SentTransaction):
         assert self.invoke_transaction is not None
 
 
-InvocationResult = InvokeResult
-
-
 @add_sync_methods
 @dataclass(frozen=True)
 class DeclareResult(SentTransaction):
@@ -181,6 +178,9 @@ class DeclareResult(SentTransaction):
     compiled_contract: str = None  # pyright: ignore
     """Compiled contract that was declared."""
 
+    declare_transaction: Declare = None  # pyright: ignore
+    """A Declare transaction that has been sent."""
+
     def __post_init__(self):
         if self._account is None:
             raise ValueError("Argument _account can't be None.")
@@ -191,6 +191,9 @@ class DeclareResult(SentTransaction):
         if self.compiled_contract is None:
             raise ValueError("Argument compiled_contract can't be None.")
 
+        if self.declare_transaction is None:
+            raise ValueError("Argument declare_transaction can't be None.")
+
     async def deploy(
         self,
         *,
@@ -199,7 +202,9 @@ class DeclareResult(SentTransaction):
         unique: bool = True,
         constructor_args: Optional[Union[List, Dict]] = None,
         nonce: Optional[int] = None,
+        tx_version: Optional[int] = None,
         max_fee: Optional[int] = None,
+        l1_resource_bounds: Optional[ResourceBounds] = None,
         auto_estimate: bool = False,
     ) -> "DeployResult":
         """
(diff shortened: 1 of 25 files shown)


starknet_py/contract.py
```diff
@@ -153,17 +154,14 @@ class InvokeResult(SentTransaction):
     contract: ContractData = None  # pyright: ignore
     """Additional information about the Contract that made the transaction."""
 
-    invoke_transaction: InvokeV1 = None  # pyright: ignore
+    invoke_transaction: Invoke = None  # pyright: ignore
     """A InvokeTransaction instance used."""
 
     def __post_init__(self):
         assert self.contract is not None
         assert self.invoke_transaction is not None
```
ddoktorski, line 164: I'd like to to initiate a discussion on whether it is better to add new parameters `tx_version` and `l1_resource_bounds` or to rename function names to include a version postfix:
1. The primary modifications should focus on `ContractFunction` and `PreparedFunctionCall` that have the following methods: `ContractFunction.invoke()`, `ContractFunction.prepare()`, `ContractFunction.call()`, `PreparedFunctionCall.invoke()`, `PreparedFunctionCall.estimate_fee()`, `PreparedFunctionCall.call()` where  `ContractFunction.invoke()` is  `ContractFunction.prepare() + PreparedFunctionCall.invoke()`. The `call()` functions do not require any changes regarding V3 transactions.
2. In terms of naming and logical coherence, it seems reasonable to me for `PreparedFunctionCall` to store `tx_version` and retain its current interface: `invoke`,  `estimate_fee` and `call` (that ignores `tx_version`). If we decide to add postfixes to both classes we may end up with situations like `prepare_v1().invoke_v3()`, which doesn't make sense.
3. We might consider appending postfixes to methods in `ContractFunction`. For now I've decided to use `tx_version` consistently, but this would be relatively easy to change.
4. Currently `tx_version` has been added also for `Contract.declare()` and `Contract.deploy_contract()`. For these functions I left default values for transaction version (depending on the cairo version in `declare` - V1 for Cairo 0  contracts and V2 for Cairo 1 and higher, V1 for `deploy_contract`), hence with the current implementation it is not a breaking change.

Disadvantages of the current approach:
- User must understand that `max_fee` is used for V1 and V2 transactions, while `l1_resource_bounds` is used for V3
- It introduces inconsistency with the `Account` interface as it includes methods like `sign_declare_v1` or `sign_invoke_v3`, etc.
- `Contract` methods have been designed to be compatible with `Account` and `PreparedFunctionCall` inherits `Call`. Users may assume that preparing a call with version 3 by using `ContractFunction.prepare(tx_version=3)` and then passing this prepared call to `Account.execute()` will execute `InvokeV3` which is not the case
- Currently many transaction V3 fields are not in use, so there is no need to add them to functions. They will be used in the future, which would require adding them to function parameters. This could make our already long list of parameters even longer and might make things more confusing for users  

What are your thoughts on this?

DelevoXDG: My thoughts on this:
> The call() functions do not require any changes regarding V3 transactions.

I think we should drop `call()` from `PreparedFunctionCall` (maybe rename it to `PreparedInvokeCall`).  It was confusing before (`prepare(...).estimate_fee()`). Now, it makes even less sense, as user might start passing some fee and version-related params to `prepare()`, unaware they're never going to be used. 

> prepare_v1().invoke_v3()

I think there are number of ways to avoid that:
1. `prepare()` does not take `tx_version`, `max_fee`, `l1_resource_bounds`. `PreparedFunctionCall` would then have `invoke_v1`, `invoke_v3`, `estimate_fee_vX`. Since we already allow passing `max_fee` in those, would that really make a difference to not allow passing those params to `prepare`? The current approach where these params can be set in multiple places seems quite confusing to me.
2. If we really want to allow setting these params in `prepare()`, `prepare()` could optional dataclasses like `ParamsV1`, `ParamsV3`. `PreparedFunctionCall` has `invoke_v1`, `invoke_v3`. The old approach the can be used then like so `max_fee=max_fee or self.params_v1.max_fee`. In that case, user at least knows that `invoke_vX` relies on whether they passed `params_vX` in `prepare` prior to that.
3. Best approach imo: make `PreparedFunctionCall` a base class. Derive `PreparedFunctionCallV1`, `PreparedFunctionCallV3` from it. Each would have `invoke` and `estimate_fee`. It would make most sense, since why would someone run `estimate_fee_v1` and `invoke_v3` on a prepared call.



> User must understand that max_fee is used for V1 and V2 transactions, while l1_resource_bounds is used for V3

This is the biggest issue from my perspective. I also think the numbers of conflicting fields in `Contract.declare()` is getting out of hand, so this is probably a good moment to either go for a `_vX` postfix approach or params approach. 

> Currently many transaction V3 fields are not in use, so there is no need to add them to functions. They will be used in the future, which would require adding them to function parameters. This could make our already long list of parameters even longer and might make things more confusing for users  

That's why we should consider adding Params dataclasses. That could actually allow us to avoid `_vX` postfixes in functions. The however will probably not fully work with `declare`, unless we create params not only for every version, but for every combination of version and tx type.

> It introduces inconsistency with the Account interface as it includes methods `like sign_declare_v1` or sign_invoke_v3, etc.

I think this is not a problem if the are alternative clues in the naming e.g. `invoke()` on a call prepared via `prepare_v3()` or mentioned params that have a versions in their name.

THenry14: This is basically the same discussion we've opened some time ago regarding jvm interfaces, so I guess we should extend it for all our sdks.

All your points are valid, and I'm not sure there is a silver bullet here. I do think the interfaces should ideally be consistent across all classes (Account, Contract etc). In jvm, if I remember correctly, we decided to just use postfixes, as this was less complicated and more explicit, with caveat that this is _ **_not_** _ the final solution (as we needed to act fast with and release asap). 

Regarding your changes, advantages of using tx_version outweight disadvantages at the moment, but I'm worried about bullet point 3. I wonder if we could adjust Account class a bit, make it an ABC/super that child classed (eg Account_v1, Account_v3) could just inherit from, allowing us to have consistent interfaces. thoughts?

I think in the long run we could also examine the spec very closely to learn if and how we can decide which version of tx we are dealing with. In my opinion, that would be preferable as tx version would be determined solely based on spec, but I'm not sure it is possible. Either way, we will have to decide on a consistent approach (be it postfixes/tx_version argument or something different).

DelevoXDG (commented):
Partial review to address the points discussed

THenry14 (commented):
the changes generally look good for currently proposed approach, but the more detailed review will come when we decide if we go with this or some other solution

ddoktorski:
Will be done in #1262

PR #1259, opened 2024-01-24
issue 1249: split deprecated schemas to their respective classes based on versions

### LINKED ISSUE #1249: Split deprecated schemas to their respective classes based on versions
Discussed [here](https://github.com/software-mansion/starknet.py/pull/1243#discussion_r1442840446)

Jonatan-Chaverri:
Closes #1249 

## Introduced changes

- Split deprecated rpc schemas to their respective classes based on version.

##

### DIFF
--- a/starknet_py/net/client_models.py
+++ b/starknet_py/net/client_models.py
@@ -191,16 +191,25 @@ class TransactionV3(Transaction):
 
 
 @dataclass
-class InvokeTransaction(DeprecatedTransaction):
+class InvokeTransactionV0(DeprecatedTransaction):
     """
-    Dataclass representing invoke transaction.
+    Dataclass representing invoke transaction v0.
     """
 
-    sender_address: int
     calldata: List[int]
-    # This field is always None for transactions with version = 1
-    entry_point_selector: Optional[int] = None
-    nonce: Optional[int] = None
+    contract_address: int
+    entry_point_selector: int
+
+
+@dataclass
+class InvokeTransactionV1(DeprecatedTransaction):
+    """
+    Dataclass representing invoke transaction v1.
+    """
+
+    calldata: List[int]
+    sender_address: int
+    nonce: int
 
 
 @dataclass
@@ -209,22 +218,43 @@ class InvokeTransactionV3(TransactionV3):
     Dataclass representing invoke transaction v3.
     """
 
-    sender_address: int
     calldata: List[int]
+    sender_address: int
+    nonce: int
     account_deployment_data: List[int]
+
+
+@dataclass
+class DeclareTransactionV0(DeprecatedTransaction):
+    """
+    Dataclass representing declare transaction v0.
+    """
+
+    sender_address: int
+    class_hash: int
+
+
+@dataclass
+class DeclareTransactionV1(DeprecatedTransaction):
+    """
+    Dataclass representing declare transaction v1.
+    """
+
+    sender_address: int
+    class_hash: int
     nonce: int
 
 
 @dataclass
-class DeclareTransaction(DeprecatedTransaction):
+class DeclareTransactionV2(DeprecatedTransaction):
     """
-    Dataclass representing declare transaction.
+    Dataclass representing declare transaction v2.
     """
 
-    class_hash: int  # Responses to getBlock and getTransaction include the class hash
     sender_address: int
-    compiled_class_hash: Optional[int] = None  # only in DeclareV2, hence Optional
-    nonce: Optional[int] = None
+    class_hash: int
+    compiled_class_hash: int
+    nonce: int
 
 
 @dataclass
@@ -233,10 +263,10 @@ class DeclareTransactionV3(TransactionV3):
     Dataclass representing declare transaction v3.
     """
 
+    sender_address: int
     class_hash: int
     compiled_class_hash: int
     nonce: int
-    sender_address: int
     account_deployment_data: List[int]
 
 
@@ -252,15 +282,15 @@ class DeployTransaction(Transaction):
 
 
 @dataclass
-class DeployAccountTransaction(DeprecatedTransaction):
+class DeployAccountTransactionV1(DeprecatedTransaction):
     """
-    Dataclass representing deploy account transaction.
+    Dataclass representing deploy account transaction v1.
     """
 
-    contract_address_salt: int
-    class_hash: int
-    constructor_calldata: List[int]
     nonce: int
+    contract_address_salt: int
+    constructor_calldata: List[int]
+    class_hash: int
 
 
 @dataclass
@@ -269,10 +299,10 @@ class DeployAccountTransactionV3(TransactionV3):
     Dataclass representing deploy account transaction v3.
     """
 
-    contract_address_salt: int
-    class_hash: int
-    constructor_calldata: List[int]
     nonce: int
+    contract_address_salt: int
+    constructor_calldata: List[int]
+    class_hash: int
 
 
 @dataclass
--- a/starknet_py/net/schemas/rpc.py
+++ b/starknet_py/net/schemas/rpc.py
@@ -10,13 +10,15 @@ from starknet_py.net.client_models import (
     ContractsNonce,
     DAMode,
     DeclaredContractHash,
-    DeclareTransaction,
     DeclareTransactionResponse,
     DeclareTransactionTrace,
+    DeclareTransactionV0,
+    DeclareTransactionV1,
+    DeclareTransactionV2,
     DeclareTransactionV3,
-    DeployAccountTransaction,
     DeployAccountTransactionResponse,
     DeployAccountTransactionTrace,
+    DeployAccountTransactionV1,
     DeployAccountTransactionV3,
     DeployedContract,
     DeployTransaction,
@@ -28,8 +30,9 @@ from starknet_py.net.client_models import (
     ExecutionResources,
     FeePayment,
     FunctionInvocation,
-    InvokeTransaction,
     InvokeTransactionTrace,
+    InvokeTransactionV0,
+    InvokeTransactionV1,
     InvokeTransactionV3,
     L1HandlerTransaction,
     L1HandlerTransactionTrace,
@@ -74,10 +77,7 @@ from starknet_py.net.schemas.common import (
     Uint64,
     Uint128,
 )
-from starknet_py.net.schemas.utils import (
-    _extract_tx_version,
-    _replace_invoke_contract_address_with_sender_address,
-)
+from starknet_py.net.schemas.utils import _extract_tx_version
 
 # pylint: disable=unused-argument, no-self-use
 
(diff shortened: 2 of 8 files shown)


ddoktorski (changes requested):
Thanks! It's definitely moving in the right direction. However, there are a few additional changes we'll need:
- The schemas should accurately reflect the [RPC specification](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_api_openrpc.json). So for instance `DeclareTransactionV0Schema` shouldn't contain `compiled_class_hash` field and in `DeclareTransactionV1Schema` field `nonce` should be required
- Ideally we would like to also break down classes in [client_models.py](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/net/client_models.py) in a similar manner. For example replacing `DeclareTransaction` with versioned classes like `DeclareTransactionV1` etc.
- Similarly, these changes should also be implemented for `DeprecatedDeployAccountTransactionSchema`

For more context, please refer to the following resources:
- [Starknet RPC specification](https://github.com/starkware-libs/starknet-specs/tree/v0.6.0) (version 0.6.0)
- [Starknet transaction types](https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/)

Please let me know if you have any questions or need further clarification on any point.

Jonatan-Chaverri:
@ddoktorski thanks for the clarification. I implemented the requested changes. Please let me know what you think. Thanks.

starknet_py/net/client_models.py
```diff
@@ -149,24 +149,26 @@ class Transaction(ABC):
...
 
 
 @dataclass
-class DeprecatedTransaction(Transaction):
+class DeprecatedTransaction(ABC):
     """
     Dataclass representing common attributes of transactions v1 and v2.
     """
 
+    hash: Optional[int]
```
Jonatan-Chaverri, line 169: I was not sure if I could modify this Deprecated classes without breaking anything... they are being used in other parts of the code... but the required changes in schemas did not include this things like "hash"... so what I did was to remove the usage of this deprecated classes (in my schemas) and create new classes. But as I say above, since I don't know the usage of this, I just moved the old fields in here and make it a base class (not inheriting from the modified transaction class), so in the end, the schema for this will be the same as before. Not sure if that's ok.

starknet_py/net/schemas/rpc.py
```diff
@@ -230,17 +231,17 @@ def make_dataclass(self, data, **kwargs) -> ResourceBoundsMapping:
...
-
-class DeprecatedTransactionSchema(TransactionSchema):
-    max_fee = Felt(data_key="max_fee", required=True)
+    type = fields.Str(required=True, validate=validate.OneOf(["DECLARE"]))
+    sender_address = Felt(data_key="sender_address", required=True)
+    signature = fields.List(Felt(), data_key="signature", required=True)
+    class_hash = Felt(data_key="class_hash", required=True)
 
 
-class TransactionV3Schema(TransactionSchema):
```
ddoktorski, line 242: I would definitely prefer to keep `TransactionV3Schema` to group common V3 transaction fields together. How about we keep things simple for now and think of enhancements later? So, here's the inheritance structure:
- `TransactionSchema` -> `DeprecatedTransactionSchema` ->:
  - `InvokeTransactionV0Schema`, `InvokeTransactionV1Schema`
  - `DeclareTransactionV0Schema`, `DeclareTransactionV1Schema` and `DeclareTransactionV2Schema`
  - `DeployAccountTransactionV1Schema`
- `TransactionSchema` -> `TransactionV3Schema` -> all V3 transaction schemas (no changes)

Please pay attention to the `required` property for all fields across different schemas.

starknet_py/net/client_models.py
ddoktorski: These changes will cause some breaks for sure, mostly in the tests. I see the following options:
- We can limit the scope of this PR only to changes in the schemas without modifications in the `client_models`
- You can try to take on everything, including edits in the `client_models` that'll cause breaks. If adapting the code to the new changes gets too complex, I can take over from there

Please let me know what you'd prefer.

starknet_py/net/client_models.py
```diff
@@ -191,28 +193,53 @@ def __post_init__(self):
 
 
 @dataclass
-class InvokeTransaction(DeprecatedTransaction):
```
ddoktorski, line 194: Just like in the schemas, let's keep `DeprecatedTransaction`, `TransactionV3` and have only version specifc classes, such as `InvokeTransactionV0`, `InvokeTransactionV1` etc. We'd like to avoid having classes with generic names like `InvokeTransaction` or `DeclareTransaction`. These names imply they are default classes for given transactions, which isn't true. We may consider introducing later something like `InvokeTransactionCommonFields`, `DeclareTransactionCommonFields` etc.

Jonatan-Chaverri:
@ddoktorski I applied your feedback. Please review again. 

Now I'm not modifying any of the base classes (like transaction or deprecatedTransaction) and also using the proposed inheritance structure. 
I also did not modify any logic related to dataclasses, old dataclasses are still there and are still being used in the code cause I don't really know how to change that part. I can also take care of updating that logic if you can explain to me how to change it, but if you think its too complicated to explain and you prefer to do it yourself, thats fine too. 

Please let me know what you think. Thanks

starknet_py/net/client_models.py
```diff
@@ -203,16 +203,41 @@ class InvokeTransaction(DeprecatedTransaction):
     nonce: Optional[int] = None
 
 
+@dataclass
+class InvokeTransactionV0(DeprecatedTransaction):
+    """
+    Dataclass representing invoke transaction v0.
+    """
+
+    type: str
```
ddoktorski, line 212: Could you remove the `type` from all classes? It's not necessary to include a field for it, as it can be deduced from the class type. Apologies for not pointing this out earlier.

starknet_py/net/schemas/rpc.py
```diff
@@ -253,50 +256,85 @@ class TransactionV3Schema(TransactionSchema):
     )
 
 
-class DeprecatedInvokeTransactionSchema(DeprecatedTransactionSchema):
-    contract_address = Felt(data_key="contract_address", load_default=None)
-    sender_address = Felt(data_key="sender_address", load_default=None)
-    entry_point_selector = Felt(data_key="entry_point_selector", load_default=None)
+class InvokeTransactionV0Schema(DeprecatedTransactionSchema):
+    type = fields.Str(required=True, validate=validate.OneOf(["INVOKE"]))
```
ddoktorski, line 260: The same comment regarding `type` as [here](https://github.com/software-mansion/starknet.py/pull/1259/files#r1467612983)

starknet_py/net/client_models.py
```diff
@@ -203,16 +203,41 @@ class InvokeTransaction(DeprecatedTransaction):
     nonce: Optional[int] = None
```
ddoktorski, line 203: Please also remove the old classes `InvokeTransaction`, `DeclareTransaction` and `DeployAccountTransaction`.

ddoktorski (changes requested):
Thanks! For the code modifications to align with these changes, it would be necessary to perform a thorough search across the entire codebase for instances of `InvokeTransaction`, `DeclareTransaction`, and `DeployAccountTransaction`. Then we could substitute each instance with the appropriate versioned class. In certain cases, the correct class replacement should be clear, though in others it might be tricky without sufficient context.

If you feel comfortable and willing to tackle this task, please go ahead :)  To test your changes locally, just run `poe test_ci_v2` and `poe test_ci_docs_v2`.

Jonatan-Chaverri:
@ddoktorski removed the old classes in code logic. Please review again.

starknet_py/net/client_models.py
```diff
@@ -252,15 +282,15 @@ class DeployTransaction(Transaction):
...
     """
 
-    contract_address_salt: int
-    class_hash: int
-    constructor_calldata: List[int]
-    nonce: int
+    nonce: str
+    contract_address_salt: str
+    constructor_calldata: List[str]
+    class_hash: str
```
ddoktorski, line 293: Please change these types to `int`, the same for `DeployAccountTransactionV3`.

starknet_py/net/schemas/rpc.py
```diff
@@ -253,50 +256,78 @@ class TransactionV3Schema(TransactionSchema):
...
+class InvokeTransactionV0Schema(DeprecatedTransactionSchema):
     calldata = fields.List(Felt(), data_key="calldata", required=True)
-    nonce = Felt(data_key="nonce", load_default=None)
+    contract_address = Felt(data_key="contract_address", required=True)
+    entry_point_selector = Felt(data_key="entry_point_selector", required=True)
 
     @post_load
-    def make_transaction(self, data, **kwargs) -> InvokeTransaction:
+    def make_transaction(self, data, **kwargs) -> InvokeTransactionV0:
         _replace_invoke_contract_address_with_sender_address(data)
```
ddoktorski, line 266: This function call can now be removed, and the function itself can also be removed.

Jonatan-Chaverri:
@ddoktorski done, please check again. Thanks

starknet_py/tests/e2e/tests_on_networks/trace_api_test.py
```diff
@@ -22,9 +22,9 @@
 @pytest.mark.asyncio
 async def test_trace_transaction(client_goerli_testnet):
     tx_to_trace: dict[type[Transaction], type[TransactionTrace]] = {
-        InvokeTransaction: InvokeTransactionTrace,
-        DeclareTransaction: DeclareTransactionTrace,
-        DeployAccountTransaction: DeployAccountTransactionTrace,
+        InvokeTransactionV1: InvokeTransactionTrace,
```
ddoktorski, line 25: It is also necessary to add here:
```
InvokeTransactionV0: InvokeTransactionTrace
```
There are both V0 and V1 invoke transactions in this block.

ddoktorski (commented):
Appreciate your efforts! I've added one more comment, but overall it looks good to me 👍

Requesting other team members to also take a look at this :)

starknet_py/tests/e2e/tests_on_networks/trace_api_test.py
Jonatan-Chaverri: Done

Jonatan-Chaverri:
@ddoktorski no problem! I made the changes, please check again.

ddoktorski (approved):
Note: I've checked locally that tests on network pass for these changes

THenry14 (approved):
lgtm, thank you for a contribution! 💪

ddoktorski:
@Jonatan-Chaverri 
Could you address issues that have occurred after `development` merge? Please add this a separate commit. Once this is resolved, we can proceed with merging this PR :muscle:

Jonatan-Chaverri:
@ddoktorski done, please check again. Thanks.

PR #1262, opened 2024-01-24
Support V3 transactions in the `Contract` class

ddoktorski:
Related to #1235 
Initial PR #1255

## Introduced changes

- `PreparedFunctionCall` class:
  - `PreparedFunctionCall` has only methods `call()` and `call_raw()`
  - Add abstract class `PreparedFunctionInvoke` and derive from it `PreparedFunctionInvokeV1` and `PreparedFunctionInvokeV3` with methods `invoke()` and `estimate_fee()`
- `Contract` class:
  - Break down `declare()` into:
    - `declare_v1()`, `declare_v2()` and `declare_v3()`
  - Break down `deploy_contract()` into:
    - `deploy_contract_v1()`, `deploy_contract_v3()` and add `salt` and `unique` parameters
- `ContractFunction` class:
  - Break down `prepare()` into:
    - `prepare_invoke_v1()`, `prepare_invoke_v3()` and `prepare_call()`
  - Break down `invoke()` into:
    - `invoke_v1()` and `invoke_v3()`
- `DeclareResult` class:
  - Break down `deploy()` into:
    - `deploy_v1()` and `deploy_v3()`
- Refactoring:
  - ~Move `SentTransaction` class to `sent_transaction.py`~
  - ~Move `ContractFunction`, `ContractData`, `InvokeResult` and prepared call/invoke related classes to `contract_function.py`~
  - Create `contract_utils.py` file
  - ~Move all `Contract` related classes to the `starknet_py/net/contract` directory~

##

Please note that documentation update and support for `Account.deploy_account()` will be addressed in separate PRs.


ddoktorski:
Regarding the refactor, splitting the code into files may not look perfect. It's tough to optimally split these closely tied classes without stumbling into circular import issues. If this refactoring doesn't make sense or complicates the review, let me know so I will revert these changes.

starknet_py/net/contract/contract_function.py
```diff
@@ -0,0 +1,481 @@
...
+        :param auto_estimate: Use automatic fee estimation, not recommend as it may lead to high costs.
+        :param nonce: Nonce of the transaction.
+        :return: InvokeResult.
+        """
+        prepared_invoke = self.prepare_invoke_v1(*args, **kwargs)
+        return await prepared_invoke.invoke(
+            max_fee=max_fee, nonce=nonce, auto_estimate=auto_estimate
+        )
+
+    def prepare_invoke_v3(
```
THenry14, line 422: I'm aware of the fact that we agreed to use postfixes. I do wonder, however, **if in this particular case**, where (prepare)_invoke(v1 | v3) can be differed by max_fee/resource_bounds and this is a helper class, would it make sense to have one prepare_invoke and one invoke, that takes those values as Union and based on type it prepares/makes v1 or v3 tx? wdyt? I know it may look like breaking existing approach to interface, but I do wonder if it does make sense in a helper class

ddoktorski: I definitely don't like the current interface. However, setting a fee as `Union` may not be the best approach:
- A unified name for `max_fee` and `l1_resource_bounds` would be required, let's say just `fee`. While this could make sense for a user with plenty of context, it might be confusing in general
- Thinking long term we will need to incorporate other V3 fields such as `l2_resource_bounds`, `tip` etc. As a result, we would need to alter this interface again

One potential solution is to create `InvokeV1Params` and `InvokeV3Params` and then use 
```
prepare_invoke(params: InvokeParams)
```
However, this approach will break consistency with other parts of our codebase.

THenry14: Sure, it was just a rough idea either way. We can leave this as is.

THenry14 (approved):
lgtm

starknet_py/net/contract/contract_function.py
```diff
@@ -0,0 +1,481 @@
...
+    """
+    Basic data of a deployed contract.
+    """
+
+    address: int
+    abi: ABI
+    cairo_version: int
+
+    @cached_property
+    def parsed_abi(self) -> Union[Abi, AbiV1, AbiV2]:
```
DelevoXDG, line 44: Maybe
```suggestion
    def parsed_abi(self) -> Union[AbiV0, AbiV1, AbiV2]:
```
?

starknet_py/net/contract/contract_function.py
```diff
@@ -0,0 +1,481 @@
...
+    invoke_transaction: Invoke = None  # pyright: ignore
+    """A InvokeTransaction instance used."""
+
+    def __post_init__(self):
+        assert self.contract is not None
+        assert self.invoke_transaction is not None
+
+
+@dataclass
+class CallToSend(Call):
```
DelevoXDG, line 93: Have you consider any other names for this one? 😅 

I'm thinking `PreparedCall` or `PreparedCallBase`.

starknet_py/net/contract/contract_function.py
```diff
@@ -0,0 +1,481 @@
...
+        :param block_number: Optional block number.
+        :return: TupleDataclass representing call result.
+        """
+        result = await self.call_raw(block_hash=block_hash, block_number=block_number)
+        return self._payload_transformer.deserialize(result)
+
+
+@add_sync_methods
+@dataclass
+class PreparedFunctionInvoke(ABC, CallToSend):
```
DelevoXDG, line 135: A bit unsure about class name ending with a verb

starknet_py/net/contract/contract_function.py
```diff
@@ -0,0 +1,481 @@
...
+        return await self._invoke(transaction)
+
+    async def estimate_fee(
+        self,
+        block_hash: Optional[Union[Hash, Tag]] = None,
+        block_number: Optional[Union[int, Tag]] = None,
+        *,
+        nonce: Optional[int] = None,
+    ) -> EstimatedFee:
+        tx = await self.get_account.sign_invoke_v1_transaction(
```
DelevoXDG, line 224: I looked into `sign_invoke_v1_transaction`. Why aren't we using query versions for fee estimation?

I understand that it was this way done before, but since RPC 0.6.0 reintroduced query versions, maybe it would make sense to adopt a more correct approach, either now or in a separate PR in the nearby future.

starknet_py/net/contract/sent_transaction.py
```diff
@@ -0,0 +1,57 @@
...
+    ) -> TypeSentTransaction:
+        """
+        Waits for transaction to be accepted on chain till ``ACCEPTED`` status.
+        Returns a new SentTransaction instance, **does not mutate original instance**.
+        """
+        if wait_for_accept is not None:
+            warnings.warn(
+                "Parameter `wait_for_accept` has been deprecated - since Starknet 0.12.0, transactions in a PENDING"
+                " block have status ACCEPTED_ON_L2."
+            )
```
DelevoXDG, line 43: Current starknet.py version is not compatible with 0.12.0 anymore, so we can probably remove this, right?

starknet_py/tests/e2e/client/full_node_test.py
DelevoXDG, line 1: Are you planning to add `invoke_v3` tests in this file as well?

ddoktorski: Thank you for pointing this out. Will add some tests for `invoke_v3` here.

starknet_py/net/contract/contract.py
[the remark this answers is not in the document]
ddoktorski: This is to improve file organization. I would prefer to move it to `starknet_py/contract`, but considering that we already have `starknet_py/net/account` in place, I think it makes sense to keep contract directory in `starknet_py/net` as well.

starknet_py/net/contract/sent_transaction.py
ddoktorski: Will be done in https://github.com/software-mansion/starknet.py/issues/1268, as there is the same parameter to remove in the `Client.wait_for_tx()`.

starknet_py/net/contract/contract.py
```diff
@@ -180,6 +53,9 @@ class DeclareResult(SentTransaction):
     compiled_contract: str = None  # pyright: ignore
     """Compiled contract that was declared."""
 
+    declare_transaction: Declare = None  # pyright: ignore
+    """A Declare transaction that has been sent."""
+
```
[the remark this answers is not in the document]
ddoktorski: It is just a helper really. It doesn't require any additional effort from users, and I think it is simply convenient to have. I also think it makes sense that `SentTransaction` contains the actual transaction that has been sent. Regarding the duplicated fields, I agree that there is some space for improvement, but I wouldn't consider this a big concern.

starknet_py/net/contract/contract_function.py
ddoktorski: Will be changed to `PrepareCallBase` 👍

starknet_py/net/contract/contract_function.py
ddoktorski: Agree that this this is not an ideal name. I would rather prefer to avoid `FunctionInvocation`, since it has a [different meaning](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_trace_api_openrpc.json#L300) in the RPC specification.

starknet_py/tests/e2e/client/full_node_test.py
ddoktorski: On second thoughts, these tests are actually meant for `FullNodeClient.get_events()` and `invoke_v1` is used here only as a helper function. We could use `invoke_v3` as well, but let's leave as it is.

starknet_py/net/contract/contract_function.py
ddoktorski: Good catch! Added `sign_for_fee_estimate()` call to `estimate_fee` methods in `PreparedFunctionInvokeV1` and `PreparedFunctionInvokeV3`.

ddoktorski:
After all, I decided to rollback the refactoring changes. This decision was driven by the user perspective, where it only created needless confusion. For instance, `DeclareResult` was found in the contract module, while `InvokeResult` resided in the contract_function module. I would prefer not to have everything in a single file, but I couldn't find a satisfactory way to logically divide these classes. Other than that, the only changes were made to address @DelevoXDG comments and fix docs build.

starknet_py/contract.py
```diff
@@ -3,14 +3,15 @@
...
 from functools import cached_property
-from typing import Dict, List, Optional, Tuple, TypeVar, Union
+from typing import Dict, List, Optional, TypeVar, Union
 
 from marshmallow import ValidationError
 
-from starknet_py.abi.model import Abi
-from starknet_py.abi.parser import AbiParser
+from starknet_py.abi import Abi as AbiV0
+from starknet_py.abi import AbiParser as AbiV0Parser
```
DelevoXDG, line 14: Nit: Maybe move it to `starknet_py.abi.v0.model` for consistency with other abis?
Might be done in a separate PR if that would clutter diff.

starknet_py/net/contract/contract_function.py
DelevoXDG: Make sense, we have `Invoke` class anyway, so that'll do

starknet_py/tests/e2e/client/full_node_test.py
DelevoXDG: Sorry, didn't look into exact purpose of the tests. If all the tests indeed use `invoke_v1` as a helper, I agree it's not necessary

starknet_py/contract.py
```diff
@@ -338,7 +356,70 @@ async def call(
...
+            )
+
+    @property
+    def get_account(self):
+        if self._account is not None:
+            return self._account
+
+        raise ValueError(
+            "The account is not defined. It is not possible to send an invoke transaction."
+        )
```
DelevoXDG, line 380: 1. Do we really want to expose account there? It like a bit confusing property to have in `PreparedFunctionInvokeVX`, since it's only used as an internal helper and user would probably be aware which account they used to create prepared invoke anyway.
2. Also, could we maybe make `_account` non-optional? We can check whether it's `None` in `prepare_invoke_vX` before creating instance `PreparedFunctionInvokeVX`.

starknet_py/contract.py
```diff
@@ -382,26 +452,75 @@ async def estimate_fee(
...
+    async def invoke(
+        self,
+        l1_resource_bounds: Optional[ResourceBounds] = None,
+        auto_estimate: bool = False,
+        *,
+        nonce: Optional[int] = None,
+    ) -> InvokeResult:
         """
-        Estimate fee for prepared function call.
+        Send an Invoke transaction version 3 for a prepared data.
```
DelevoXDG, line 487: Nit:
```suggestion
        Send an Invoke transaction version 3 for the prepared data.
```

starknet_py/contract.py
```diff
@@ -382,26 +452,75 @@ async def estimate_fee(
...
 
-        :param block_hash: Estimate fee at specific block hash.
-        :param block_number: Estimate fee at given block number
-            (or "latest" / "pending" for the latest / pending block), default is "pending".
+        :param l1_resource_bounds: Max amount and max price per unit of L1 gas (in Fri) used when executing
+            this transaction.
+        :param auto_estimate: Use automatic fee estimation, not recommend as it may lead to high costs.
```
DelevoXDG, line 491:
```suggestion
        :param auto_estimate: Use automatic fee estimation; not recommended as it may lead to high costs.
```
or 
```suggestion
        :param auto_estimate: Use automatic fee estimation (not recommended, as it may lead to high costs).
```
(both can be found in the codebase)

starknet_py/contract.py
```diff
@@ -701,35 +942,112 @@ async def deploy_contract(
...
         :param account: BaseAccount used to sign and send deploy transaction.
         :param class_hash: The class_hash of the contract to be deployed.
         :param abi: An abi of the contract to be deployed.
         :param constructor_args: a ``list`` or ``dict`` of arguments for the constructor.
         :param deployer_address: Address of the UDC. Is set to the address of
-            the default UDC (same address on mainnet/testnet/devnet) by default.
+            the default UDC (same address on mainnet/goerli/sepolia) by default.
             Must be set when using custom network other than ones listed above.
         :param cairo_version: Version of the Cairo in which contract is written.
+            By default, it is set to 0.
```
DelevoXDG, line 959: Nit: I would consider removing default there, as we're slowly moving away from cairo 0.

Alternatively, and that's just a rough idea for the future, we could check the version of declare transaction used to declare contract at given class hash and determine cairo version that way.

starknet_py/tests/e2e/docs/code_examples/test_contract_function.py
```diff
@@ -2,12 +2,23 @@
...
 
 from starknet_py.contract import Contract, ContractFunction
+from starknet_py.net.client_models import ResourceBounds
 
 
-def test_prepare(map_contract: Contract):
-    # docs-start: prepare
-    prepared_function_call = map_contract.functions["put"].prepare(key=10, value=20)
-    # docs-end: prepare
+def test_prepare_invoke_v1(map_contract: Contract):
```
DelevoXDG, line 8: Should we also add `test_prepare_call`?

starknet_py/contract.py
ddoktorski: 1. The main motivation here is to avoid pyright errors when using `_account`. We are sure that it isn't `None`, as we make this verification in `__post_init__` (excluding cases where a user might change it manually on an existing `PreparedFunctionInvokeVX` instance). Considering that an account is a prerequisite for transaction invocation, it kind of makes sense to have such a property in `PreparedFunctionInvoke`. As an alternative, I could make `_get_account` internal.
2. The `account` in the `Contract` class might be `None`, resulting in a None when calling prepare_invoke_vX. We can potentially make this assertion there, but then we would need to add this in four places instead of just one.

starknet_py/tests/e2e/docs/code_examples/test_contract_function.py
ddoktorski: I intentionally left out `test_prepare_call` because I think `prepare_call` isn't generally very useful. However I didn't want to completely remove it from the codebase. There are examples for `ContractFunction.call` and `ContractFunction.call_raw`, which should be sufficient.

starknet_py/contract.py
DelevoXDG: > As an alternative, I could make _get_account internal.

That is one solution, because it's overly clear what it does in the context of `PreparedFunctionInvokeVX`

> We can potentially make this assertion there, but then we would need to add this in four places instead of just one.

More like 4 instead of 2. Also, can we maybe check `_account` right away when it's passed to constructor?

starknet_py/contract.py
DelevoXDG: Are you planning to add an issue for this?

starknet_py/tests/e2e/docs/code_examples/test_contract_function.py
DelevoXDG: Makes sense

DelevoXDG (approved):
lgtm 🔥

starknet_py/contract.py
ddoktorski: Thanks! Will be done in #1270

starknet_py/contract.py
ddoktorski: https://github.com/software-mansion/starknet.py/issues/1271

starknet_py/contract.py
ddoktorski: > More like 4 instead of 2

It is implemented now only within the abstract `PreparedFunctionInvoke` class

> can we maybe check _account right away when it's passed to constructor?

This is a `dataclass`, so the only option is `__post_init__`

PR #1263, opened 2024-01-25
Fix serializer_for_event for events from contracts in version v2

### LINKED ISSUE #1238: [BUG] different attribute name in EventType in comparison with Abi.Event
### What happened

[_AbiV2Parser_](https://github.com/software-mansion/starknet.py/blob/1859b5d2d2be9963c2fa178567cc2ba82a282823/starknet_py/abi/v2/parser.py#L81C9-L81C37) produces abi which attribute 'events' has type Dict[str, **EventType**], meanwhile [_AbiParser_](https://github.com/software-mansion/starknet.py/blob/1859b5d2d2be9963c2fa178567cc2ba82a282823/starknet_py/abi/parser.py#L99C17-L99C47) produces one with type Dict[str, **Event**].
**EventType** has fields _name_ and _types_, but **Event** has fields _name_ and _data_. Serialization works with **Event** rn, not with  **EventType**. An exception is raised when you try to pass object of type  **EventType** to _serializer_for_event_ because it has no attribute 'data' ('types' instead).

### Stack trace

-

### Steps to reproduce

1. make a contract with cairo version = 2 and try to make a serialization for some of its events.

### SDK Version

0.18.3

### Python version

3.9

Linux

tkumor3:
Closes #1238 

## Introduced changes

- 
- 

##

### DIFF
--- a/starknet_py/serialization/factory.py
+++ b/starknet_py/serialization/factory.py
@@ -11,6 +11,7 @@ from starknet_py.cairo.data_types import (
     BoolType,
     CairoType,
     EnumType,
+    EventType,
     FeltType,
     NamedTupleType,
     OptionType,
@@ -148,13 +149,16 @@ def serializer_for_outputs(payload: List[CairoType]) -> OutputSerializer:
     )
 
 
-def serializer_for_event(event: Abi.Event) -> PayloadSerializer:
+def serializer_for_event(event: Abi.Event | EventType) -> PayloadSerializer:
     """
     Create serializer for an event.
 
     :param event: parsed event.
     :return: PayloadSerializer that can be used to (de)serialize events.
     """
+    if isinstance(event, EventType):
+        return serializer_for_payload(event.types)
+
     return serializer_for_payload(event.data)
 
 
--- a/starknet_py/serialization/factory.py
+++ b/starknet_py/serialization/factory.py
@@ -148,24 +148,24 @@ def serializer_for_outputs(payload: List[CairoType]) -> OutputSerializer:
         serializers=[serializer_for_type(cairo_type) for cairo_type in payload]
     )
 
-EventV1 = AbiV1.Event
+
 EventV0 = Abi.Event
+EventV1 = AbiV1.Event
 EventV2 = EventType
 
-def serializer_for_event(
-    event: EventV0 | EventV1 | EventV2
-) -> PayloadSerializer:
+
+def serializer_for_event(event: EventV0 | EventV1 | EventV2) -> PayloadSerializer:
     """
     Create serializer for an event.
 
     :param event: parsed event.
     :return: PayloadSerializer that can be used to (de)serialize events.
     """
-    if isinstance(event, EventType):
-        return serializer_for_payload(event.types)
-    if isinstance(event, AbiV1.Event):
+    if isinstance(event, EventV0):
+        return serializer_for_payload(event.data)
+    if isinstance(event, EventV1):
         return serializer_for_payload(event.inputs)
-    return serializer_for_payload(event.data)
+    return serializer_for_payload(event.types)
 
 
 def serializer_for_function(abi_function: Abi.Function) -> FunctionSerializationAdapter:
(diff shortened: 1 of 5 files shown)


starknet_py/serialization/factory.py
```diff
@@ -148,13 +149,16 @@ def serializer_for_outputs(payload: List[CairoType]) -> OutputSerializer:
     )
 
 
-def serializer_for_event(event: Abi.Event) -> PayloadSerializer:
+def serializer_for_event(event: Abi.Event | EventType) -> PayloadSerializer:
```
ddoktorski, line 152: What about `Event` type for V1 Abi? I assume we can have the same issue that we encountered with V2.

starknet_py/tests/e2e/fixtures/abi_v2_structures.py
```diff
@@ -0,0 +1,193 @@
...
+    "outputs": [{"type": "test::MyEnum::<core::integer::u128>"}],
+    "state_mutability": "view",
+}
+foo_view = Abi.Function(
+    name="foo_view",
+    inputs=OrderedDict(a=FeltType(), b=UintType(128)),
+    outputs=[my_enum],
+)
+
+abi_v2 = Abi(
```
ddoktorski, line 177: If I understand correctly, and we only require `events` for the current tests, then why do we need to define all the other fields?

starknet_py/serialization/factory_test.py
```diff
@@ -95,6 +100,12 @@ def test_getting_payload_serializer():
     )
 
 
+def test_getting_payload_serializer_v2():
```
ddoktorski, line 103: Could we also add explicit tests for `serializer_for_event` for V1 and V2? An approach similar to [this](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/serialization/serialization_test.py#L124), could be done for actual contracts.

starknet_py/serialization/factory.py
tkumor3: In the context of V1 abi, the event has `Abi.Event` type, the same as v0.  I think I have to add tests also for V1

starknet_py/tests/e2e/fixtures/abi_v2_structures.py
tkumor3: Probably we will need to add more tests for abi v2 so I created the fixture which will be useful in other tests

starknet_py/serialization/factory.py
ddoktorski: The names are a bit tricky here. As I understand, we have two `Abi.Event` types in different modules, corresponding to V0 and V1 respectively.

starknet_py/tests/e2e/fixtures/abi_v2_structures.py
ddoktorski: In that case, could you either add more Abi V2 tests in this PR or stick with just the `events` field?

starknet_py/serialization/factory.py
tkumor3: I was sure that both events were taken from the same place, but it wasn't true! So I added support for AbiV1.Event. Good catch!

starknet_py/serialization/serialization_test.py
```diff
@@ -133,3 +152,27 @@ def test_event_serialization():
...
         expected_serialized
     )
+
+
+def test_event_serialization_v1():
+    serializer = serializer_for_event(parsed_abi_v1.events["Approval"])
+
+    approval = {"owner": 1, "spender": 2, "value": 3}
+    serialized = serializer.serialize(approval)
+    assert serialized == [1, 2, 3, 0]
```
ddoktorski, line 162: What does `0` at the end mean here?

tkumor3: I think it's related to that, the value is `core::integer::u256`, so this 0 is added to save info about the type.  @THenry14 could you confirm?

THenry14: If I remember correctly this is partly true. I think this is because the value is u256, and this cannot fit into single felt - that is why this is being split into `high` and `low` value (and 3 is very small, so its part can fit into felt, leaving the other part as 0).

ddoktorski (approved):
LGTM

PR #1265, opened 2024-01-25
Add `Account.deploy_account_v3` static method

ddoktorski:
Related to #1235 and #1262 

## Introduced changes

- Add static method `Account.deploy_account_v3()`
- Rename:
  - `deploy_account` -> `deploy_account_v1` 
  - `execute` -> `execute_v1`
  - helper methods for parsing calls, previous names with v1 and v2 postfixes were confusing

Please note that documentation update will be done in a separate PR.

##

### DIFF
--- a/starknet_py/contract.py
+++ b/starknet_py/contract.py
@@ -241,7 +241,7 @@ class DeclareResult(SentTransaction):
             calldata=constructor_args,
             cairo_version=self._cairo_version,
         )
-        res = await self._account.execute(
+        res = await self._account.execute_v1(
             calls=deploy_call, nonce=nonce, max_fee=max_fee, auto_estimate=auto_estimate
         )
 
@@ -728,7 +728,7 @@ class Contract:
             calldata=constructor_args,
             cairo_version=cairo_version,
         )
-        res = await account.execute(
+        res = await account.execute_v1(
             calls=deploy_call, nonce=nonce, max_fee=max_fee, auto_estimate=auto_estimate
         )
 
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -627,31 +627,20 @@ class Account(BaseAccount):
         :param max_fee: max fee to be paid for deployment, must be less or equal to the amount of tokens prefunded.
         :param auto_estimate: Use automatic fee estimation, not recommend as it may lead to high costs.
         """
-        address = parse_address(address)
         calldata = (
             constructor_calldata
             if constructor_calldata is not None
             else [key_pair.public_key]
         )
 
-        if address != (
-            computed := compute_address(
-                salt=salt,
-                class_hash=class_hash,
-                constructor_calldata=calldata,
-                deployer_address=0,
-            )
-        ):
-            raise ValueError(
-                f"Provided address {hex(address)} is different than computed address {hex(computed)} "
-                f"for the given class_hash and salt."
-            )
-
-        account = Account(
+        account = _prepare_account_to_deploy(
             address=address,
-            client=client,
+            class_hash=class_hash,
+            salt=salt,
             key_pair=key_pair,
+            client=client,
             chain=chain,
+            calldata=calldata,
         )
 
         deploy_account_tx = await account.sign_deploy_account_v1_transaction(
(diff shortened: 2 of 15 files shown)


starknet_py/net/account/account.py
```diff
@@ -663,12 +651,7 @@ async def deploy_account(
...
 
-        if chain in (
-            StarknetChainId.SEPOLIA_TESTNET,
-            StarknetChainId.SEPOLIA_INTEGRATION,
-            StarknetChainId.GOERLI,
-            StarknetChainId.MAINNET,
-        ):
+        if chain in StarknetChainId:
             balance = await account.get_balance()
             if balance < deploy_account_tx.max_fee:
```
ddoktorski, line 656: I am not sure how to implement a similar check for the V3 transaction now. I'm guessing eventually we'd need to pull the balance from the STRK contract and an exchange rate to WEI?

Let me know if I am missing something here. Generally I think it is not crucial, but definitely nice to have.

THenry14: Good question. I suppose it could depend on fee DA mode as well, but I'm not 100% sure. Let's open an issue to investigate this, we might need to ask SW as well.

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -13,6 +13,7 @@
     Call,
     DeployAccountTransaction,
```
THenry14, line 14: Should we conform already existing names to new ones?

starknet_py/contract.py
```diff
@@ -241,7 +241,7 @@ async def deploy(
             calldata=constructor_args,
             cairo_version=self._cairo_version,
         )
-        res = await self._account.execute(
+        res = await self._account.execute_v1(
```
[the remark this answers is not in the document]
ddoktorski: It has been done in https://github.com/software-mansion/starknet.py/pull/1243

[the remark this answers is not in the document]
THenry14: ahh shit, I have too short memory. Thanks for linking this!

starknet_py/net/account/account.py
ddoktorski: https://github.com/software-mansion/starknet.py/issues/1266

starknet_py/tests/e2e/account/account_test.py
ddoktorski: Work in progress in #1259

THenry14 (approved):

starknet_py/net/account/account.py
```diff
@@ -605,53 +605,41 @@ async def deploy_account(
...
     ) -> AccountDeploymentResult:
-        # pylint: disable=too-many-locals
         """
         Deploys an account contract with provided class_hash on Starknet and returns
         an AccountDeploymentResult that allows waiting for transaction acceptance.
 
         Provided address must be first prefunded with enough tokens, otherwise the method will fail.
 
-        If using Client for either TESTNET or MAINNET, this method will verify if the address balance
+        If using Client for GOERLI, SEPOLIA or MAINNET, this method will verify if the address balance
```
DelevoXDG, line 614: Maybe we should be explicit about `SEPOLIA` and `SEPOLIA_INTEGRATION`

starknet_py/net/account/account.py
DelevoXDG: This is not crucial since tx will be rejected with a descriptive enough error message anyway

> I'm guessing eventually we'd need to pull the balance from the STRK contract and an exchange rate to WEI?

It should probably be done the same way, except we'll have to pass STRK token address to `get_balance`. I don't really understand why'd you want to convert to WEI though?

DelevoXDG (approved):
LGTM

starknet_py/net/account/account.py
ddoktorski: You're absolutely right, there is no need for conversion. I thought that `L1_GAS` in the resource bounds is specified in WEI, even though V3 transaction costs are paid in FRI, which is not the case. Apologies for the confusion.

We can indeed check it the same way as for V1, but with the STRK contract. On the other hand, as @DelevoXDG said, it is not crucial, since it will throw an error anyway. @THenry14 Wdyt?

THenry14: Agreed, let's update #1266 with the findings and do necessary changes under that issue at some later date

PR #1295, opened 2024-02-19
Change way of setting token_addres

### LINKED ISSUE #1264: Use union of chain and strring for custom networks
In the `Account` class we check in the functions `_default_token_address_for_chain` and `deploy_account` whether provided `chain` is a known network. Investigate if this approach makes any sense, if not then remove it from the code, otherwise consider implementing consistent approach that allows using custom networks.

tkumor3:
Closes #1264

## Introduced changes

- 
- 

##

### DIFF
--- a/starknet_py/net/account/account.py
+++ b/starknet_py/net/account/account.py
@@ -4,7 +4,7 @@ from collections import OrderedDict
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 from starknet_py.common import create_compiled_contract, create_sierra_compiled_contract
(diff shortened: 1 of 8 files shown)


starknet_py/net/account/account.py
```diff
@@ -102,7 +107,16 @@ def __init__(
                 account_address=self.address, key_pair=key_pair, chain_id=chain
             )
         self.signer: BaseSigner = signer
-        self._chain_id = chain
+
+        if token_address is not None:
+            self._token_address = token_address
+        else:
+            if hasattr(signer, "chain_id"):
+                self._token_address = default_token_address_for_chain(signer.chain_id)  # type: ignore
```
tkumor3, line 115: check if there is a better solution `# type: ignore`

PR #1290, opened 2024-02-16
add info about Account.estimate_fee()

tkumor3:
Closes #1283 

## Introduced changes

- 
- 

##

- [ ] This PR contains breaking changes

PR #1298, opened 2024-02-20
Add migration guide 

tkumor3:
Closes #

## Introduced changes

- 
- 

##

### DIFF
--- a/docs/migration_guide.rst
+++ b/docs/migration_guide.rst
@@ -5,6 +5,8 @@ Migration guide
 0.20.0 Migration guide
 **********************
 
+This update adds support for python 3.12
+
 Changes in the :class:`~starknet_py.net.account.account.Account`:
 
 .. currentmodule:: starknet_py.net.account.account
--- a/docs/migration_guide.rst
+++ b/docs/migration_guide.rst
@@ -2,6 +2,16 @@ Migration guide
 ===============
 
 **********************
+0.20.0 Migration guide
+**********************
+
+Changes in the :class:`~starknet_py.net.account.account.Account`:
+
+.. currentmodule:: starknet_py.net.account.account
+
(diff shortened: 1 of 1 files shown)


docs/migration_guide.rst
```diff
@@ -1,6 +1,18 @@
...
+0.20.0 Migration guide
+**********************
+
+This update adds support for python 3.12
+
+Changes in the :class:`~starknet_py.net.account.account.Account`:
+
+.. currentmodule:: starknet_py.net.account.account
+
+- :meth:`~Account.estimate_fee` has been added
```
ddoktorski, line 14: I would recommend against adding this information here, as it doesn't align to the context of the migration guide. Instead, it could be placed elsewhere:
- In the release notes
- In the guide under [Transaction Fee](https://starknetpy.readthedocs.io/en/latest/guide/account_and_client.html#transaction-fee) or as a separate section

ddoktorski (commented):
I think it would make sense to combine this with https://github.com/software-mansion/starknet.py/pull/1290, but leave that decision up to you.

docs/migration_guide.rst
```diff
@@ -1,6 +1,18 @@
 Migration guide
 ===============
 
+**********************
+0.20.0 Migration guide
+**********************
+
+This update adds support for python 3.12
```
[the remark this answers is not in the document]
tkumor3: Information about supporting python 3.12 has to be merged with https://github.com/software-mansion/starknet.py/pull/1282/files#diff-d97482026f068d245e2dbf3126057c340a49047f45ad012889bad449c130261eR8

docs/migration_guide.rst
tkumor3: I added this in https://github.com/software-mansion/starknet.py/pull/1290/files

Issue #1231, opened 2023-11-30
Refactor `client_models`


ddoktorski:
#1179 

Previously tracked in #347 and #1047

PR #1307, opened 2024-02-29
Support RPC 0.7.0

### LINKED ISSUE #1281: Add support for 0.7.0 RPC version

ddoktorski:
Closes #1281

Initial PR #1282 

## Introduced changes

- Replace `StarknetBlockCommon` with `BlockHeader` and add `PendingBlockHeader` to align with the RPC specification
  - Rename `parent_block_hash`, `root` -> `parent_hash`, `new_root`
  - Use `BlockHeader` in `StarknetBlock`, `StarknetBlockWithTxHashes` and `StarknetBlockWithReceipts`, the same for respective pending classes
- Change logic in auto estimation for V3 transactions in the `Account` class
- Add `FullNodeClient.get_block_with_receipts`
- Increase `max_fee`/`l1_resource_bounds` for devnet tests

##

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,7 +1,7 @@
 name: Checks
 
 env:
-  DEVNET_SHA: "1bd447d"
+  DEVNET_SHA: "79a90fd"
 
 on:
   push:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,8 +1,6 @@
 name: Checks
 
 env:
-  STARKNET_VERSION: "0.13.0"
-  RPC_SPEC_VERSION: "0.6.0"
   DEVNET_SHA: "1bd447d"
 
 on:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,7 +1,7 @@
 name: Checks
 
 env:
-  DEVNET_SHA: "79a90fd"
+  DEVNET_SHA: "c6ffb99"
 
 on:
   push:
--- a/docs/guide/account_and_client.rst
+++ b/docs/guide/account_and_client.rst
@@ -27,7 +27,8 @@ To enable auto estimation, set the ``auto_estimate`` parameter to ``True``.
 
     It is strongly discouraged to use automatic fee estimation in production code as it may lead to an unexpectedly high fee.
 
-The returned estimated fee is multiplied by ``1.5`` to mitigate fluctuations in price.
+The returned estimated fee is multiplied by ``1.5`` for V1 and V2 transactions to mitigate fluctuations in price.
+For V3 transactions, ``max_amount`` and ``max_price_per_unit`` are scaled by ``1.5`` and ``1.5`` respectively.
 
 .. note::
     It is possible to configure the value by which the estimated fee is multiplied,
--- a/docs/guide/account_and_client.rst
+++ b/docs/guide/account_and_client.rst
@@ -27,9 +27,7 @@ To enable auto estimation, set the ``auto_estimate`` parameter to ``True``.
 
     It is strongly discouraged to use automatic fee estimation in production code as it may lead to an unexpectedly high fee.
 
-The returned estimated fee is multiplied by ``1.5`` for V1 and V2 transactions to mitigate fluctuations in price.
-For V3 transactions, ``max_amount`` and ``max_price_per_unit`` are scaled by ``1.5``.
-
+The returned estimated fee is multiplied by ``1.5`` to mitigate fluctuations in price.
 
 .. note::
     It is possible to configure the value by which the estimated fee is multiplied,
--- a/docs/guide/account_and_client.rst
+++ b/docs/guide/account_and_client.rst
@@ -28,7 +28,7 @@ To enable auto estimation, set the ``auto_estimate`` parameter to ``True``.
     It is strongly discouraged to use automatic fee estimation in production code as it may lead to an unexpectedly high fee.
 
 The returned estimated fee is multiplied by ``1.5`` for V1 and V2 transactions to mitigate fluctuations in price.
-For V3 transactions, ``max_amount`` and ``max_price_per_unit`` are scaled by ``1.1`` and ``1.5`` respectively.
+For V3 transactions, ``max_amount`` and ``max_price_per_unit`` are scaled by ``1.5``.
 
 
 .. note::
--- a/docs/migration_guide.rst
+++ b/docs/migration_guide.rst
@@ -5,12 +5,7 @@ Migration guide
 0.21.0 Migration guide
 **********************
 
-
-Version 0.21.0 of **starknet.py** comes with support for RPC 0.7.0 and python 3.12!
-
-New classes added to mirror the recent changes in the RPC v0.7.0 specification include:
-:class:`StarknetBlockWithReceipts`, :class:`PendingStarknetBlockWithReceipts`, :class:`TransactionWithReceipt` :class:`DataResources`.
-
+Version 0.21.0 of **starknet.py** comes with support for RPC 0.7.0!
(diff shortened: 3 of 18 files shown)


ddoktorski:
Note: Some tests on network will fail, because the testnet networks don't support Starknet `0.13.1` yet.

docs/guide/account_and_client.rst
```diff
@@ -27,9 +27,7 @@ To enable auto estimation, set the ``auto_estimate`` parameter to ``True``.
 
     It is strongly discouraged to use automatic fee estimation in production code as it may lead to an unexpectedly high fee.
 
-The returned estimated fee is multiplied by ``1.5`` for V1 and V2 transactions to mitigate fluctuations in price.
-For V3 transactions, ``max_amount`` and ``max_price_per_unit`` are scaled by ``1.1`` and ``1.5`` respectively.
-
+The returned estimated fee is multiplied by ``1.5`` to mitigate fluctuations in price.
```
DelevoXDG, line 30: I think this is a bit misleading, as in case of V3 transactions the overall fee is actually multiplied by 1.5*1.5

```suggestion
The returned estimated fee is multiplied by ``1.5`` for V1 and V2 transactions to mitigate fluctuations in price.
For V3 transactions, ``max_amount`` and ``max_price_per_unit`` are scaled by ``1.5`` and ``1.5`` respectively.
```

.github/workflows/checks.yml
```diff
@@ -1,9 +1,7 @@
 name: Checks
 
 env:
-  STARKNET_VERSION: "0.13.0"
-  RPC_SPEC_VERSION: "0.6.0"
-  DEVNET_SHA: "1bd447d"
+  DEVNET_SHA: "79a90fd"
```
DelevoXDG, line 4: Just in case, you can use the commit from the main branch now:
https://github.com/0xSpaceShard/starknet-devnet-rs/commit/c6ffb99597eeea58d1a6a30fc061de69e586c779

docs/migration_guide.rst
```diff
@@ -1,6 +1,32 @@
...
+- Starknet - `0.13.1 <https://docs.starknet.io/documentation/starknet_versions/version_notes/#version0.13.1>`_
+- RPC - `0.7.0 <https://github.com/starkware-libs/starknet-specs/releases/tag/v0.7.0>`_
+
+0.21.0 Breaking changes
+-----------------------
+
+.. currentmodule:: starknet_py.net.client_models
+
+1. :class:`ExecutionResources` has a new required field ``data_availability``
+2. :class:`InvokeTransactionTrace`, :class:`DeclareTransactionTrace` and :class:`DeployAccountTransactionTrace` have a new required field ``execution_resources``
```
DelevoXDG, line 22: Should we consider addition of new fields changes breaking, assuming that we don't instantiate those manually?

starknet_py/net/account/account.py
```diff
@@ -165,7 +165,8 @@ async def _get_resource_bounds(
 
             l1_resource_bounds = ResourceBounds(
                 max_amount=int(
-                    estimated_fee.gas_consumed * Account.ESTIMATED_AMOUNT_MULTIPLIER
+                    (estimated_fee.overall_fee / estimated_fee.gas_price)
```
DelevoXDG, line 168: Should we account for `gas_price = 0` just in case? 

(0,0) resources can be used in `estimateFee` / `simulateTransactions`. It normally shouldn't happen, but I suppose better safe than sorry?

starknet_py/tests/e2e/block_test.py
```diff
@@ -58,11 +60,14 @@ async def test_block_with_tx_hashes_latest(
...
-    assert blk.parent_block_hash is not None
+    assert blk.parent_hash is not None
     assert blk.block_number is not None
-    assert blk.root is not None
+    assert blk.new_root is not None
     assert blk.timestamp is not None
     assert blk.sequencer_address is not None
+    assert blk.l1_gas_price is not None
+    assert blk.l1_data_gas_price is not None
+    assert blk.l1_da_mode is not None
```
DelevoXDG, line 70: Why check that these are not `None`? 

I assume as they're required that shouldn't ever happen, so maybe we should check the values instead, similarly to how it's done in `estimate_fee` tests below? No need to use specific values, `!= 0` should suffice for latest/pending blocks.

docs/migration_guide.rst
ddoktorski: Good question. To me it makes sense to classify the addition of new fields as a minor change and to consider the removal of existing keys a breaking change. @THenry14 wdyt?

starknet_py/tests/e2e/block_test.py
ddoktorski: I would consider these tests to be kind of a sanity check, given that we have limited capabilities to test the exact values. I've added slightly more specific asserts for `l1_gas_price`, `l1_data_gas_price` and `l1_da_mode`.

docs/migration_guide.rst
THenry14: yup, makes sense to me as well

starknet_py/net/schemas/rpc.py
```diff
@@ -444,67 +473,93 @@ class TypesOfTransactionsSchema(OneOfSchema):
...
     timestamp = fields.Integer(data_key="timestamp", required=True)
+    sequencer_address = Felt(data_key="sequencer_address", required=True)
     l1_gas_price = fields.Nested(
         ResourcePriceSchema(), data_key="l1_gas_price", required=True
     )
+    l1_data_gas_price = fields.Nested(
+        ResourcePriceSchema(), data_key="l1_data_gas_price", required=True
+    )
+    l1_da_mode = L1DAModeField(data_key="l1_da_mode", required=True)
     starknet_version = fields.String(data_key="starknet_version", required=True)
```
DelevoXDG, line 504: Nit: Would it be reasonable to add an intermediate class to avoid some code duplication?

DelevoXDG (approved):

starknet_py/net/schemas/rpc.py
tkumor3: On the other hand, it's "copied" from starknet-spec. IMO it's cleaner when we keep this basic spec class implement explicit

.github/workflows/checks.yml
```diff
@@ -1,9 +1,7 @@
 name: Checks
 
 env:
-  STARKNET_VERSION: "0.13.0"
-  RPC_SPEC_VERSION: "0.6.0"
-  DEVNET_SHA: "1bd447d"
+  DEVNET_SHA: "c6ffb99"
```
tkumor3, line 4: shouldn't we use devnet version: 230b74c? https://github.com/0xSpaceShard/starknet-devnet-rs/releases/tag/v0.0.3

starknet_py/net/schemas/rpc.py
```diff
@@ -1,3 +1,5 @@
+# pylint: disable=too-many-lines
```
tkumor3, line 1: we could think about splitting this file into the smallest. but in separate task

tkumor3 (approved):

.github/workflows/checks.yml
ddoktorski: It doesn't make any difference, as there is no relevant changes in the commits following `c6ffb99` (only changes to CI and package version).

starknet_py/net/schemas/rpc.py
ddoktorski: https://github.com/software-mansion/starknet.py/issues/1316

starknet_py/net/schemas/rpc.py
ddoktorski: We can consider introducing an additional class, such as `BlockHeaderCommon` in #1231

PR #1312, opened 2024-03-05
Ignoring unknown keys in RPC responses by default

### LINKED ISSUE #1305: Allow ignoring unknown keys in RPC responses
- Remove hardcoded `unknown=EXCLUDE` from deserializing in `FullNodeClient` methods
- Allow constructor argument like `exclude_unknown_fields` in `FullNodeClient`
- Pass the above argument when deserializing RPC responses, make sure that it is propagated to the nested schemas e.g. `TransactionReceipt` and field `execution_resources`

Similar change has been done in https://github.com/software-mansion/starknet-jvm/pull/405

tkumor3:
Closes #1305

- Remove hardcoded unknown=EXCLUDE from deserializing in FullNodeClient
- Ignoring unknown keys in RPC responses by default
- In our test check if the schemas are correct

I had to leave some `unknown=EXCLUDE` because devnet sometimes returns more fields than starknet-spec claim should be sent

## Introduced changes

- 
- 

##


tkumor3:
Rest of `unknown="EXCLUDE"` I will remove after merging RPC 0.7.0 development

starknet_py/net/full_node_client.py
```diff
@@ -167,11 +167,11 @@ async def get_block_with_receipts(
...
             return cast(
                 PendingStarknetBlockWithReceipts,
-                PendingStarknetBlockWithReceiptsSchema().load(res, unknown=EXCLUDE),
+                PendingStarknetBlockWithReceiptsSchema().load(res),
             )
         return cast(
             StarknetBlockWithReceipts,
-            StarknetBlockWithReceiptsSchema().load(res, unknown=EXCLUDE),
+            StarknetBlockWithReceiptsSchema().load(res),
         )
```
ddoktorski, line 176: Why `unknown=EXCLUDE` is not removed from all methods e.g. `get_block_with_txs`?

ddoktorski (commented):
Could we add some unit tests by setting a temporary environment variable?

starknet_py/net/full_node_client.py
tkumor3: The prime problem of this is working with devnet because in a few cases, they are inconsistent with starknet-spec.

tkumor3:
> Could we add some unit tests by setting a temporary environment variable?

It's problematic because we need to keep broken schema in our code to test it

starknet_py/net/client_models.py
```diff
@@ -950,6 +951,7 @@ class L1HandlerTransactionTrace:
     Dataclass representing a transaction trace of an L1_HANDLER transaction.
     """
 
+    execution_resources: Optional[ExecutionResources]
```
THenry14, line 954: @DelevoXDG please let's make sure we have this field in jvm/swift as well, as per latest discussion on swm-sw channel this is a spec bug

starknet_py/net/schemas/rpc.py
```diff
@@ -964,6 +964,11 @@ def make_dataclass(self, data, **kwargs) -> DeployAccountTransactionTrace:
 
 
 class L1HandlerTransactionTraceSchema(Schema):
+    # TODO (#1323): Explain with starknet, because spec doesn't contain execution_resources
```
THenry14, line 967: we can probably remove this comment/issue, as Ariel already admitted this to be a bug in spec that will be fixed soon

starknet_py/__init__.py
```diff
@@ -0,0 +1,3 @@
+from marshmallow import EXCLUDE, RAISE, Schema
+
+Schema.Meta.unknown = EXCLUDE  # type: ignore
```
THenry14, line 3: just to be double sure - you have checked that this won't overwrite a user setting in script, right? I mean:
```
# import marshmallow
# set Meta to whatever
# import starknet.py
# the rest of the script
```

so the importing starknet.py (and thus this file as well), does not overwrite the user setting

THenry14 (approved):
lgtm, can be merged after all comments are addressed

starknet_py/net/schemas/rpc.py
tkumor3: I am not sure how they will change L1HandlerTransactionTraceSchema. execution_resources can be required or optional. IMO is better to wait for space change

starknet_py/net/full_node_client.py
THenry14: you mean starknet-devnet-rs :+1: ok then, the issue of pending+latest blocks was raised a couple of times with the devnet team, so I think we should just wait

starknet_py/utils/schema.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+from marshmallow import EXCLUDE, RAISE
+from marshmallow import Schema as SchemaToCopy
+
+print(os.environ.get("marshmallow_raise"))
```
THenry14, line 6:
```suggestion
```

starknet_py/utils/schema.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class Schema(SchemaToCopy):
+    class Meta:
+        unknown = RAISE if os.environ.get("marshmallow_raise") else EXCLUDE
```
THenry14, line 11:
```suggestion
        unknown = RAISE if os.environ.get("STARKNET_PY_MARSHMALLOW_META_MODE") else EXCLUDE
```
or something similar. env variables should always be capital letters, and it may make sense to make it more descriptive + starknet.py specific

THenry14 (commented):
To sum up - the now default behaviour will be to configure everything with an env variable, because it is virtually impossible to do so in-script, right?

This PR should be updated with apropriate docs section that will explain this feature and give some examples.

tkumor3:
> This PR should be updated with apropriate docs section that will explain this feature and give some examples.
Any idea about an example, IMO the description how to set it seems to be clear

starknet_py/utils/schema.py
```diff
@@ -0,0 +1,13 @@
...
+
+from marshmallow import EXCLUDE, RAISE
+from marshmallow import Schema as MarshmallowSchema
+
+
+class Schema(MarshmallowSchema):
+    class Meta:
+        unknown = (
+            EXCLUDE
+            if os.environ.get("STARKNET_PY_MARSHMALLOW_UKNOWN_EXCLUDE")
```
THenry14, line 11: > This can be achieved by setting the environment variable, STARKNET_PY_MARSHMALLOW_UKNOWN_EXCLUDE,
to true. Consequently, any unknown fields in response will be excluded.

does this code checks this? I'm not sure, but I think it may just check if the env variable is set (so eg `STARKNET_PY_MARSHMALLOW_UKNOWN_EXCLUDE=false` may also set it, resulting in turning on the exclusion). let's double check.

THenry14:
> > This PR should be updated with apropriate docs section that will explain this feature and give some examples.
> 
> Any idea about an example, IMO the description how to set it seems to be clear

The description you've provided should be enough

starknet_py/utils/schema.py
```diff
@@ -0,0 +1,15 @@
...
+from marshmallow import Schema as MarshmallowSchema
+
+MARSHMALLOW_UKNOWN_EXCLUDE = os.environ.get("STARKNET_PY_MARSHMALLOW_UKNOWN_EXCLUDE")
+
+class Schema(MarshmallowSchema):
+    class Meta:
+        unknown = (
+            EXCLUDE
+            if MARSHMALLOW_UKNOWN_EXCLUDE
+            and MARSHMALLOW_UKNOWN_EXCLUDE.lower() == "true"
```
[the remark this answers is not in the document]
tkumor3: This change doesn't work `AttributeError: 'NoneType' object has no attribute 'lower'`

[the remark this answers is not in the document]
THenry14:
```
str(MARSHMALLOW_UKNOWN_EXCLUDE).lower() == "true"
```
or 
```
(MARSHMALLOW_UKNOWN_EXCLUDE or "").lower() == "true"
```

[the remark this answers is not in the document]
THenry14: nitpick though

THenry14 (approved):

PR #1313, opened 2024-03-06
Add parsed_abi property to SierraContractClass

### LINKED ISSUE #1306: `get_class_by_hash` returns a string in abi for Cairo 1 classes
### What happened

Trying to iterate over ABI items in Cairo 1 classes returns individual characters. Casting to json makes it possible to iterate on items. For Cairo 0, items are available w/o casting as expected.

### Stack trace

```
Traceback (most recent call last):
  File "/home/amanusk/Code/Starknet/starknet-python-wallet/cairo1_bug.py", line 37, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/home/amanusk/Code/Starknet/starknet-python-wallet/cairo1_bug.py", line 33, in main
    print(item["type"])
TypeError: string indices must be integer
```

### Steps to reproduce

```
    full_node_client = FullNodeClient(node_url)

    cairo_class = await full_node_client.get_class_by_hash(
        "0x054262ACDDA5A51FB77B3D7A3E977CFE056F34FAF5E75E793AFEA58026AA7470"
    )

    items = cairo_class.abi
    # items = json.loads(cairo_class.abi)
    for item in items:
        print(item["type"])
```

### SDK Version

0.19.0

### Python version

3.19

Linux

tkumor3:
Closes #1306

- ~change SierraCompiledContract abi type from string to AbiDictList and use marshmallow to verify it~
- ~add `raw_abi` abi in string. It's needed `compute_sierra_class_hash`~
- CHANGE: add to SierraCompiledContract `parsed_abi`

## Introduced changes

- 
- 

##


starknet_py/net/client_models.py
```diff
@@ -711,7 +724,28 @@ class SierraContractClass:
     contract_class_version: str
     sierra_program: List[str]
     entry_points_by_type: SierraEntryPointsByType
-    abi: Optional[str] = None
+    raw_abi: Optional[str] = None
```
ddoktorski, line 727: Maybe let's leave the name `abi` here to align with [the RPC returned type](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3291)? Then we can rename the method to something like `parse_abi`/`parsed_abi`.

starknet_py/net/schemas/gateway.py
```diff
@@ -116,7 +116,7 @@ def _deserialize(self, value, attr, data, **kwargs):
 
 
 class SierraCompiledContractSchema(SierraContractClassSchema):
-    abi = AbiField(data_key="abi", required=True)
+    raw_abi = AbiField(data_key="abi", required=True)
```
ddoktorski, line 119: I believe there is not reason to have `AbiField` here instead of `fields.String`, but please double check.

ddoktorski (commented):
Could you please fix lint, failing docs tests and the PR title? It would be also beneficial to add some tests for this change. Generally this fix looks good to me.

starknet_py/net/client_models.py
tkumor3: I added it from perspective `get_class_by_hash`, the goal was to make api `ContractClass` and `SierraContractClass` as similar as possible. TBH I haven't got a strong opinion which approach I prefer 
```
    async def get_class_by_hash(
        self, class_hash: Hash
    ) -> Union[ContractClass, SierraContractClass]:
```

starknet_py/net/schemas/gateway.py
tkumor3: no idea why, but I get `marshmallow.exceptions.ValidationError: {'abi': ['Not a valid string.']}`

ddoktorski: Is there any specific test that throws this error?

starknet_py/net/client_models.py
ddoktorski: It also makes sense. @THenry14 @DelevoXDG what are your thoughts on this?

starknet_py/net/schemas/gateway.py
tkumor3: Screenshot 2024-03-08 at 09 15 51
Many of them

tkumor3: Screenshot 2024-03-08 at 09 29 12

THenry14: tip: let's not use screenshots but code blocks instead. makes it way easier for a reviewer to find the places you're reffering to in code.

tkumor3: I investigated it, and my conclusion is: SierraContractClassSchema we use to map [model from RPC](https://github.com/starkware-libs/starknet-specs/blob/eedf5f899aa51a85a841333175023aa5d615aa33/api/starknet_api_openrpc.json#L3240), so abi is string. SierraCompiledContractSchema we use to load compiled contract from a file, so abi is Dict (maybe str)

THenry14 (commented):
this seems like a breaking change, I suppose we should include this change in the migration guide. otherwise looks good.

ddoktorski (approved):
LGTM, please update the PR title before merging to something like "Add `parsed_abi` property to `SierraContractClass`"

docs/migration_guide.rst
```diff
@@ -1,6 +1,16 @@
...
 
+******************************
+0.21.0 Migration guide
+******************************
+
+
+0.21.0 Minor changes
+-----------------------
+
+1. :class:`starknet_py.net.client_models.SierraContractClass` has an additional propery ``parsed_abi``
```
ddoktorski, line 12: This should be added below, there is already the section for `0.21.0`. Before the release we only need to remove "alpha" from the header.

PR #1326, opened 2024-03-18
Compile contract with scarb

### LINKED ISSUE #1321: Move compiling contracts on github actions
Investigate if we need contracts compiled with the obsolete Cairo version (using deployed contract on integration/testnet might be enough). Use Scarb to compile contracts.

tkumor3:
Closes #1321

## Introduced changes

- Compile contracts on CI with Scarb
- Update contract names to be consistent

##


starknet_py/tests/e2e/fixtures/misc.py
```diff
@@ -94,6 +99,47 @@ def get_tx_status_full_node_client():
...
+    if version == ContractVersion.V1:
+        artifacts_path = CONTRACTS_V1_ARTIFACTS
+        directory = CONTRACTS_V1_COMPILED
+
+    loaded_artifcats = json.loads((artifacts_path).read_text("utf-8"))
+
+    for item in loaded_artifcats["contracts"]:
+        if item["contract_name"] == contract_name:
+            artifacts = item["artifacts"]
+            break
```
ddoktorski, line 130: Let's try to write this in a more idiomatic way for Python. The suggestion below is one option, but there may be better ones.
```suggestion
    artifacts = next((item["artifacts"] for item in loaded_artifacts["contracts"] if item["contract_name"] == contract_name), None)
```

.github/workflows/manual_tests.yml
ddoktorski, line 1: I am not sure if we need this workflow at all

starknet_py/tests/e2e/fixtures/misc.py
```diff
@@ -94,6 +99,47 @@ def get_tx_status_full_node_client():
...
+            f"Artifacts for contract {contract_name} haven't be founded"
+        )
+
+    sierra = (directory / artifacts["sierra"]).read_text("utf-8")
+    casm = (directory / artifacts["casm"]).read_text("utf-8")
+
+    return {"casm": casm, "sierra": sierra}
+
+
 def read_contract(file_name: str, *, directory: Optional[Path] = None) -> str:
```
ddoktorski, line 143: Can we somehow combine `load_contract` and `read_contract` to have only one function for this?

.github/workflows/checks.yml
```diff
@@ -96,54 +96,24 @@ jobs:
...
 
-      # ====================== CONTRACTS v2 ====================== #
-
-      - name: Cache contracts v2
-        id: cache-contracts-v2
-        uses: actions/cache@v3
+      - uses: software-mansion/setup-scarb@v1.3.2
         with:
-          path: starknet_py/tests/e2e/mock/contracts_compiled_v2
-          key: ${{ runner.os }}-contracts-${{ hashFiles('starknet_py/tests/e2e/mock/contracts_v2') }}-${{ env.CAIRO_V2_SHA }}
```
THenry14, line 106: `CAIRO_V2_SHA` (and other envs added in previous PR in stack) is no longer needed, let's remove them

.github/workflows/checks.yml
```diff
@@ -96,54 +96,24 @@ jobs:
         run: |
           poetry install
 
-      # ====================== CONTRACTS v2 ====================== #
-
-      - name: Cache contracts v2
-        id: cache-contracts-v2
```
THenry14, line 102: I think we might want to cache artifacts to speed CI up

docs/development.rst
```diff
@@ -10,6 +10,7 @@ Development dependencies
 - `poetry <https://python-poetry.org/>`_ - dependency manager.
 - `pyenv <https://github.com/pyenv/pyenv>`_ - recommended for installing and switching python versions locally.
 - `cairo-lang <https://pypi.org/project/cairo-lang/>`_ - required to compile contracts  (`poe compile_contracts`)
+- `asdf <https://asdf-vm.com/>`_ - required to compile contracts  (`poe compile_contracts`)
```
THenry14, line 13:
```suggestion
- `asdf <https://asdf-vm.com/>`_ - required to install `scarb`, that is used for contracts compilation  (`poe compile_contracts`)
```

starknet_py/tests/e2e/fixtures/misc.py
```diff
@@ -13,6 +14,10 @@
     CONTRACTS_COMPILED_V0_DIR,
     CONTRACTS_COMPILED_V1_DIR,
     CONTRACTS_COMPILED_V2_DIR,
+    CONTRACTS_V1_ARTIFACTS,
+    CONTRACTS_V1_COMPILED,
```
THenry14, line 18: what is the difference between `artifacts` and `compiled`? I though artifacts are just compiled contracts?

starknet_py/tests/e2e/fixtures/misc.py
```diff
@@ -94,6 +99,47 @@ def get_tx_status_full_node_client():
...
+            version = ContractVersion.V1
+        if "--contract_dir=v2" in sys.argv:
+            version = ContractVersion.V2
+
+    artifacts_path = CONTRACTS_V2_ARTIFACTS
+    directory = CONTRACTS_V2_COMPILED
+
+    if version == ContractVersion.V1:
+        artifacts_path = CONTRACTS_V1_ARTIFACTS
+        directory = CONTRACTS_V1_COMPILED
```
THenry14, line 123: lets use if..else if..else to handle this logic. we will be more explicit, and avoid overwriting variables

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,70 @@
...
+    if ! command -v asdf >/dev/null 2>&1; then
+        echo "Install asdf, before executing this script is required to manage scarb version"
+        exit 1
+    fi
 
-# delete all artifacts except precompiled ones
-find $CONTRACTS_COMPILED_DIRECTORY -maxdepth 1 -type f -delete
+    asdf plugin-add scarb
+    asdf install scarb 0.4.1
+    asdf install scarb 2.6.3
```
THenry14, line 11: it may be troublesome to remember all the places that define the versions. Ideally, this script would hold paths to .tool_versions and take them directly from the file, so you only change scarb version in one place

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,70 @@
...
 
-# compile Cairo test contracts
-echo "Compiling Cairo contracts with $(poetry run starknet-compile-deprecated --version)"
+    output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
+        echo "Running \"scarb fmt\" in directory: $PWD"
+        find target/dev -maxdepth 1 -type f -delete
+        scarb build
+    ' \;)
```
THenry14, line 17: this is overcomplicated - it is suboptimal to run find inside a find. I think scarb recompiles at each run, so we probably should not bother with artifacts removal

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,70 @@
...
 
-number_of_contracts=0
-for contract in "$CONTRACTS_DIRECTORY"/*.cairo; do
-    basename=$(basename "$contract")
+    echo "$output"
+    if grep -iq "Diff" <<<"$output"; then
```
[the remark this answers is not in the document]
THenry14: I *think* this is supposed to check the scarb status, to determine if scarb build suceeded? But I don't think this will work tbh, we should check the status of build directly

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,70 @@
...
+    output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
+        echo "Running \"scarb fmt\" in directory: $PWD"
+        find target/dev -maxdepth 1 -type f -delete
+        scarb build
+    ' \;)
 
-number_of_contracts=0
-for contract in "$CONTRACTS_DIRECTORY"/*.cairo; do
-    basename=$(basename "$contract")
+    echo "$output"
```
THenry14, line 19: useless echo

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,70 @@
...
+    if grep -iq "Diff" <<<"$output"; then
+        exit 1
+    fi
 
-    output=$CONTRACTS_COMPILED_DIRECTORY/"${basename%.*}_compiled.json"
-    abi=$CONTRACTS_COMPILED_DIRECTORY/"${basename%.*}_abi.json"
+    echo "Compiled $number_of_contracts Cairo1 files successfully"
```
THenry14, line 24:
```suggestion
    echo "Successfully compiled contracts!"
```
or something along the lines

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,70 @@
...
+            account_contract_flag="--account_contract"
+        fi
+
+        echo "Compiling $contract..."
+        # run starknet-compile-deprecated
+        poetry run starknet-compile-deprecated $account_contract_flag --cairo_path $CONTRACTS_DIRECTORY:$MOCK_DIRECTORY --output $output --abi $abi $contract
+        number_of_contracts=$((number_of_contracts + 1))
+    done
+
+    echo "Compiled $number_of_contracts Cairo files successfully"
```
THenry14, line 58: the same, I don't think we need the number of contracts, this is useless information and brings unnecessary complexity to the code

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,70 @@
...
-    # run starknet-compile-deprecated
-    poetry run starknet-compile-deprecated $account_contract_flag --cairo_path $CONTRACTS_DIRECTORY:$MOCK_DIRECTORY --output $output --abi $abi $contract
-    number_of_contracts=$((number_of_contracts+1))
-done
 
-echo "Compiled $number_of_contracts Cairo files successfully"
+if [ -z "$@" ]; then
+    compile_contracts_with_scarb
+    compile_cairo_v0
+fi
```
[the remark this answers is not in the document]
THenry14: Imo we should take it a step further.

the compile_contracts_with_scarb should take an argument that specifies the version to be compiled. 
Similarly, the script itself should take an optional argument that specifies the version. We should be then able to either compile specific version only (eg v0, v1, v2) or all versions at once (when no version is specified).

starknet_py/tests/e2e/mock/contracts_v1/Scarb.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
THenry14, line 1: Should the Scarb.lock for the file be empty? of so, should it be included in the repo?

starknet_py/tests/e2e/mock/compile_contracts.sh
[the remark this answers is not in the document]
tkumor3: I assume we have two scenarios, set up the project and compile contract v0 on CI. The third is to install a new Scarb and update the local v2 contract after the release new Cairo version. I will split this script

starknet_py/tests/e2e/fixtures/misc.py
tkumor3: `starknet_artifacts.json` contains information about all compiled contracts in Scarb project, use it to get paths to my contract by they names

THenry14: This file is a map, that resides in target directory. This still does not explain why you need `CONTRACTS_V1_ARTIFACTS` and `CONTRACTS_V1_COMPILED`

THenry14 (changes requested):

.github/workflows/manual_tests.yml
tkumor3: we don't run it on CI, I so lets remove it

starknet_py/tests/e2e/mock/contracts_v1/Scarb.toml
tkumor3: IMO remove it, but no strong opinion

starknet_py/tests/e2e/fixtures/misc.py
tkumor3:
```
CONTRACTS_V2_COMPILED = MOCK_DIR / "contracts_v2/target/dev"
CONTRACTS_V2_ARTIFACTS = CONTRACTS_V2_COMPILED / "contracts_v2.starknet_artifacts.json"

CONTRACTS_V1_COMPILED = MOCK_DIR / "contracts_v1/target/dev"
CONTRACTS_V1_ARTIFACTS = CONTRACTS_V1_COMPILED / "contracts_v1.starknet_artifacts.json"
```

It seems to me to be cleaner.  `contracts_v1.starknet_artifacts.json` the project name is included in this starknet_artifacts file name

starknet_py/tests/e2e/fixtures/misc.py
tkumor3: I simplified `read_contract` and left both functions, let me know, what do you think

tkumor3:
@THenry14 @ddoktorski what do you think about using an exeternal oppenzeppelin contract in our tests
```
openzeppelin = { git = "https://github.com/OpenZeppelin/cairo-contracts.git", branch="main" }

[[target.starknet-contract]]
build-external-contracts = [
    "openzeppelin::presets::account::Account",
    "openzeppelin::presets::erc20::ERC20"
]
```

THenry14 (commented):
leaving a comment so you can re-request review when all changes are ready

starknet_py/tests/e2e/fixtures/contracts.py
```diff
@@ -340,11 +339,13 @@ async def account_with_validate_deploy_class_hash(
 async def argent_cairo1_account_class_hash(
     pre_deployed_account_with_validate_deploy: BaseAccount,
 ) -> int:
+    # Use precompiled argent account contracts
+    # we don't have the source code for this contract."
```
THenry14, line 343:
```suggestion
    # we don't have the source code for this contract
```

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,81 @@
...
-CONTRACTS_DIRECTORY="$MOCK_DIRECTORY"/contracts
-CONTRACTS_COMPILED_DIRECTORY="$MOCK_DIRECTORY"/contracts_compiled
+CONTRACTS_DIRECTORY_V1="$MOCK_DIRECTORY"/contracts_v1
+CONTRACTS_DIRECTORY_V2="$MOCK_DIRECTORY"/contracts_v2
 
-# delete all artifacts except precompiled ones
-find $CONTRACTS_COMPILED_DIRECTORY -maxdepth 1 -type f -delete
+setup_asdf() {
+    if ! command -v asdf >/dev/null 2>&1; then
+        echo "Install asdf, before executing this script is required to manage scarb version"
```
THenry14, line 9:
```suggestion
        echo "asdf not found in PATH! Install asdf and run this script again"
```

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,81 @@
...
+        exit 1
+    fi
+    asdf plugin-add scarb
+}
 
-# compile Cairo test contracts
-echo "Compiling Cairo contracts with $(poetry run starknet-compile-deprecated --version)"
+compile_contracts_with_scarb() {
+    CONTRACTS_DIRECTORY=$1
+    SCARB_WITH_VERSION=$(cat "$CONTRACTS_DIRECTORY/.tool-versions")
```
THenry14, line 17: useless use of cat

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,81 @@
...
 
-number_of_contracts=0
-for contract in "$CONTRACTS_DIRECTORY"/*.cairo; do
-    basename=$(basename "$contract")
+    setup_asdf
```
THenry14, line 19: you add a plugin and install scarb each time the script is invoked, this has to be changed. Let's move this to another function (eg `ensure_scarb`?) that will check if required scarb version (and asdf plugin) is installed, and only installs it if it is not.

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,81 @@
...
+    asdf install $SCARB_WITH_VERSION
 
-    # set account contract flag
-    account_contract_flag=""
-    if [[ $basename == *"account"* ]]; then
-      account_contract_flag="--account_contract"
-    fi
+    echo "Compiling Cairo contracts with $SCARB_WITH_VERSION"
+
+    (cd $CONTRACTS_DIRECTORY && scarb clean && scarb build)
```
THenry14, line 25: let's not run those commands in a different sunshell, there is no need to that.
instead cd we should use pushd and popd

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,81 @@
...
+    compile_contracts_v0
+fi
+
+if [[ "$@" =~ .*"v1".* ]]; then
+    compile_contracts_with_scarb $CONTRACTS_DIRECTORY_V1
+fi
+
+if [[ "$@" =~ .*"v2".* ]]; then
+    compile_contracts_with_scarb $CONTRACTS_DIRECTORY_V2
+fi
```
THenry14, line 77: Let's check a $1 directly, as this is the only argument we will allow to the script. no need to check the whole array elements.

Let's also use case instead of if-s

THenry14 (changes requested):

starknet_py/tests/e2e/mock/compile_contracts.sh
tkumor3: I'm not sure how you want to do that, sth like:
```
SCARB_WITH_VERSION=$(<"$CONTRACTS_DIRECTORY/.tool-versions")
```

THenry14:
```
SCARB_WITH_VERSION="${CONTRACTS_DIRECTORY}/.tool-versions"
```

starknet_py/tests/e2e/mock/compile_contracts.sh
tkumor3: I assume that asdf manages it. In case when plugin and scarb are instaled asdf returns 
```
Plugin named scarb already added
scarb 2.6.3 is already installed
```

But if it's not enough I will add `ensure_scarb`

starknet_py/tests/e2e/mock/compile_contracts.sh
tkumor3: Maybe I miss sth but I want to take the content of this `.tool-versions` file 
```
    SCARB_WITH_VERSION_A=$(<"$CONTRACTS_DIRECTORY/.tool-versions")
    SCARB_WITH_VERSION="$CONTRACTS_DIRECTORY/.tool-versions"
    echo "SCARB_WITH_VERSION_A"
    echo $SCARB_WITH_VERSION_A
    echo "SCARB_WITH_VERSION"
    echo $SCARB_WITH_VERSION
```
returns

```
SCARB_WITH_VERSION_A
scarb 2.6.3
SCARB_WITH_VERSION
starknet_py/tests/e2e/mock/contracts_v2/.tool-versions
```

THenry14: sure, in that case:

```
SCARB_VERSION=$(awk '/scarb/ {print $2}' ${CONTRACTS_DIRECTORY}/.tool-versions)
```

should work

THenry14: this prints just a version, if you need to have the whole `scarb x.y.z` string, just print $0 in awk

starknet_py/tests/e2e/mock/compile_contracts.sh
THenry14: let's not depend on external binary (asdf) with that. check the installed versions (should be rather easy using asdf), and only act depending on the output

starknet_py/tests/e2e/mock/compile_contracts.sh
tkumor3: version is enough, thx!

.github/workflows/checks.yml
```diff
@@ -97,53 +95,43 @@ jobs:
...
+      - uses: software-mansion/setup-scarb@v1.3.2
+        with:
+          tool-versions: "starknet_py/tests/e2e/mock/contracts_v2/.tool-versions"
+
+      - name: Compile contracts v2
+        working-directory: "starknet_py/tests/e2e/mock/contracts_v2"
+        if: steps.cache-contracts_v2.outputs.cache-hit != 'true'
         run: |
-          poetry run poe compile_contracts_v2
+          scarb build
```
THenry14, line 114: it may be a good idea to use our new script to build contracts on CI as well

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,85 @@
 #!/bin/bash
-
+set -e
 MOCK_DIRECTORY=starknet_py/tests/e2e/mock
-CONTRACTS_DIRECTORY="$MOCK_DIRECTORY"/contracts
-CONTRACTS_COMPILED_DIRECTORY="$MOCK_DIRECTORY"/contracts_compiled
-
-# delete all artifacts except precompiled ones
-find $CONTRACTS_COMPILED_DIRECTORY -maxdepth 1 -type f -delete
+CONTRACTS_DIRECTORY_V1="$MOCK_DIRECTORY"/contracts_v1
```
THenry14, line 4:
```suggestion
CONTRACTS_DIRECTORY_V1="$MOCK_DIRECTORY/contracts_v1"
```
Always double quote whole variables :+1:

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,85 @@
...
+CONTRACTS_DIRECTORY_V2="$MOCK_DIRECTORY"/contracts_v2
 
-# compile Cairo test contracts
-echo "Compiling Cairo contracts with $(poetry run starknet-compile-deprecated --version)"
+setup_scarb() {
+    SCARB_VERSION=$1
```
THenry14, line 8:
```suggestion
    SCARB_VERSION="$1"
```

Please double quote variables to prevent globbing or splitting.
This comment applies to all variables you use in this script

.github/workflows/checks.yml
tkumor3: It works!

.github/workflows/checks.yml
```diff
@@ -97,53 +95,32 @@ jobs:
...
         with:
-          path: starknet_py/tests/e2e/mock/contracts_compiled_v2
-          key: ${{ runner.os }}-contracts-${{ hashFiles('starknet_py/tests/e2e/mock/contracts_v2') }}-${{ env.CAIRO_V2_SHA }}
-    
-      - uses: actions/checkout@v4
-        with:
-          repository: starkware-libs/cairo
-          path: cairo_v2
-          ref: ${{ env.CAIRO_V2_SHA }}
+          path: starknet_py/tests/e2e/mock/contracts_v2/target
```
THenry14, line 102: are you sure only `target` directory must be cached? what about the lock?

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -1,32 +1,84 @@
 #!/bin/bash
+set -e
 
-MOCK_DIRECTORY=starknet_py/tests/e2e/mock
-CONTRACTS_DIRECTORY="$MOCK_DIRECTORY"/contracts
-CONTRACTS_COMPILED_DIRECTORY="$MOCK_DIRECTORY"/contracts_compiled
+MOCK_DIRECTORY="starknet_py/tests/e2e/mock"
```
THenry14, line 4: nitpick: this script will be only runnable from the repo root atm. Maybe we should do `MOCK_DIRECTORY="$(git rev-parse --show-toplevel)/starknet_py/tests/e2e/mock"` so it can be run from anywhere in the repo?

.github/workflows/checks.yml
tkumor3: Because of adding external contracts, I have changed my mind and the best solution will be to add scarb.lock to the repo to ensure that we use the same contracts locally and on CI. Then caching only `target` should be enough. WDYT?

THenry14: hmm, may make sense actually. I can see you're using file hashes for cac he keys, so this should work. :+1:

THenry14 (approved):

PR #1327, opened 2024-03-20
Simplifies using a custom network

### LINKED ISSUE #1264: Use union of chain and strring for custom networks

tkumor3:
Closes #1264 

## Introduced changes

- Account constructor map string chainId in to int
- Extends chainId type in code 

##


starknet_py/net/models/transaction.py
```diff
@@ -63,7 +63,7 @@ def type(self) -> TransactionType:
         """
 
     @abstractmethod
-    def calculate_hash(self, chain_id: StarknetChainId) -> int:
+    def calculate_hash(self, chain_id: Union[int, StarknetChainId]) -> int:
```
ddoktorski, line 66: I think in internal functions we can just use `chain_id: int` if pyright doesn't complain about it.

starknet_py/net/account/account.py
```diff
@@ -86,6 +86,10 @@ def __init__(
         self._address = parse_address(address)
         self._client = client
         self._cairo_version = None
+        if isinstance(chain, str):
+            self._chain_id = int(chain, 16)
```
ddoktorski, line 90: I am not sure if we should expect a hex string here. Allowing a `str` to be passed suggests that the user can for example pass `SN_GOERLI` and in my opinion we should treat it that way.

starknet_py/net/account/account.py
```diff
@@ -95,14 +99,13 @@ def __init__(
                 raise ValueError(
                     "Either a signer or a key_pair must be provided in Account constructor."
                 )
-            if chain is None:
+            if self._chain_id is None:
```
ddoktorski, line 102: Maybe let's fetch chain id if it's `None`?

starknet_py/net/account/account.py
tkumor3: The main reason why I want to do that is that our method `get_chain_id` returns `hex string` and we also want to support strings for example "KATANA". Even if someone passes `SN_GOERLI` it will work. I know it isn't the best solution, but we have some examples in docs on how to use it correctly with enym

starknet_py/net/account/account.py
tkumor3: No strong opinion, on the other hand, is ok to demand users, to explicitly set things like this, It might protect from some unconscious actions

starknet_py/net/account/account.py
ddoktorski: In the current solution if a string is passed, for example `chain="KATANA"` then it will throw an error `ValueError: invalid literal for int() with base 16: 'KATANA'`.

tkumor3: Thx for catch it! I pushed fix

starknet_py/net/account/account.py
ddoktorski: In my opinion it makes sense to allow passing it, but always fetch it if not passed. At the moment I don't see any potential issues that it may cause.

starknet_py/net/account/account.py
```diff
@@ -86,6 +86,13 @@ def __init__(
         self._address = parse_address(address)
         self._client = client
         self._cairo_version = None
+        if isinstance(chain, str):
+            try:
+                self._chain_id = int(chain, 16)
+            except ValueError:
+                self._chain_id = int(chain.encode().hex(), 16)
+        else:
+            self._chain_id = chain
```
ddoktorski, line 95: Please add some tests for this logic in `net/account/account_test.py`

starknet_py/net/account/account.py
```diff
@@ -72,7 +72,7 @@ def __init__(
         client: Client,
         signer: Optional[BaseSigner] = None,
         key_pair: Optional[KeyPair] = None,
-        chain: Optional[StarknetChainId] = None,
+        chain: Optional[Union[StarknetChainId, str, int]] = None,
```
ddoktorski, line 75: nit: For readability we can extract this type to something like `ChainId = Union[StarknetChainId, str, int]`, then it cal be also used it in `StarkCurveSigner`

tkumor3: the problem is that StarkCurveSigner `chain_id: Union[StarknetChainId, int]` could be simplified to `chain_id: int` but I assume that we want to explicitly show StarknetChainId

starknet_py/net/account/account_test.py
```diff
@@ -88,27 +88,62 @@ def test_create_account():
...
+        address=0x2,
+        client=FullNodeClient(node_url=""),
+        key_pair=key_pair,
+        chain="0x534e5f5345504f4c4941",
+    )
+
+    assert account.address == 0x2
+    assert account.signer.public_key == key_pair.public_key
+    assert isinstance(account.signer, StarkCurveSigner)
+    assert account.signer.chain_id == 393402133025997798000961
```
ddoktorski, line 132: nit:
```suggestion
    assert account.signer.chain_id == 0x534e5f5345504f4c4941
```

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -884,3 +884,34 @@ async def test_account_execute_v3(account, deployed_balance_contract):
         call=get_balance_call
     )
     assert initial_balance + 100 == balance_after_increase
+
+
+@pytest.mark.asyncio
+async def test_execute_v3_with_get_chain_id(
```
ddoktorski, line 890: What is the rationale behind this test?

starknet_py/net/account/account.py
ddoktorski: In that case let's leave `chain_id: Union[StarknetChainId, int]` in `StarkCurveSigner`.

ddoktorski (changes requested):
In the scope of this PR please also go through functions in the `Account` class and remove `chain_id` from function parameters (e.g `get_balance`). For static methods, make an additional call to get chain id.

starknet_py/tests/e2e/account/account_test.py
tkumor3: It was the first test I added and the aim was to check if I could init Account with get_chain_id response. I've forgotten to remove it. thx for catch

starknet_py/net/account/account.py
```diff
@@ -72,7 +73,7 @@ def __init__(
         client: Client,
         signer: Optional[BaseSigner] = None,
         key_pair: Optional[KeyPair] = None,
-        chain: Optional[StarknetChainId] = None,
+        chain: Optional[Chain] = None,
```
THenry14, line 76: nit: maybe ChainId would be slightly better name?

PR #1328, opened 2024-03-21
Remove Goerli and migrate network tests on sepolia

tkumor3:
Closes #

- Remove support for Goerli
- run network test on sepolia
##

- Remove StarknetChainId GOERLI


starknet_py/tests/e2e/tests_on_networks/client_test.py
```diff
@@ -132,40 +132,41 @@ async def test_transaction_not_received_invalid_nonce(account_goerli_testnet):
...
-        to_addr=int(EMPTY_CONTRACT_ADDRESS_GOERLI_TESTNET, 0),
+        to_addr=int(EMPTY_CONTRACT_ADDRESS_SEPOLIA_TESTNET, 0),
         selector=get_selector_from_name("empty"),
         calldata=[],
     )
     sign_invoke = await account.sign_invoke_v1(calls=call, max_fee=int(1e16))
     sign_invoke = dataclasses.replace(sign_invoke, signature=[0x21, 0x37])
-
-    with pytest.raises(ClientError, match=r"Signature.*is invalid") as exc:
+    # 0x0000000000004163636f756e743a20696e76616c6964207369676e6174757265 -> Account: invalid signature
```
tkumor3, line 144: Sepolia returns only hex version of this string

starknet_py/tests/e2e/tests_on_networks/client_test.py
```diff
@@ -371,22 +374,23 @@ async def test_get_block_with_tx_hashes_new_header_fields(client_goerli_testnet)
     "tx_hash, tx_type",
     [
         (
-            0x7B31376D1C4F467242616530901E1B441149F1106EF765F202A50A6F917762B,
+            0x054270D103C875A613E013D1FD555EDCFF2085FECA9D7B4532243A8257FD5CF3,
             DeclareTransactionV3,
         ),
+        # Find DeployAccountTransactionV3 and uncomment it with correct hash
```
tkumor3, line 380: Do we have our account deployed as v3 transaction @THenry14 @ddoktorski

starknet_py/tests/e2e/tests_on_networks/trace_api_test.py
```diff
@@ -22,48 +24,46 @@
 
 
 @pytest.mark.asyncio
-async def test_trace_transaction(client_goerli_integration):
+async def test_trace_transaction(client_sepolia_integration):
```
tkumor3, line 27: Should we create a block with all of those transactions? do we need it?

starknet_py/tests/e2e/tests_on_networks/client_test.py
```diff
@@ -237,39 +239,39 @@ async def test_estimate_message_fee_throws(
...
-        # deploy_account: https://integration.voyager.online/tx/0x593c073960140ab7af7951fadb6a129572cc504ef0b9107992c5c1efe5a0fb5
-        (307054, 1),
-        # invoke: https://integration.voyager.online/tx/0x6225b92ce88603645e42fc4b664034f788ec9f01a5aadd9855646dd721898e5
-        (307163, 0),
-        # l1_handler: https://integration.voyager.online/tx/0x66e2db10edbed4b262e01ee0f89ff77907f9ca1b4fe11603d691f16370248f7
-        (307061, 3),
+        # declare: https://integration-sepolia.starkscan.co/tx/0x0544a629990d2bed8ccf11910b30f2f1e18228ed5be06660bea20cae13b2aced
+        (9707, 0),
+        # deploy: https://integration-sepolia.starkscan.co/tx/0x022ffb771d8b847899c49f8efe48b4d70fd4825658e3d41ad38d3d3da3045891
+        (9708, 2),
```
ddoktorski, line 261: It can also be removed.

starknet_py/tests/e2e/tests_on_networks/client_test.py
ddoktorski: Nope, but we can easily create `DEPLOY_ACCOUNT_V3` transaction by using `Account.deploy_account_v3` static function.

starknet_py/tests/e2e/tests_on_networks/trace_api_test.py
ddoktorski: We can try to find a block that has many different transaction types, not necessarily all the possible types. It should be easier to find on testnet.

ddoktorski (changes requested):

starknet_py/tests/e2e/tests_on_networks/trace_api_test.py
tkumor3: I found block with various transactions on integration. I have all this 
```
        InvokeTransactionV1, InvokeTransactionV3, DeployAccountTransactionV1,L1HandlerTransaction
```
It's not easy to find sth better on Testnet or I'm not able to do that :)

starknet_py/tests/e2e/tests_on_networks/client_test.py
```diff
@@ -132,40 +132,41 @@ async def test_transaction_not_received_invalid_nonce(account_goerli_testnet):
...
     )
     sign_invoke = await account.sign_invoke_v1(calls=call, max_fee=int(1e16))
     sign_invoke = dataclasses.replace(sign_invoke, signature=[0x21, 0x37])
-
-    with pytest.raises(ClientError, match=r"Signature.*is invalid") as exc:
+    # 0x0000000000004163636f756e743a20696e76616c6964207369676e6174757265 -> Account: invalid signature
+    with pytest.raises(
+        ClientError,
+        match=r"0x0000000000004163636f756e743a20696e76616c6964207369676e6174757265",
+    ) as exc:
```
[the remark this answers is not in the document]
tkumor3: Isn't it too general?

.github/workflows/checks.yml
```diff
@@ -202,14 +202,12 @@ jobs:
     strategy:
       fail-fast: false
     env:
-      GOERLI_INTEGRATION_RPC_URL: ${{ secrets.INTEGRATION_RPC_URL }}
-      GOERLI_TESTNET_RPC_URL: ${{ secrets.TESTNET_RPC_URL }}
       SEPOLIA_INTEGRATION_RPC_URL: ${{ secrets.SEPOLIA_INTEGRATION_RPC_URL }}
       SEPOLIA_TESTNET_RPC_URL: ${{ secrets.SEPOLIA_TESTNET_RPC_URL }}
```
THenry14, line 206: Let's make sure we use our new rpc urls, if I can see correctly at the moment we are not (they were updated some time ago, whereas new nodes were made yesterday)

.github/workflows/checks.yml
```diff
@@ -202,14 +202,12 @@ jobs:
...
-      GOERLI_INTEGRATION_ACCOUNT_PRIVATE_KEY: ${{ secrets.INTEGRATION_ACCOUNT_PRIVATE_KEY }}
-      GOERLI_INTEGRATION_ACCOUNT_ADDRESS: ${{ secrets.INTEGRATION_ACCOUNT_ADDRESS }}
-      GOERLI_TESTNET_ACCOUNT_PRIVATE_KEY: ${{ secrets.TESTNET_ACCOUNT_PRIVATE_KEY }}
-      GOERLI_TESTNET_ACCOUNT_ADDRESS: ${{ secrets.TESTNET_ACCOUNT_ADDRESS }}
+      SEPOLIA_TESTNET_ACCOUNT_ADDRESS: ${{ secrets.SEPOLIA_TESTNET_ACCOUNT_ADDRESS }}
+      SEPOLIA_TESTNET_ACCOUNT_PRIVATE_KEY: ${{ secrets.SEPOLIA_TESTNET_ACCOUNT_PRIVATE_KEY }}
+      SEPOLIA_INTEGRATION_ACCOUNT_PRIVATE_KEY: ${{ secrets.SEPOLIA_INTEGRATION_ACCOUNT_PRIVATE_KEY }}
+      SEPOLIA_INTEGRATION_ACCOUNT_ADDRESS: ${{ secrets.SEPOLIA_INTEGRATION_ACCOUNT_ADDRESS }}
```
THenry14, line 210: This accounts were previously created by us, right?

starknet_py/hash/transaction_test.py
```diff
@@ -46,9 +46,9 @@ def get_resource_bounds():
                 0x64,
                 [],
                 0x0,
-                StarknetChainId.GOERLI,
+                StarknetChainId.MAINNET,
```
THenry14, line 49: let's create an issue to switch to sepolia when it is unblocked by devnet

starknet_py/net/networks.py
```diff
@@ -3,19 +3,16 @@
 from starknet_py.constants import FEE_CONTRACT_ADDRESS
 
 MAINNET = "mainnet"
-GOERLI = "goerli"
 SEPOLIA_TESTNET = "sepolia_testnet"
```
THenry14, line 6: I know this was already here, but I think this should not be `sepolia_testnet` but rather `sepolia`

starknet_py/tests/e2e/fixtures/accounts.py
```diff
@@ -199,7 +205,7 @@ def pre_deployed_account_with_validate_deploy(
         address=address,
         client=FullNodeClient(node_url=network + "/rpc"),
         key_pair=KeyPair.from_private_key(int(private_key, 16)),
-        chain=StarknetChainId.GOERLI,
+        chain=StarknetChainId.MAINNET,
```
THenry14, line 208: is that right? maybe I misunderstand something, but this may be sepolia in case of network tests, right? so this might fail?

starknet_py/tests/e2e/tests_on_networks/client_test.py
THenry14: this is a regular string, you should be able to just convert its representation to get a proper human-readable message

.github/workflows/checks.yml
tkumor3: Yes, it's taken from a pinned post in our slack

starknet_py/tests/e2e/fixtures/accounts.py
tkumor3: I think we use this function only with devnet, maybe simplify it. Another solution, after the merge https://github.com/software-mansion/starknet.py/pull/1327 can use the option without setting chain id so it will be taken from `get_chan_id`

ddoktorski (changes requested):

starknet_py/tests/e2e/fixtures/accounts.py
tkumor3: I think there are more places with similar logic. Probably all things related to `--net` parameter could be simplified. I create an issue for it. WDYT?

THenry14: let's briefly look through the code and try to assess it first. My suspicion is that there won't be that many places tbh

tkumor3: I remove it and tests works https://github.com/software-mansion/starknet.py/pull/1328/commits/c306b0077e67a9e47a9b6219dbbdfe7d415cc5f6

docs/development.rst
```diff
@@ -33,21 +33,12 @@ If you choose to install `starknet-devnet-rs <https://github.com/0xSpaceShard/st
 
 In order to be able to run tests on testnet and integration networks (``starknet_py/tests/e2e/tests_on_networks/``), you must set some environmental variables:
 
-    - ``GOERLI_INTEGRATION_RPC_URL``
-    - ``GOERLI_TESTNET_RPC_URL``
     - ``SEPOLIA_INTEGRATION_RPC_URL``
-    - ``SEPOLIA_TESTNET_RPC_URL``
```
THenry14, line 39: SEPOLIA_RPC_URL is missing

docs/migration_guide.rst
```diff
@@ -1,6 +1,17 @@
...
+******************************
+0.22.0 Migration guide
+******************************
+
+Version 0.22.0 of **starknet.py**
+
+0.22.0 Breaking changes
+-----------------------
+
+1. Support for Goerli has been removed
```
THenry14, line 13: \+ SEPOLIA_TESTNET -> SEPOLIA , this is also a breaking change

ddoktorski, line 13: Let's also include in the breaking changes that `StarknetChainId.SEPOLIA_TESTNET` has been renamed to `StarknetChainId.SEPOLIA`.

starknet_py/tests/e2e/fixtures/accounts.py
THenry14: ok, so after this was removed there is no way to run this test on network. should we have this test for sepolia as well, or was it something that is non-essential/tested in some other way?
@ddoktorski also keen to hear what you have to say about this

applicable to all the places that changes this

starknet_py/tests/e2e/fixtures/devnet.py
```diff
@@ -35,11 +35,13 @@ def get_start_devnet_command(devnet_port: int) -> List[str]:
...
         "--state-archive-capacity",
         "full",
+        "--chain-id",
+        "MAINNET",
     ]
 
 
 @pytest.fixture(scope="package")
-def run_devnet() -> Generator[str, None, None]:
+def network() -> Generator[str, None, None]:
```
THenry14, line 44: name of this function is now really confusing

docs/development.rst
tkumor3: no, it's here https://github.com/software-mansion/starknet.py/pull/1328/files#diff-9b171ca00cb885ccd9c997cb6f65fbb26d7eae92bf54bd7581b83c4541c156f2R37

THenry14: ahh shiet, I misinterpreted the diff, my bad sorry ~

starknet_py/tests/e2e/fixtures/devnet.py
THenry14: previous name of `run_devnet` seems reasonable

starknet_py/tests/e2e/fixtures/accounts.py
```diff
@@ -34,12 +30,13 @@
...
     account: BaseAccount,
     class_hash: int,
-    network: str,
+    run_devnet,
 ) -> Tuple[str, str]:
     """
     Deploys an Account and adds fee tokens to its balance (only on devnet).
     """
 
+    network = run_devnet
```
tkumor3, line 39: I'm not convinced of this solution, any better idea?

THenry14: I'm not sure I understand this change tbh

tkumor3: I removed [fixture network]( https://github.com/software-mansion/starknet.py/pull/1328/files#diff-d260393b7dd1835a3d244b5250f275d5a1802a4695b30c3e195426a296ac4394L52) which was used in a few places. Currently, we need only the devnet URL which can be obtained by calling `run_devnet` but IMO it's a little bit confusing:
```
@pytest.fixture(name="client", scope="package")
def create_full_node_client(run_devnet) -> FullNodeClient:
    return FullNodeClient(node_url=run_devnet + "/rpc")
```

THenry14: what is `run_devnet`s type? is this a function or a string? can we just pass the url, if this is all we need?

tkumor3: It's Generator[str, None, None], so if you call run_devnet you will get a string 

```
@pytest.fixture(scope="package")
def run_devnet() -> Generator[str, None, None]:
    """
    Runs devnet instance once per module and returns it's address.
    """
    devnet_port, proc = start_devnet()
    yield f"http://localhost:{devnet_port}"
    proc.kill()
```

ddoktorski: For me it can be either `run_devnet_and_get_url` or simply `devnet`. In the second option, I would pass `devnet` directly without the assignment `network = devnet`.

THenry14: let's do that then

ddoktorski (approved):

PR #1340, opened 2024-04-23
Update test_creating_account.py

ldbfpiaoran:
fix  https://starknetpy.readthedocs.io/en/latest/quickstart.html StarkCurveSigner import does not show up bug


ddoktorski (changes requested):
Hi @ldbfpiaoran
Thank you for your contribution! Please run `poetry run poe precommit` locally and then update the branch. This will fix the failing check. Also, please change the base branch from `master` to `development`.

Issue #1353, opened 2024-05-14
Update `TypedData` in line with `SNIP-12`


NikitaMishin:
### Feature Request

Do you guys plan to support snip-12

I see that currently only revision 0 supported https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md
https://github.com/software-mansion/starknet.py/blob/345b0ba9a5b717b6635783e9304784d586cd6d0e/starknet_py/net/models/typed_data.py#L17

ddoktorski:
Hi @NikitaMishin
Thank you for opening the issue. Could you please let me know if you need this implementation for a project you're currently working on?

NikitaMishin:
yes but not critical since we can use revision0 but for better user experience eventually good to have support of revision1

> Hi @NikitaMishin Thank you for opening the issue. Could you please let me know if you need this implementation for a project you're currently working on?

ddoktorski:
@NikitaMishin 

I can confirm that we are going to work towards supporting `SNIP-12`. Unfortunately, I am unable to provide a timeline at the moment.

PR #1360, opened 2024-06-14
Allow `TypedData` revision 1

franciszekjob:
Relates #1353 

## Introduced changes

Basic support for revision 1 `TypedData`, featured in SNIP-12. Full support for SNIP-12 and both revision 0 and 1 basic/preset types and types verification will be added in separate PRs.

- Add `Revision`
- Rename `StarkNetDomain` to `Domain`
- Add `Domain.revision` field; Allow revision 1 to be an integer, ref: https://github.com/starknet-io/SNIPs/pull/79
- Add `HashMethod`

##

- `StarkNetDomain` is renamed to `Domain`
- `TypedData.domain` is now an instance of `Domain` dataclass instead of a dict


starknet_py/net/schemas/common.py
```diff
@@ -336,3 +336,17 @@ class StorageEntrySchema(Schema):
...
+        if isinstance(value, int):
+            value = str(value)
+        return value
+
+
+class RevisionField(fields.Field):
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, str):
+            value = int(value)
+        return value
```
franciszekjob, line 352: Revision can be both string and int, ref https://github.com/starknet-io/SNIPs/pull/79

starknet_py/net/models/typed_data.py
```diff
@@ -14,14 +15,24 @@ class Parameter(TypedDict):
...
+
+class Domain(TypedDict):
+    """
+    TypedDict representing a domain object (both StarkNetDomain, StarknetDomain)
     """
 
     name: str
     version: str
     chainId: Union[str, int]
+    revision: Optional[Union[str, int]]
```
franciszekjob, line 35: We could also use `Optional[Revision]`, don't have a strong opinion here, wdyt @ddoktorski ?

starknet_py/utils/typed_data.py
```diff
@@ -28,9 +107,17 @@ class TypedData:
...
     message: dict
 
+    def __post_init__(self):
+        self._verify_types()
+
+    def _hash_method(self) -> HashMethod:
+        if self.domain.resolved_revision is Revision.V0:
+            return HashMethod.PEDERSEN
+        return HashMethod.POSEIDON
+
```
franciszekjob, line 120: I'm not a fan that we each time when we want to retrieve hash method, we need to call `_hash_method()`. However, having `hash_method` as a field, would require from use to remove `frozen=True` from `TypedData`. Considering this fact, imo current solution is better.

starknet_py/utils/typed_data.py
```diff
@@ -20,6 +23,82 @@ class Parameter:
...
+        Create Domain dictionary from dataclass.
+
+        :return: Domain dictionary.
+        """
+        domain_dict = asdict(self)
+        # Rename chain_id to chainId when converting to dictionary
+        domain_dict["chainId"] = domain_dict.pop("chain_id")
+        return domain_dict
+
+
```
franciszekjob, line 101: `to_dict()` has been added because when converting `Domain` instance to dict, a field `chain_id` should serialize to `chainId`. However using `asdict()` doesn't give such a result. Maybe there's any other way, without creating an extra method?

starknet_py/utils/typed_data.py
```diff
@@ -20,6 +23,50 @@ class Parameter:
...
+    revision: Optional[Union[str, int]] = None
+
+    def __post_init__(self):
+        self.resolved_revision = (
+            Revision(self.revision) if self.revision else Revision.V0
+        )
+        self.separator_name = self._resolve_separator_name()
+
+    def _resolve_separator_name(self):
+        if self.resolved_revision is Revision.V0:
```
ddoktorski, line 44: Generally it is recommended to use `==`, is there any specific reason to use `is` here?
```suggestion
        if self.resolved_revision == Revision.V0:
```

starknet_py/utils/typed_data.py
ddoktorski: `asdict` is a function from `dataclasses` module and it doesn't perform a conversion using a marshmallow schema here. Let's try something similar to this:
```
    return cast(Dict, DomainSchema().dump(obj=self))
```

starknet_py/utils/typed_data.py
ddoktorski: I think that's fine. We can also add a `@property` decorator and then it can be accessed without parentheses.

ddoktorski (changes requested):
Please also update the [migration guide ](https://starknetpy.readthedocs.io/en/latest/migration_guide.html) with all the breaking changes. It can done in a separate PR.

starknet_py/utils/typed_data.py
```diff
@@ -20,6 +23,84 @@ class Parameter:
...
+        "shortstring",
+    ]
+
+    @staticmethod
+    def get_by_revision(revision: Revision) -> List[str]:
+        return BasicType.V0 if revision is Revision.V0 else BasicType.V1
+
+
+class PresetType:
+    V1 = ["u256", "TokenAmount", "NftId"]
```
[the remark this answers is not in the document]
ddoktorski: It depends on how we are going to use it, but if it's going to be added in the next PR maybe let's discuss it there.

starknet_py/net/schemas/common.py
franciszekjob: Yes, to be more precise - type of revision is an integer for revision 1 and a string for revision 0. Please look at [the test for revision as string](https://github.com/software-mansion/starknet-jvm/blob/7ee0e50c5291e61920ca347ad299828fcf15524a/lib/src/test/kotlin/starknet/data/TypedDataTest.kt#L806) in JVM sdk.

starknet_py/utils/typed_data.py
franciszekjob: Yes, let's use `==` as it compares the actual value 👍

starknet_py/net/schemas/common.py
ddoktorski: Ok it makes sense 👍 Aside from that we want to return a `Revision` enum in `_deserialize`, right? Furthermore, we can verify if the value is either `0` or `1`, otherwise raise an error.

starknet_py/utils/typed_data_test.py
```diff
@@ -73,29 +89,111 @@ def test_type_hash(example, type_name, type_hash):
...
+    "included_type",
+    [
+        "felt",
+        "felt*",
+        "string",
+        "selector"
+    ],
+)
+def test_invalid_types(included_type):
+    with pytest.raises(ValueError):
```
ddoktorski, line 198: nit:
```suggestion
    with pytest.raises(ValueError, match=f"Reserved type name: {included_type}"):
```

starknet_py/net/schemas/common.py
franciszekjob: Sure, I've added `Revision._missing_()` method with custom error message, informing about possible revision values.

starknet_py/utils/typed_data.py
```diff
@@ -19,6 +21,47 @@ class Parameter:
...
+    def from_dict(data: DomainDict) -> "Domain":
+        """
+        Create Domain dataclass from dictionary.
+
+        :param data: Domain dictionary.
+        :return: Domain dataclass instance.
+        """
+        return cast(Domain, DomainSchema().load(data))
+
+    def to_dict(self) -> dict:
```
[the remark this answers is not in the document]
franciszekjob: I'va added it to `TypedData` and `Parameter` 👍

starknet_py/net/schemas/common.py
franciszekjob: Update: after online discussion, `Revision._missing_()` has been removed. Checking invalid revision value is performed in `RevisionField._deserialize()`.

starknet_py/net/schemas/common.py
```diff
@@ -336,3 +336,17 @@ class StorageEntrySchema(Schema):
     def make_dataclass(self, data, **kwargs):
         # pylint: disable=no-self-use
         return StorageEntry(**data)
+
+
+class ChainIdField(fields.Field):
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, int):
```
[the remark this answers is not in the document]
franciszekjob: Update: after online discussion we decided to accept `chainId` only as `str`. It could be extended to accept `int` in future, if there is such a need.

starknet_py/utils/typed_data.py
```diff
@@ -40,6 +100,24 @@ def from_dict(data: TypedDataDict) -> "TypedData":
         """
         return cast(TypedData, TypedDataSchema().load(data))
 
+    def to_dict(self) -> dict:
```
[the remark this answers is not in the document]
franciszekjob: Yes, the reason it was done this way is because initially `TypedData`'s nested fields weren't dumped to dict correctly. However, the change from
```domain = fields.Dict(data_key="domain", required=True)```
to 
```domain = fields.Nested(DomainSchema, required=True)```
lets use rely on `dump` method 👍

File: starknet_py/utils/typed_data.py [hunk 11]
```diff
@@ -18,6 +20,55 @@ class Parameter:
     name: str
     type: str
 
+    def to_dict(self) -> dict:
```
ddoktorski, line 23: Do we need this method for the `Parameter` class?

starknet_py/net/schemas/common.py
```diff
@@ -336,3 +337,27 @@ class StorageEntrySchema(Schema):
...
+class Revision(Enum):
+    """
+    Enum representing the revision of the specification to be used.
+    """
+
+    V0 = 0
+    V1 = 1
+
+
+class RevisionField(fields.Field):
```
ddoktorski, line 351: We need to implement the `_serialize_` method here as well, to ensure that the `dump` method works properly, for example with `TypedDataSchema`.

starknet_py/utils/typed_data.py
franciszekjob: No, it's not used anywhere, removed

starknet_py/net/schemas/common.py
franciszekjob: Added 🫡

starknet_py/net/schemas/common.py
```diff
@@ -336,3 +337,34 @@ class StorageEntrySchema(Schema):
...
+    def _serialize(self, value: Any, attr: str, obj: Any, **kwargs):
+        if value is None:
+            return str(Revision.V0.value)
+        if value == Revision.V0:
+            return str(value.value)
+        return int(value.value)
```
ddoktorski, line 357: nit: Casting is not required here
```suggestion
        return value.value
```

starknet_py/net/schemas/common.py
```diff
@@ -336,3 +337,34 @@ class StorageEntrySchema(Schema):
...
+    V0 = 0
+    V1 = 1
+
+
+class RevisionField(fields.Field):
+    def _serialize(self, value: Any, attr: str, obj: Any, **kwargs):
+        if value is None:
+            return str(Revision.V0.value)
+        if value == Revision.V0:
+            return str(value.value)
```
ddoktorski, line 356: nit:
```suggestion
        if value is None or value == Revision.V0:
            return str(Revision.V0.value)
```

ddoktorski (approved):

PR #1363, opened 2024-06-20
Add `MerkleTree` and merkletree support in `TypedData`

franciszekjob:
Relates #1353 

## Introduced changes

- Add `HashMethod.hash_many()`
- Add `Parameter.contains` field to support `merkletree` type
- Add `MerkleTree` dataclass
- Add `BasicType` enum with `felt`, `selector`, `merkletree` and `shortstring`
- Add merkle tree tests; Extend typed data tests with merkle tree cases

##


starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+from starknet_py.net.client_utils import _to_rpc_felt
+
+
+@dataclass
+class MerkleTree:
+    """
+    Dataclass representing a MerkleTree object.
+    """
+
+    leaves: List[str]
```
ddoktorski, line 14: Why do we store the leaves as `str` instead of `int`?

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+
+        while len(leaves) > 1:
+            if len(leaves) != len(leaf_hashes):
+                branches.append(leaves[:])
+
+            new_leaves = []
+            for i in range(0, len(leaves), 2):
+                a, b = leaves[i], leaves[i + 1] if i + 1 < len(leaves) else "0"
+                a, b = int(a, 16), int(b, 16)
+                new_leaves.append(_to_rpc_felt(MerkleTree.hash(a, b, hash_method)))
```
ddoktorski, line 43: Do we need the static function `MerkleTree.hash` at all?
```suggestion
                new_nodes.append(_to_rpc_felt(hash_method.hash(*sorted([a, b]))))
```

ddoktorski (changes requested):

starknet_py/utils/merkle_tree.py
franciszekjob: I wanted to store them as `str` representing felt values. However, considering that hashing methods use `int` it may be more suitable to rely on mentioned `int`. Also, it reduces extra conversions. Changes applied 👍

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+            )
+
+    @staticmethod
+    def build(
+        leaf_hashes: List[str], hash_method: HashMethod
+    ) -> Tuple[str, List[List[str]]]:
+        if not leaf_hashes:
+            raise ValueError("Cannot build Merkle tree from an empty list of leaves.")
+
+        leaves = leaf_hashes[:]
```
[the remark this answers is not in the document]
franciszekjob: The suggestions are mostly valid, I just wonder about the inconsistency (in variables naming) with existing implementations in other SDKs (JVM, Swift, JS). Wdyt @ddoktorski ?

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+        branches: List[List[str]] = []
+
+        while len(leaves) > 1:
+            if len(leaves) != len(leaf_hashes):
+                branches.append(leaves[:])
+
+            new_leaves = []
```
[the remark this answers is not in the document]
franciszekjob: Same as [above](https://github.com/software-mansion/starknet.py/pull/1363/files#r1649305975).

starknet_py/utils/merkle_tree.py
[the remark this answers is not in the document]
ddoktorski: Ideally, we should aim for consistency in names across all SDKs. Although if there's a space for relevant improvements, I wouldn't follow the other SDKs only for the sake of consistency. In this case I think using names like `branches` and `leaves` in this context is just confusing.

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclass
+class MerkleTree:
+    """
+    Dataclass representing a MerkleTree object.
+    """
+
+    leaves: List[str]
+    hash_method: HashMethod
+    root_hash: str = field(init=False)
+    branches: List[List[str]] = field(init=False)
```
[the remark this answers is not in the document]
franciszekjob: Renamed

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __post_init__(self):
+        if self.leaves:
+            self.root_hash, self.branches = MerkleTree.build(
+                self.leaves, self.hash_method
+            )
+
+    @staticmethod
+    def build(
+        leaf_hashes: List[str], hash_method: HashMethod
```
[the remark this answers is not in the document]
franciszekjob: Renamed

starknet_py/utils/merkle_tree.py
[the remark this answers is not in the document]
franciszekjob: Renamed

starknet_py/utils/merkle_tree.py
franciszekjob: 1. I think it's better to keep sorting `a` and `b` inside the method, rather than passing result of `sorted()` as an arg.  Using `hash_method.hash()` assumed that args will be passed on proper order.
2. `hash_method` has `hash()` and `hash_many()` which may pose some uncertainty while using.
3. Other SDKs have same method, so it would be good to keep the consistency.

Wdyt @ddoktorski ?

starknet_py/net/models/typed_data.py
```diff
@@ -36,3 +37,8 @@ class TypedDataDict(TypedDict):
     primaryType: str
     domain: DomainDict
     message: Dict[str, Any]
+
+
+class TypeContext(TypedDict):
+    parent: str
+    key: str
```
franciszekjob, line 44: Ideally we would name it `Context`, but there is already a [class with such name](https://github.com/software-mansion/starknet.py/blob/56c60c1bb0ac0b56dd220c597fffef8745845acb/starknet_py/serialization/_context.py#L15).

franciszekjob:
Add variable naming changes have been applied.

starknet_py/utils/merkle_tree.py
ddoktorski: The main question is: do we really need the static `hash` function in the `MerkleTree`? Is it going to be useful for users?

starknet_py/hash/hash_method.py
```diff
@@ -14,7 +14,14 @@ class HashMethod(Enum):
     PEDERSEN = "pedersen"
     POSEIDON = "poseidon"
 
-    def hash(self, values: List[int]):
+    def hash(self, left: int, right: int):
```
ddoktorski, line 17: Not sure why we need this method. Why can't we simply pass a list with two elements instead?

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @staticmethod
+    def build(
+        leaves: List[int], hash_method: HashMethod
+    ) -> Tuple[int, List[List[int]]]:
+        if not leaves:
+            raise ValueError("Cannot build Merkle tree from an empty list of leaves.")
+
+        curr_level_nodes = leaves[:]
+        branches: List[List[int]] = []
```
ddoktorski, line 32: Should be renamed as well.

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    leaves: List[int]
+    hash_method: HashMethod
+    root_hash: int = field(init=False)
+    levels: List[List[int]] = field(init=False)
+
+    def __post_init__(self):
+        if self.leaves:
+            self.root_hash, self.branches = MerkleTree.build(
```
ddoktorski, line 20:
```suggestion
            self.root_hash, self.levels = MerkleTree.build(
```

starknet_py/utils/merkle_tree_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+import pytest
+from poseidon_py.poseidon_hash import poseidon_hash
+
+from starknet_py.hash.hash_method import HashMethod
+from starknet_py.hash.utils import pedersen_hash
+from starknet_py.utils.merkle_tree import MerkleTree
+
+
+@pytest.mark.parametrize(
```
ddoktorski, line 11: Just to confirm - have the examples below been taken from other SDKs?

starknet_py/utils/merkle_tree.py
```diff
@@ -0,0 +1,52 @@
...
+            for i in range(0, len(curr_level_nodes), 2):
+                a, b = (
+                    curr_level_nodes[i],
+                    curr_level_nodes[i + 1] if i + 1 < len(curr_level_nodes) else 0,
+                )
+                new_nodes.append(MerkleTree.hash(a, b, hash_method))
+
+            curr_level_nodes = new_nodes
+
+        return curr_level_nodes[0], branches
```
ddoktorski, line 48: Why don't we store leaves and root hash in the `levels`?

starknet_py/utils/merkle_tree_test.py
```diff
@@ -0,0 +1,142 @@
...
+):
+    tree = MerkleTree(leaves, hash_method)
+    assert tree.root_hash is not None
+    assert tree.branches is not None
+    assert tree.root_hash == expected_root
+    assert len(tree.branches) == expected_branch_count
+
+
+@pytest.mark.parametrize("hash_method", list(HashMethod))
+def test_build_from_1_element(hash_method: HashMethod):
```
ddoktorski, line 85: Please refactor these tests to have only one parameterized test, which takes as an input leaves, hash method and precalculated root hash.

starknet_py/net/models/typed_data.py
ddoktorski: The current name is fine 👍

ddoktorski (changes requested):

starknet_py/utils/merkle_tree_test.py
franciszekjob: Yup, they've been taken from [Swift SDK](https://github.com/software-mansion/starknet.swift/pull/167/files#diff-38e25e09052ab3aafadd4ca0ab04af05e2737084545cf2a61c2bd1075dacaf78).

starknet_py/utils/typed_data_test.py
```diff
@@ -134,18 +152,18 @@ def test_message_hash(example, account_address, msg_hash):
...
-        {"name": "name", "type": "shortstring"},
-        {"name": "version", "type": "shortstring"},
-        {"name": "chainId", "type": "shortstring"},
-        {"name": "revision", "type": "shortstring"},
+        Parameter(name="name", type="shortstring"),
+        Parameter(name="version", type="shortstring"),
+        Parameter(name="chainId", type="shortstring"),
+        Parameter(name="revision", type="shortstring"),
     ]
 }
```
franciszekjob, line 168: Update list of `dict` to list of `Parameter`, according to `TypedData.types` type [ref](https://github.com/software-mansion/starknet.py/blob/a59c8a278c986002efda9e0c64afe2d46086f3fe/starknet_py/utils/typed_data.py#L85).

starknet_py/utils/typed_data_test.py
```diff
@@ -162,13 +180,6 @@ def test_message_hash(example, account_address, msg_hash):
     revision=Revision.V1,
 )
 
-domain_object_v1 = {
-    "name": "DomainV1",
-    "version": "1",
-    "chainId": "1234",
-    "revision": 1
-}
-
```
franciszekjob, line 171: Deleted, as it wasn't used anywhere.

starknet_py/utils/merkle_tree.py
franciszekjob: I've added it in these two commits
https://github.com/software-mansion/starknet.py/pull/1363/commits/c490faf860c77bd7f060f087987076c13e37eeb6
https://github.com/software-mansion/starknet.py/pull/1363/commits/faee6d4b97fa6e6772978090bb683611f2c84636
Now we return leaves, levels and the root hash.

starknet_py/hash/hash_method.py
franciszekjob: Imho it's better to keep `hash()` and `hash_many()` as separate methods. Generally hash methods for two elements produce different output than for many elements. For instance, such code
```python
print(poseidon_hash(1, 2))
print(poseidon_hash_many([1, 2]))
```
gives following output
```python
2636648219362971850283425434366427370362725365790740855428580782178634926362
1557996165160500454210437319447297236715335099509187222888255133199463084263
```

starknet_py/utils/merkle_tree.py
franciszekjob: Offline discussion: we decided to remove `MerkleTree.hash()` as it didn't simplify much and shouldn't be used by users.

starknet_py/utils/typed_data.py
```diff
@@ -204,14 +239,44 @@ def message_hash(self, account_address: int) -> int:
...
+
+    def _get_merkle_tree_leaves_type(self, context: TypeContext) -> str:
+        parent, key = context.parent, context.key
+
+        if parent not in self.types:
+            raise ValueError(f"Parent {parent} is not defined in types.")
+
+        parent_type = self.types[parent]
+
+        # target_type = next((item for item in parent_type if item.type == key), None)
```
ddoktorski, line 260: Let's uncomment it and remove the code below.

starknet_py/utils/typed_data.py
```diff
@@ -235,13 +295,28 @@ def escape(s: str, revision: Revision) -> str:
...
+
+
 # pylint: disable=unused-argument
 # pylint: disable=no-self-use
 
 
 class ParameterSchema(Schema):
     name = fields.String(data_key="name", required=True)
     type = fields.String(data_key="type", required=True)
+    contains = fields.String(data_key="contains", required=False)
```
ddoktorski, line 319: Should it be required?

starknet_py/utils/typed_data.py
```diff
@@ -204,14 +239,39 @@ def message_hash(self, account_address: int) -> int:
...
+
+        parent_type = self.types[parent]
+
+        target_type = next((item for item in parent_type if item.name == key), None)
+        if target_type is None:
+            raise ValueError(
+                f"Key {key} is not defined in type {parent} or multiple definitions are present."
+            )
+
+        if not target_type.contains:
```
ddoktorski, line 266:
```suggestion
        if target_type.contains is None:
```

starknet_py/utils/typed_data.py
```diff
@@ -204,14 +239,39 @@ def message_hash(self, account_address: int) -> int:
...
         ]
 
-        return self._hash_method.hash(message)
+        return self._hash_method.hash_many(message)
+
+    def _prepare_merkle_tree_root(self, value: List, context: TypeContext) -> int:
+        merkle_tree_type = self._get_merkle_tree_leaves_type(context)
+        struct_hashes = list(
+            map(lambda struct: self._encode_value(merkle_tree_type, struct), value)
+        )
```
ddoktorski, line 248: nit:
```suggestion
        struct_hashes = [self._encode_value(merkle_tree_type, struct) for struct in value]
```

starknet_py/utils/typed_data.py
franciszekjob: Could you please clarify your question (because `contains` is actually not required)? `contains` appears in merkle tree parameter, in standard one it's not present - thats's why it's optional.

ddoktorski: Never mind, for some reason I saw `True` there, all good here.

ddoktorski (approved):
Please fix the lint only and then it will be good to go 🚀

PR #1371, opened 2024-06-26
feat: from_keystore method

### LINKED ISSUE #1368: dev: support keystore
Right now, afaik it's not possible to create a signer using a keystore path that would be great to improve security of backend services using starknet.py.

Happy to provide a PR if necessary.

## References
https://github.com/xJonathanLEI/starknet-rs/blob/016f3ee3293c8355f8a7cb64777adea6119493b9/starknet-signers/src/key_pair.rs#L60

EvolveArt:
Closes #1368 

## Introduced changes

- Adds https://pypi.org/project/eth-keyfile/ as a dependency
- Adds a `from_keystore` method to the `KeyPair` class

##


poetry.lock
ddoktorski, line 1: Could you please update the dependencies using `poetry update eth-keyfile`? This way we can avoid updating dependencies that are not relevant for this change.

pyproject.toml
ddoktorski, line 1: What's the reason to format the `pyproject.toml` file?

starknet_py/net/signer/stark_curve_signer.py
```diff
@@ -32,6 +34,18 @@ def from_private_key(key: Hash) -> "KeyPair":
             key = int(key, 0)
         return KeyPair(private_key=key, public_key=private_to_stark_key(key))
 
+    @staticmethod
+    def from_keystore(path: str, password: str) -> "KeyPair":
+        """
+        Create a key pair from a keystore file.
+        The keystore file should follow the Ethereum keystore format.
+        :param path: Path to the keystore file.
```
ddoktorski, line 42: A newline is necessary for correct formatting in the docs.
```suggestion
        The keystore file should follow the Ethereum keystore format.
        
        :param path: Path to the keystore file.
```

ddoktorski (changes requested):
Thank you for your contribution! I've left a few minor comments that need to be addressed before merging :)

pyproject.toml
EvolveArt: just ran `poe format` in the repo

EvolveArt: reverted

poetry.lock
EvolveArt: done

ddoktorski (approved):
Thank you! LGTM 🔥 

_Note: Tests on network fail due to unset secrets in the fork repository_

PR #1372, opened 2024-06-27
Support `u128`, `i128`, `timestamp` in `TypedData`

franciszekjob:
Relates #1353 

## Introduced changes

Support `u128`, `i128`, `timestamp` basic types
- Add `encode_u128()`, `encode_i128()`
- Add `u128` and `i128` encoding tests

##

### DIFF
--- a/starknet_py/utils/typed_data.py
+++ b/starknet_py/utils/typed_data.py
@@ -116,7 +116,7 @@ def _encode_value_v1(
 
 def _encode_value_v0(
     basic_type: BasicType,
-    value: Union[int, str, dict, list],
+    value: Union[int, str],
 ) -> Optional[int]:
     if basic_type in (
         BasicType.FELT,
--- a/starknet_py/utils/typed_data.py
+++ b/starknet_py/utils/typed_data.py
@@ -91,6 +91,42 @@ class BasicType(Enum):
     TIMESTAMP = "timestamp"
 
 
+def _encode_value_v1(
+    basic_type: BasicType, value: Union[int, str, dict, list]
+) -> Optional[int]:
+    if basic_type in (
+        BasicType.FELT,
+        BasicType.SHORT_STRING,
+        BasicType.CONTRACT_ADDRESS,
+        BasicType.CLASS_HASH,
+    ) and isinstance(value, (int, str)):
+        return parse_felt(value)
+
+    if basic_type in (
+        BasicType.U128,
+        BasicType.TIMESTAMP,
+    ) and isinstance(value, (int, str)):
+        return encode_u128(value)
+
+    if basic_type == BasicType.I128 and isinstance(value, (int, str)):
+        return encode_i128(value)
+
+    return None
+
+
+def _encode_value_v0(
+    basic_type: BasicType,
+    value: Union[int, str, dict, list],
+) -> Optional[int]:
+    if basic_type in (
+        BasicType.FELT,
+        BasicType.STRING,
+    ) and isinstance(value, (int, str)):
+        return parse_felt(value)
+
+    return None
+
+
 @dataclass(frozen=True)
 class TypedData:
     """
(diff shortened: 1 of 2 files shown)


starknet_py/utils/typed_data.py
```diff
@@ -360,6 +377,62 @@ def encode_bool(value: Union[bool, str, int]) -> int:
...
+
+        if value.startswith("0x"):
+            int_value = int(value, 16)
+        elif is_digit_string(value):
+            int_value = int(value)
+
+        if int_value is not None and is_in_range(int_value):
+            return int_value
+
+    raise ValueError(f"Value [{value}] is out of range for '{BasicType.U128}'.")
```
ddoktorski, line 406: This is the sole error we return. Therefore, any incorrect string value not starting with `0x` will result in `out of range` error. An example of this would be the `example` case in `test_encode_invalid_u128`. A better approach would be to handle the invalid number error separately. The same for `encode_i128`.

starknet_py/utils/typed_data.py
```diff
@@ -158,6 +163,18 @@ def _encode_value(
         ] and isinstance(value, (int, str)):
             return parse_felt(value)
 
+        if (basic_type, self.domain.resolved_revision) in [
```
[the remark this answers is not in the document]
franciszekjob: Sure, it makes sense for me. I've added these functions 👍

starknet_py/utils/typed_data.py
```diff
@@ -128,7 +169,6 @@ def to_dict(self) -> dict:
     def _is_struct(self, type_name: str) -> bool:
         return type_name in self.types
 
-    # pylint: disable=too-many-return-statements
```
franciszekjob, line 131: This directive is not needed anymore.

starknet_py/utils/typed_data.py
```diff
@@ -360,6 +398,66 @@ def encode_bool(value: Union[bool, str, int]) -> int:
...
+
+        if value.startswith("0x"):
+            int_value = int(value, 16)
+        elif is_digit_string(value):
+            int_value = int(value)
+        else:
+            raise ValueError(f"Value [{value}] is not a valid number.")
+
+        if int_value is not None and is_in_range(int_value):
+            return int_value
```
ddoktorski: Maybe something like this? The point is to avoid duplicating the same error twice and to throw `not valid number` error for `None` as well. We should apply similar modifications to `i128`.
```suggestion
    if isinstance(value, str) and value.startswith("0x"):
        int_value = int(value, 16)
    elif isinstance(value, str) and is_digit_string(value):
        int_value = int(value)
    elif isinstance(value, int):
        int_value = value
    else:
        raise ValueError(f"Value [{value}] is not a valid number.")

    if is_in_range(int_value):
        return int_value
```

starknet_py/utils/typed_data.py
```diff
@@ -360,6 +398,54 @@ def encode_bool(value: Union[bool, str, int]) -> int:
...
+    elif isinstance(value, int):
+        int_value = value
+    else:
+        raise ValueError(f"Value [{value}] is not a valid number.")
+
+    if abs(int_value) >= FIELD_PRIME:
+        raise ValueError(
+            f"Values outside the range (-FIELD_PRIME, FIELD_PRIME) are not allowed, [{value}] given."
+        )
+    int_value %= FIELD_PRIME
```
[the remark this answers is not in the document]
franciszekjob: In case of **signed** number, integer value need to be in range `(-P, P)`, because of they way how negative numbers are calculated. Here's an official [docs reference](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/#data_types_greater_than_252_bits).

Also, as other reference, here's a [similar implementation](https://github.com/software-mansion/starknet-jvm/blob/main/lib/src/main/kotlin/com/swmansion/starknet/data/types/Felt.kt#L117-L122) in JVM SDK.

ddoktorski (approved):

PR #1373, opened 2024-06-27
Support `string` revision 1 in `TypedData`

franciszekjob:
Relates #1353 

## Introduced changes

- Support `string` revision 1 (long string)
- Add typed data with basic types example as tests case

##


starknet_py/utils/typed_data.py
```diff
@@ -103,6 +104,8 @@ class TypedData:
     domain: Domain
     message: dict
 
+    _byte_array_serializer: ByteArraySerializer = ByteArraySerializer()
```
franciszekjob, line 107: I decided to use already existing serializer for byte arrays.

starknet_py/utils/typed_data.py
```diff
@@ -105,6 +106,7 @@ class TypedData:
 
     def __post_init__(self):
         self._verify_types()
+        self._byte_array_serializer = ByteArraySerializer()
```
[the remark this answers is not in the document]
ddoktorski: In our case instantiating `ByteArraySerializer` in `_prepare_long_string` each time should be fine. Alternatively, as we discussed offline we can create a singleton class that contains `ByteArraySerializer`.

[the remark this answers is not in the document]
franciszekjob: Ok, let's stick to the first option.

starknet_py/utils/typed_data.py
```diff
@@ -169,6 +134,45 @@ def to_dict(self) -> dict:
...
+        if basic_type == BasicType.I128 and isinstance(value, (int, str)):
+            return encode_i128(value)
+
+        if basic_type == BasicType.STRING and isinstance(value, str):
+            return self._prepare_long_string(value)
+
+        return None
+
+    # pylint: disable=no-self-use
+    def _encode_value_v0(
```
franciszekjob, line 163: `_encode_value_v0()` could be a function, however having it and `_encode_value_v1()` as respectively function and `TypedData` method doesn't look good. Let's keep both as methods.

starknet_py/utils/typed_data.py
```diff
@@ -350,6 +354,11 @@ def _get_merkle_tree_leaves_type(self, context: TypeContext) -> str:
 
         return target_type.contains
 
+    def _prepare_long_string(self, value: str) -> int:
```
ddoktorski, line 357:
```suggestion
    def _encode_long_string(self, value: str) -> int:
```

ddoktorski (approved):

PR #1382, opened 2024-07-03
Support `enum` basic type in `TypedData`

franciszekjob:
Relates #1353 

## Introduced changes

- Support revision 1  `enum` basic type in `TypedData`
- Make `Parameter` abstract
- Add `StandardParameter`, `EnumParameter`, `MerkleTreeParameter`
- Add enum-related tests
##

- `Parameter` is now abstract; `StandardParameter`, `EnumParameter`, `MerkleTreeParameter` should be used


starknet_py/utils/typed_data.py
```diff
@@ -242,58 +283,114 @@ def _verify_types(self):
...
-            for type_name in self.types
-            for ref_type in self.types[type_name]
-        }
+        referenced_types = set()
+        for type_name in self.types:
+            for ref_type in self.types[type_name]:
+                if isinstance(ref_type, MerkleTreeParameter):
+                    referenced_types.add(ref_type.contains)
+                elif isinstance(ref_type, EnumParameter):
+                    self._validate_enum_type()
```
ddoktorski, line 292: Not sure about the validation here. What about other basic types from revision 1 that are not validated this way? Why is it only for `Enum`?

starknet_py/utils/typed_data.py
```diff
@@ -341,6 +438,14 @@ def _prepare_merkle_tree_root(self, value: List, context: TypeContext) -> int:
         return MerkleTree(struct_hashes, self._hash_method).root_hash
 
     def _get_merkle_tree_leaves_type(self, context: TypeContext) -> str:
+        target_type = self._resolve_type(context)
+
+        if not isinstance(target_type, MerkleTreeParameter):
+            raise ValueError("Missing 'contains' field in target type.")
```
ddoktorski, line 444: Is it correct error message?

ddoktorski (changes requested):

starknet_py/utils/typed_data.py
```diff
@@ -198,9 +238,9 @@ def _encode_value(
         ):
             encoded_value = self._encode_value_v0(basic_type, value)
         elif self.domain.resolved_revision == Revision.V1 and isinstance(
-            value, (str, int)
+            value, (str, int, dict)
```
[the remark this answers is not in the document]
franciszekjob: I don't think it can be omitted, because the range of types that `value` can be in `_encode_value_vX()` is smaller than in `_encode_value()`. Without this check, `value` could be for example of type `list`, which isn't considered in `_encode_value_v1()`.

starknet_py/utils/typed_data.py
```diff
@@ -385,10 +536,8 @@ def strip_pointer(value: str) -> str:
     return value
 
 
-def escape(s: str, revision: Revision) -> str:
-    if revision == Revision.V0:
-        return s
-    return f'"{s}"'
+def is_enum(value: str) -> bool:
```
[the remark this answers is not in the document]
franciszekjob: Yes, according to [enum subsection](https://github.com/starknet-io/SNIPs/pull/77/files#diff-d594c362a3c260ea39dacaca55a6f02927013cebd19a5c8d202117ff9147ff14R124) it checks for enum variant type. Let's change it 👍

starknet_py/utils/typed_data.py
franciszekjob: I've changed it to `Target type is not a merkletree type.` 👍

starknet_py/utils/typed_data.py
franciszekjob: This validation is performed, because a custom enum type **could not** be defined in revision 0. For example, `timestamp` is defined in rev 1, so if user defined it in revision 0, it would be simply treated as a custom type. However, as per [type identification](https://github.com/starknet-io/SNIPs/pull/77/files#diff-d594c362a3c260ea39dacaca55a6f02927013cebd19a5c8d202117ff9147ff14R124), this does not apply to enums.

starknet_py/utils/typed_data_test.py
```diff
@@ -4,16 +4,18 @@
...
 from pathlib import Path
-from typing import Union
+from typing import Dict, List, Union
 
 import pytest
 
 from starknet_py.net.models.typed_data import Revision
 from starknet_py.tests.e2e.fixtures.constants import TYPED_DATA_DIR
-from starknet_py.utils.typed_data import (
+from starknet_py.utils.typed_data import (  # Parameter,
```
ddoktorski, line 13: 👀

starknet_py/utils/typed_data.py
ddoktorski: Thanks, as discussed offline let's leave it as it is.

starknet_py/utils/typed_data.py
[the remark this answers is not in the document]
ddoktorski: I overlooked the fact that this would cause some `pyright` errors, it can stay as it is.

starknet_py/utils/typed_data.py
```diff
@@ -242,58 +272,114 @@ def _verify_types(self):
...
+        referenced_types = set()
+        for type_name in self.types:
+            for ref_type in self.types[type_name]:
+                if isinstance(ref_type, MerkleTreeParameter):
+                    referenced_types.add(ref_type.contains)
+                elif isinstance(ref_type, EnumParameter):
+                    self._validate_enum_type()
+                    referenced_types.add(ref_type.contains)
+                else:
+                    referenced_types.add(strip_pointer(ref_type.type))
```
ddoktorski, line 284: Is it expected that we add enum variant types here like `(u128)`? I think it doesn't break anything, but it should be handled separately.

franciszekjob: Yes, there's actually a [test with such case](https://github.com/software-mansion/starknet.py/pull/1382/files#diff-acfd7a6c6e0206824de06e6c789bf7df469ea1261e4440a685ef2357c55058d6R13).

franciszekjob: Correct, I've added a check for it 👍

starknet_py/utils/typed_data.py
```diff
@@ -354,17 +457,67 @@ def _get_merkle_tree_leaves_type(self, context: TypeContext) -> str:
...
-        return target_type.contains
+    def _encode_enum(self, value: dict, context: TypeContext):
+        if len(value.keys()) != 1:
+            raise ValueError(
+                f"'{BasicType.ENUM.name}' value must contain a single variant."
+            )
+
+        variant_name, variant_data = next(iter(value.items()))
+        variants = self._get_enum_variants(context)
+        variant_type = next(
```
ddoktorski, line 472: nit:
```suggestion
        variant_definition = next(
```
Or something similar, "type" is a bit misleading here.

starknet_py/utils/typed_data.py
```diff
@@ -354,17 +457,67 @@ def _get_merkle_tree_leaves_type(self, context: TypeContext) -> str:
...
+            (item for item in variants if item.name == variant_name), None
+        )
+
+        if variant_type is None:
+            raise ValueError(
+                f"Variant [{variant_name}] is not defined in '${BasicType.ENUM.name}' "
+                f"type [{context.key}] or multiple definitions are present."
+            )
+
+        variant_index = variants.index(variant_type)
```
ddoktorski, line 482: nit: Please move this line right above the return statement.

ddoktorski:
Please also fix the docs build. It's failing because class `Parameter` does not exist.

starknet_py/utils/typed_data.py
```diff
@@ -242,58 +272,121 @@ def _verify_types(self):
...
+                if isinstance(param, EnumParameter):
+                    self._validate_enum_type()
+                    extracted_types = [param.contains]
+                else:
+                    extracted_types = [param.type]
+
+                extracted_types = [
+                    strip_pointer(extr_type) for extr_type in extracted_types
+                ]
+                for extracted_type in extracted_types:
```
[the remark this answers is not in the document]
franciszekjob: In fact we should also check if `param.type` is an enum variant type, like this:
```python
elif is_enum_variant_type(param.type):
    self._validate_enum_type()
    extracted_types = _extract_enum_types(param.type)
```
Updated.

ddoktorski (approved):
Please address the last comment before merging 🚀

Issue #1316, opened 2024-03-08
Split RPC schemas into multiple files


ddoktorski:
- Split schemas from the file `/net/schemas/rpc.py`
- Rename `/net/schemas/gateway.py` and remove unused schemas

PR #1385, opened 2024-07-05
Split RPC schemas into multiple files

kkawula:
Closes [#1316](https://github.com/software-mansion/starknet.py/issues/1316)

## Introduced changes
- rpc.py spilted into 5 files `transctions.py` `trace_api.py` `block.py` `contract.py` `event.py`

##

- Move schemas from `gateway.py` to `rpc.py`
- Split schemas from `rpc.py` to other files in rpc directory
- Rename `CompiledContract`, `ContractClass` and `ContractClassSchema` to `DeprecatedCompiledContract`, `DeprecatedContractClass` and  `DeprecatedContractClassSchema`


starknet_py/net/models/transaction.py
```diff
@@ -29,14 +29,16 @@
...
     DAMode,
+    DeprecatedContractClass,
     ResourceBoundsMapping,
     SierraContractClass,
     TransactionType,
 )
 from starknet_py.net.schemas.common import Felt, TransactionTypeField
-from starknet_py.net.schemas.gateway import (
+
+# from starknet_py.net.schemas.gateway import SierraContractClassSchema
```
ddoktorski, line 40: 👀

starknet_py/net/schemas/rpc/contract.py
```diff
@@ -1,74 +1,68 @@
+# pylint: disable=too-many-lines
```
ddoktorski, line 1: Is it necessary?

starknet_py/net/schemas/rpc/contract.py
```diff
@@ -1,74 +1,68 @@
...
     SierraContractClass,
     SierraEntryPoint,
     SierraEntryPointsByType,
+    SyncStatus,
 )
-from starknet_py.net.schemas.common import Felt
+from starknet_py.net.schemas.common import Felt, NumberAsHex
+from starknet_py.utils.schema import Schema
 
 # pylint: disable=unused-argument, no-self-use
```
ddoktorski, line 25: I think it can be removed now.
```suggestion
```

starknet_py/net/schemas/rpc/contract.py
```diff
@@ -87,48 +81,68 @@ def make_dataclass(self, data, **kwargs) -> SierraEntryPointsByType:
...
+    )
+
+    @post_load
+    def make_dataclass(self, data, **kwargs) -> EntryPointsByType:
+        return EntryPointsByType(**data)
+
+
 class SierraContractClassSchema(Schema):
+    sierra_program = fields.List(
+        fields.String(), data_key="sierra_program", required=True
```
ddoktorski, line 102: Actually, it would be better to align it with the [spec](https://github.com/starkware-libs/starknet-specs/blob/e5ce8c6f99672d9ddf85a8e1d6ef8e87169d6e8d/api/starknet_api_openrpc.json#L3039) and have it as `Felt`. To maintain consistency with the rest of the code, we should treat felts as ints. This requires making changes in the [SierraContractClass](https://github.com/software-mansion/starknet.py/blob/da1aa2fad8c3066711c6614f2e3140f444f5cbe9/starknet_py/net/client_models.py#L815) and consequently in the [compute_sierra_class_hash](https://github.com/software-mansion/starknet.py/blob/da1aa2fad8c3066711c6614f2e3140f444f5cbe9/starknet_py/hash/sierra_class_hash.py#L34) function as well.
```suggestion
        Felt(), data_key="sierra_program", required=True
```

starknet_py/net/schemas/rpc/transactions.py
```diff
@@ -0,0 +1,420 @@
...
+
+class DeployAccountTransactionResponseSchema(SentTransactionSchema):
+    address = Felt(data_key="contract_address", required=True)
+
+    @post_load
+    def make_dataclass(self, data, **kwargs) -> DeployAccountTransactionResponse:
+        return DeployAccountTransactionResponse(**data)
+
+
+class PendingStarknetBlockWithReceiptsSchema(PendingBlockHeaderSchema):
```
ddoktorski, line 373: Why are the `Block` schemas from this file not in `rpc/block.py` instead?

starknet_py/net/client_models.py
```diff
@@ -773,7 +773,7 @@ class ContractClass:
 
 
 @dataclass
-class CompiledContract(ContractClass):
+class CompiledDeprecatedContract(DeprecatedContractClass):
```
ddoktorski, line 776:
```suggestion
class DeprecatedCompiledContract(DeprecatedContractClass):
```

starknet_py/net/schemas/rpc/contract.py
```diff
@@ -181,3 +195,28 @@ class CasmClassSchema(Schema):
     @post_load
     def make_dataclass(self, data, **kwargs) -> CasmClass:
         return CasmClass(**data)
+
+
+class DeprecatedCompiledContractSchema(ContractClassSchema):
```
ddoktorski, line 200: Let's move this class to have it right below the `DeprecatedContractClassSchema`.

ddoktorski (changes requested):
Good job 💪

I've left a few comments. In addition to those, let's also update the migration guide with the information about renamed classes (`CompiledContract` and `ContractClass`).

starknet_py/net/schemas/rpc/transactions.py
kkawula: These schemas have been in the `transactions.py` because of their dependency with transaction schemas in this file, moved to `block.py` cause circular import error.

ddoktorski: Please try to fix this. If we move all block schemas to `block.py`, we will have a problem where we depend on `ExecutionResourcesSchema` within `transactions.py`. We could consider creating an additional file, perhaps `general.py` into which we can move elements like `ExecutionResourcesSchema` and `EstimatedFeeSchema`.

starknet_py/net/schemas/rpc/trace_api.py
```diff
@@ -0,0 +1,293 @@
...
+from starknet_py.net.schemas.common import (
+    CallTypeField,
+    EntryPointTypeField,
+    Felt,
+    PriceUnitField,
+)
+from starknet_py.net.schemas.rpc.block import StateDiffSchema
+from starknet_py.utils.schema import Schema
+
+# pylint: disable=unused-argument, no-self-use
```
ddoktorski, line 29: It seems that it can be removed as well.

ddoktorski (approved):
LGTM

PR #1391, opened 2024-07-10
Remove unknown=EXCLUDE after devnet fix response

### LINKED ISSUE #1323: Remove unknown=EXCLUDE after devnet fix response
For `getStateUpdate` and`getBlockWithTxHashes` devnet returns a response that contains more fields than should. Remove unknown=EXCLUDE when Devnet will fix this [issue](https://github.com/0xSpaceShard/starknet-devnet-rs/issues/381). Also set L1HandlerTransactionTraceSchema `execution_resources` as required after starknet update `starknet-spec`

kkawula:
Closes #1323 

I've tested these requests on pending block on devnet and everything is correctly according to the specification

## Introduced changes

- deleted `unknown=EXCLUDE` option from deserializing `PendingStarknetBlockWithTxHashesSchema` and `PendingBlockStateUpdateSchema`
- devnet version bump 0.7.0 -> 0.1.2

##


ddoktorski (commented):
Generally looks good, but I believe `unknown=EXCLUDE` could also be removed from the [get_block](https://github.com/software-mansion/starknet.py/blob/ef4992dbcb259034c596a08315b6b106e4c0cacd/starknet_py/net/full_node_client.py#L108) function

ddoktorski (approved):

PR #1402, opened 2024-07-18
Support Ledger signer

### LINKED ISSUE #1386: Support Ledger Signer
### Feature Request

Ledger now has a [Starknet App](https://support.ledger.com/hc/en-us/articles/16765262529821-Starknet-STRK?docs=true), supporting signing transactions with Stark Curve on Ledger. Any existing Starknet account can now be used with Ledger as a signer. The goal of this issue is to implement a custom `LedgerSigner` as an extension of the `BaseSigner`, which will allow signing transactions in starknet.py using a ledger device.

For reference, starknet-rs implementation:

https://github.com/xJonathanLEI/starknet-rs/blob/db1fa598232f0698d942cc974f481b5d888ac080/starknet-signers/src/ledger.rs

franciszekjob:
Closes #1386 

## Introduced changes

Support signing with Ledger wallet

- Add `LedgerSigner`
- Add `LedgerStarknetApp`
- Add tests for `LedgerSigner`

Testing, building and APDU specification of the Starknet app can be found [here](https://github.com/LedgerHQ/app-starknet )

##


starknet_py/net/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,173 @@
...
+@pytest.mark.asyncio
+async def test_deploy_account_and_transfer(client):
+    signer = LedgerSigner(
+        derivation_path_str="m/2645'/1195502025'/1470455285'/0'/0'/0",
+        chain_id=StarknetChainId.SEPOLIA,
+    )
+
+    class_hash = int(
+        "0x61dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f", 16
+    )
```
ddoktorski, line 112:
```suggestion
    class_hash = 0x61dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f
```

starknet_py/net/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,173 @@
...
+        contract_address_salt=salt,
+        constructor_calldata=calldata,
+        auto_estimate=True,
+    )
+
+    await client.deploy_account(signed_tx)
+
+    recipient_address = int(
+        "0x1323cacbc02b4aaed9bb6b24d121fb712d8946376040990f2f2fa0dcf17bb5b", 16
+    )
```
ddoktorski, line 145:
```suggestion
    recipient_address = 0x1323cacbc02b4aaed9bb6b24d121fb712d8946376040990f2f2fa0dcf17bb5b
```

starknet_py/net/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,173 @@
...
+    sender_balance_before = (await _get_account_balance_eth(client, address))[0]
+    recipient_balance_before = (
+        await _get_account_balance_eth(client, recipient_address)
+    )[0]
+    print(f"Balance of {address} before: {sender_balance_before}")
+    print(f"Balance of {recipient_address} before: {recipient_balance_before}")
```
ddoktorski, line 151:
```suggestion
```

starknet_py/net/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,173 @@
...
+
+    eth_address = int(
+        "0x49D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7", 16
+    )
+
+    call = Call(
```
ddoktorski, line 157: Maybe it would be better to use `Contract` class for interactions with the ETH contract?

starknet_py/tests/e2e/mock/ledger/confirm_sign_hash_automation.json
ddoktorski, line 1: As discussed offline, isn't the automation logic deprecated?

ddoktorski (commented):
Partial review

pyproject.toml
```diff
@@ -29,6 +29,8 @@ furo = { version = "^2024.5.6", optional = true }
 pycryptodome = "^3.17"
 crypto-cpp-py = "1.4.4"
 eth-keyfile = "^0.8.1"
+ledgerwallet = "^0.5.0"
+bip-utils = "^2.9.3"
```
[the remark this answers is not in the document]
franciszekjob: Which library are you referring to? `ledgerwallet`, `bip-utils` or both?

starknet_py/constants.py
```diff
@@ -37,3 +37,11 @@
...
 
 ROOT_PATH = Path(__file__).parent
+
+# Ledger constants
+STARKNET_CLA = 0x5A
+EIP_2645_PURPOSE = 0x80000A55
+EIP_2645_PATH_LENGTH = 6
+PUBLIC_KEY_LENGTH = 65
+SIGNATURE_LENGTH = 65
+VERSION_LENGTH = 3
```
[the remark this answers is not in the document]
franciszekjob: `STARKNET_CLA` is taken from [starknet app APDU spec](https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md).
`EIP_2645_PURPOSE` is hexadecimal value of EIP-2546 purpose
(derivation path structure is m’ / purpose’ / coin_type’ / account’ / change / address_index)
`EIP_2645_PATH_LENGTH` - it's the length of whole path, it should always be 6
The rest are just the length of specific responses. Imo it's good to store them in constants, as they should always remain the same.

starknet_py/net/signer/ledger_signer.py
```diff
@@ -0,0 +1,169 @@
...
+
+        data = _derivation_path_to_bytes(derivation_path)
+        self.client.apdu_exchange(
+            ins=0,
+            data=data,
+            p1=0,
+            p2=0,
+        )
+
+        # for some reason the Ledger app expects the data to be left shifted by 4 bits
```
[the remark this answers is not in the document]
franciszekjob: Unfortunately I could't find any official information about this :/. [APDU spec](https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#command-2) seems like a place in which this should be documented. Initially, without shifting, the test for deploy account and transfer failed because of invalid signature. After shifting, the signature was correct and it passed

In [rust implementation](https://github.com/xJonathanLEI/starknet-rs/blob/f6d339c6b897fb38c839485608ca2fe374a6275d/starknet-signers/src/ledger.rs#L318) there was a same remark.

pyproject.toml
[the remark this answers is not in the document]
franciszekjob: I assume you're referring to `ledgerwallet`. Available libraries are listed [here](https://speculos.ledger.com/user/clients.html), the one that could be potentially utilized is [blue-loader-python](https://github.com/ledgerhq/blue-loader-python). However, it provides a set of scripts that we can ran (it's less low-level), which I don't think is the best choice in our case.

The communication with ledger device is performed via APDU protocol that bases on request/response. Every app (starknet, btc, etc.) has its explicit APDU documentation, which should let us pretty easily determine what is the request and response of each command. `ledgerwallet` lets us to make these data exchanges. It's also more low-level than `blue-loader-python`, which I think is a benefit here.

Unfortunately, I couldn't find any official documentation for `ledgerwallet`.
The advantage of `blue-loader-python` is that it has docs, but as I said - it provides scripts, which don't seem to be adequate in our case.

Let me know what you think about it.

starknet_py/tests/e2e/mock/ledger/confirm_sign_hash_automation.json
franciszekjob: Yes, it is deprecated, even though there is no information about it on the website 🫤 .  
This press automation is a perfect choice in our case as we don't have to perform presses directly from the code. 
Btw I'm wondering what's the actual reason behind marking it as deprecated 🤔 it's a very handy feature.

(Below a screen from speculos depreciation info)
image

docs/guide/signing.rst
```diff
@@ -12,6 +12,30 @@ signing algorithm, it is possible to create ``Account`` with custom
...
+    )
+    client = FullNodeClient(node_url="your.node.url")
+    # Create an `Account` instance with the ledger signer
+    account = Account(
+        client=client,
+        address=0x1111,
+        signer=signer,
+        chain=StarknetChainId.SEPOLIA,
+    )
+    # Now you can use Account as you'd always do
```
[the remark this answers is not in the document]
franciszekjob: Ofc, moved 🫡

starknet_py/net/signer/test_ledger_signer.py
franciszekjob: Sure, updated.

franciszekjob: That's how the docs look now
image
image
Unfortunately, we cannot get rid of the prefund-related code in the second section. Maybe we can just start the example after prefunding, wdyt @ddoktorski ?

.github/workflows/checks.yml
```diff
@@ -197,6 +200,42 @@ jobs:
...
+            cd /apps/app-starknet
+            cargo clean
+            cargo ledger build nanox
+            
+      - name: Start Speculos emulator container
+        uses: addnab/docker-run-action@v3
+        with:
+          image: ghcr.io/ledgerhq/ledger-app-builder/ledger-app-dev-tools
+          options: --rm -d --name speculos-emulator -v ${{ github.workspace }}:/apps --publish 5001:5001 --publish 9999:9999
+          run: speculos -m nanox --apdu-port 9999 --api-port 5001 --display headless /apps/app-starknet/target/nanox/release/starknet --automation '{"version":1,"rules":[{"text":"Confirm Hash to sign","conditions":[],"actions":[["button",2,true],["button",2,false]]},{"regexp":".*Hash \\(.*\\)","conditions":[],"actions":[["button",2,true],["button",2,false]]},{"text":"Reject","conditions":[],"actions":[["button",2,true],["button",2,false]]},{"text":"Approve","conditions":[],"actions":[["button",1,true],["button",2,true],["button",1,false],["button",2,false]]}]}'
```
franciszekjob, line 227: note: This will be moved to separate file. For now I've put the automation json here, in order to check if it works properly.

.github/workflows/checks.yml
```diff
@@ -197,6 +200,48 @@ jobs:
...
+          run: |
+            speculos \
+            -m nanox \
+            --apdu-port 9999 \
+            --api-port 5000 \
+            --display headless \
+            /apps/app-starknet/target/nanox/release/starknet
+
+      - name: Wait for Speculos to start
+        run: sleep 5
```
franciszekjob, line 236: note: We need to wait some time here, without this job `Update automation rules` is failing.

starknet_py/net/signer/ledger_signer.py
franciszekjob, line 1: While reviewing `LedgerStarknetApp` command methods, please refer to [APDU docs](https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md). Methods correspond to located there commands.

starknet_py/net/signer/speculos_automation.json
franciszekjob, line 1: The file could be stored in a different location, wdyt @ddoktorski ?

starknet_py/net/signer/ledger_signer.py
```diff
@@ -0,0 +1,172 @@
...
+        :return: Signature as a list of two integers.
+        """
+
+        data = _derivation_path_to_bytes(derivation_path)
+        self.client.apdu_exchange(
+            ins=0,
+            data=data,
+            p1=0,
+            p2=0,
+        )
```
ddoktorski, line 82: What is the purpose of this command?

starknet_py/net/signer/ledger_signer.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def sign_message(self, typed_data: TypedData, account_address: int) -> List[int]:
+        msg_hash = typed_data.message_hash(account_address)
+        return self.app.sign_hash(
+            derivation_path=self.derivation_path, hash_val=msg_hash
+        )
+
+
+def _parse_derivation_path_str(derivation_path_str) -> Bip32Path:
```
ddoktorski, line 139: nit:
```suggestion
def _parse_derivation_path_str(derivation_path_str: str) -> Bip32Path:
```

starknet_py/net/signer/ledger_signer.py
```diff
@@ -0,0 +1,169 @@
...
+            p1=int(device_confirmation),
+            p2=0,
+        )
+
+        if len(response) != PUBLIC_KEY_LENGTH:
+            raise ValueError(
+                f"Unexpected response length (expected: {PUBLIC_KEY_LENGTH}, actual: {len(response)}"
+            )
+
+        public_key = int.from_bytes(response[1:33], byteorder="big")
```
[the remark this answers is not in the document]
franciszekjob: It's because the length of byte array containing the key should be alway equal to 32.

starknet_py/net/signer/ledger_signer.py
franciszekjob: It lets to check the starknet app version. Please refer to https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#getversion

franciszekjob: It's not needed here, removed.

pyproject.toml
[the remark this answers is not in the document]
franciszekjob: I haven't found aby more official lib than this one. We could always implement needed utils on our own.

starknet_py/net/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,189 @@
...
+        DeployAccountV1(
+            class_hash=0x1,
+            contract_address_salt=0x2,
+            constructor_calldata=[1, 2, 3, 4],
+            max_fee=10000,
+            signature=[],
+            nonce=23,
+            version=1,
+        ),
+        DeclareV1(
```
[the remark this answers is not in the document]
franciszekjob: I changed it to V3 and kept it the same as in [test_sign_transaction](https://github.com/software-mansion/starknet.py/blob/392db46904a9a34362bf2aaea99155bab7b20a21/starknet_py/tests/unit/signer/test_stark_curve_signer.py#L45) for stark curve signer

.github/workflows/checks.yml
```diff
@@ -199,6 +201,48 @@ jobs:
...
+
+      - name: Pull speculos image
+        run: docker pull ghcr.io/ledgerhq/ledger-app-builder/ledger-app-dev-tools
+
+
+      - name: Clone LedgerHQ Starknet app repository
+        run: git clone https://github.com/LedgerHQ/app-starknet.git
+
+      - name: Build the app inside Docker container
+        uses: addnab/docker-run-action@v3
```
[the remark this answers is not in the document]
franciszekjob: Normally, we would build the app and run speculos API in one container. But, we need to keep in mind, that the first command `cargo ledger build nanox` takes about takes some time (usually ~2 minutes, example in the screenshot below). Combined with the fact that container with speculos API needs to run in detached mode, we don't really know how long to wait. We already wait 10 seconds to ensure that API has started. We could set some "safe" time amount to wait (like 3/4 minutes), but I don't think that it's the best solution.

Having two containers, first running in attached mode, we can naturally wait for the building process to finish.

Wdyt @ddoktorski ?
image

starknet_py/tests/unit/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,189 @@
...
+            class_hash=0x1,
+            contract_address_salt=0x2,
+            constructor_calldata=[1, 2, 3, 4],
+            max_fee=10000,
+            signature=[],
+            nonce=23,
+            version=1,
+        ),
+        DeclareV3(
+            contract_class=sierra_contract_class,
```
ddoktorski, line 67: I guess we can use a mock object here, perhaps even for entire transactions. It is possible to achieve for example by `Mock` library from `unittest.mock`.

.github/workflows/checks.yml
```diff
@@ -161,6 +161,8 @@ jobs:
       matrix:
         python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
     env:
+      LEDGER_PROXY_ADDRESS: 127.0.0.1
+      LEDGER_PROXY_PORT: 9999
```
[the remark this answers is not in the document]
franciszekjob: They are needed for the connection with speculos emulator.

.github/workflows/checks.yml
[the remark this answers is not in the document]
franciszekjob: Yup.

starknet_py/tests/unit/signer/test_ledger_signer.py
franciszekjob: Maybe let's address it in separate PR (dedicated for test improvements), as it also applies to the same test for stark curve signer

starknet_py/tests/unit/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,208 @@
...
+
+    await client.deploy_account(signed_tx)
+
+    recipient_address = (
+        0x1323CACBC02B4AAED9BB6B24D121FB712D8946376040990F2F2FA0DCF17BB5B
+    )
+
+    recipient_balance_before = (
+        await _get_account_balance_eth(client, recipient_address)
+    )[0]
```
ddoktorski, line 194: It shouldn't be included in the docs

starknet_py/tests/unit/signer/test_ledger_signer.py
```diff
@@ -0,0 +1,208 @@
...
+
+    contract = await Contract.from_address(
+        provider=account, address=ETH_FEE_CONTRACT_ADDRESS
+    )
+    invocation = await contract.functions["transfer"].invoke_v1(
```
ddoktorski, line 199: nit: It would be better to use `v3` in examples

ddoktorski (approved):
Two more comments to the docs, but overall LGTM 🔥

PR #1449, opened 2024-08-19
Add BoundedInt terminal to grammar definition for parsing ABI

kkawula:
Closes #

BoundedInt was [refactored](https://github.com/starkware-libs/cairo/blob/95fd09bab0cc44c4f101987c62491454890e3ba4/corelib/src/circuit.cairo#L43
) in new release of cairo 2.7.0 and now u96 abi is generated new way

## Introduced changes

- Updated grammar definition for parsing ABI

##

### DIFF
--- a/starknet_py/abi/v2/parser_transformer.py
+++ b/starknet_py/abi/v2/parser_transformer.py
@@ -1,3 +1,4 @@
+from math import log2
 from typing import Any, List, Optional
 
 import lark
@@ -25,6 +26,7 @@ ABI_EBNF = """
         | type_felt
         | type_bytes
         | type_uint
+        | type_bounded_int
         | type_contract_address
         | type_class_hash
         | type_storage_address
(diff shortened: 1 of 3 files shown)


starknet_py/abi/v2/parser_transformer.py
```diff
@@ -109,6 +112,17 @@ def type_uint(self, value: List[Token]) -> UintType:
         """
         return UintType(int(value[0]))
 
+    def type_bounded_int(self, value: List[Token]) -> UintType:
+        """
+        BoundedInt Uint type contains information about its ranges. They are present in the value[0] and value[2].
+        """
+        if value[0] != "0":
+            raise ValueError("BoundedInt should start from 0.")
```
THenry14, line 120: is this documented somewhere, eg in starknet docs/comments in the code/whatever? I'd like to make sure this is actually true :)

kkawula: It isn't documented, but to prevent unexpected behavior in the code, I added this line. I'm not aware of any other usage of BoundedInt. (It isn't documented)

THenry14: sure, so where did you get this info from? Is this assumption safe?

kkawula: In starknet.py we do not have any data type that is limited from left. So I made this assumption myself. I do not know if in ABI can be occurred BoundedInt with lower bound other than 0 eg. `"type": "core::internal::BoundedInt::<32, 63>"`. If not this raise never be raised if yes, users actually won't be able to use their ABI. I think there is other solution, I can remove this 'if' and 'raise' and believe in users that they will be using values higher than lower bound. But I don't think so such datatype exists.

THenry14: ok, thanks for clarification. I guess we can leave this as is for now, but it would be cool if we could seek a clarification on this somehow

THenry14 (approved):

Issue #1334, opened 2024-04-05
Remove declaration of cairo 0 contracts


THenry14:
### Feature Request

Now that it is not possible to declare cairo 0 contracts anymore we should remove a support for it from the sdk altogether

enitrat:
Hi @THenry14  @kkawula,
Starknet-py is not only used for Starknet Mainnet, but also for custom appchains that run Cairo Zero code. As such, we need these features to be back. Could you please revert this commit ?

franciszekjob:
Hey @enitrat, thanks for your report. Will updated you on this once we discuss this idea internally in our team 👍 .

PR #1451, opened 2024-08-21
Remove `sign_declare_v1` and `declare_v1` methods

kkawula:
Relates #1334

## Introduced changes

- Removed `Contract.declare_v1`
- Removed `Account.sign_declare_v1`
- Remove DeclareV1 schema
- Updated docs

##

Declaring contracts made in Cairo 0 is no longer supported
- Removed `Contract.declare_v1`
- Removed `Account.sign_declare_v1`


starknet_py/net/models/transaction.py
```diff
@@ -181,50 +175,6 @@ def calculate_hash(self, chain_id: int) -> int:
...
-        return compute_declare_transaction_hash(
-            contract_class=self.contract_class,
-            chain_id=chain_id,
-            sender_address=self.sender_address,
-            max_fee=self.max_fee,
-            version=self.version,
-            nonce=self.nonce,
-        )
-
-
```
franciszekjob, line 227: Could you please explain why we remove this class? Other declareV1-related classes are kept, because we want to remove the ability for a user to send the declare v1 requests. But they still should be able to get the declare v1 transactions from the network.

franciszekjob (commented):
Left one little question, overall looks good 👍

starknet_py/net/models/transaction.py
kkawula: Correct me if I am wrong but `DeclareV1` was used for declaration only, but for collecting transactions from the block there is used [DeclareTransactionV1Schema](https://github.com/software-mansion/starknet.py/blob/kkawula/remove-sign-declare-v1/starknet_py/net/schemas/rpc/transactions.py#L176) that wasn't removed

franciszekjob: Right, thanks for clarification 👍

franciszekjob (approved):
LGTM

franciszekjob (approved):

Issue #1399, opened 2024-07-17
Improve CI performance


ddoktorski:
The current CI process takes a long time to execute and requires multiple machines. As a result, there are frequent delays in starting the CI as all available machines are often occupied by already running actions.

A few ideas how to improve this process:
1. Limit the number of tests on CI for pull requests, for push to the `development` branch it can stay as it is now
2. Consider running tests on CI for pull requests using only 1 or 2 Python versions (instead of 5)
3. Add a devnet cache mechanism for devnet binary for Windows tests (this could speed up these tests by ~30 minutes)
4. Research potential ways to improve the efficiency of the execution of the tests themselves

PR #1458, opened 2024-08-26
Limit number of python versions used on pull request

franciszekjob:
Relates #1399 

## Introduced changes

Limit python versions on CI - use `3.8` and `3.12`

##


.github/workflows/checks.yml
```diff
@@ -133,13 +133,22 @@ jobs:
   # ---------------------------------------------------------- #
 
   run-tests:
-    name: Tests
+    name: Tests (${{ matrix.python-version }})
```
franciszekjob, line 136: note: I've added displaying python version here because otherwise `is_pr_event` value will be shown as well (as below).
image
`is_pr_event` shows because it's the default behavior for matrix to show all variables of matrix combination.

.github/workflows/checks.yml
franciszekjob, line 1: note: Unfortunately when a job does not start (because event is pull request) python versions is not displayed (as below).
image

Imo it's not a deal braker, wdyt @kkawula ?

kkawula (approved):
LGTM!

PR #1452, opened 2024-08-22
Enable deploying contracts without providing abi

kkawula:
Relates #1409

## Introduced changes

- Made abi parameter in `Contract.deploy_contract_v3` optional

##

- [ ] This PR contains breaking changes

PR #1462, opened 2024-08-30
Update readme

### LINKED ISSUE #1410: Update README
Update README to include relevant information for new users or remove its content and leave only a link to the documentation.

kkawula:
Closes #1410

## Introduced changes

- Added examples in `README.md`

##


franciszekjob (commented):
Generally it looks good, but I wonder if we could use the same content as in [Quickstart](https://starknetpy.readthedocs.io/en/latest/quickstart.html) section in the docs. Quickstart is something light and is suitable for README. This way, we would also maintain consistency between the examples in README <-> Docs.

Let's update Quickstart with:
- example of using contract without abi (which is possible after https://github.com/software-mansion/starknet.py/pull/1452)
- use Cairo 1 instead of Cairo 0
image

and then use it in the README.

Wdyt @kkawula? Do you see other improvements to the Quickstart section?

franciszekjob (approved):

PR #1470, opened 2024-09-04
Remove skipping of certain tests

### LINKED ISSUE #1419: Remove skipping of certain tests
Due to an update of devnet [#547](https://github.com/0xSpaceShard/starknet-devnet-rs/pull/547) the bug was removed which now affects our tests. Declaring multiple contracts with the same class hash isn't possible right now.
The short-term solution was to skip these tests.

kkawula:
Closes #1419

In the sense of the solution, duplicates of contracts that have to be redeclared was added, because contracts are compiled before tests, we cannot salt them in runtime. Additionally, these duplicates were added only in cairoV2 version, to avoid massive code duplication. However, these tests are only used for docs purposes, so this functionality is still covered by other tests that involve two versions of Cairo.

## Introduced changes

- Added copy of contracts that have to be declared more than one time

##


starknet_py/tests/e2e/block_test.py
franciszekjob, line 1: Looks like [this line](https://github.com/software-mansion/starknet.py/pull/1470/files#diff-129e49f80171fec9a84a1e967524c2b0c039dde352ad130be0736198f6a6d83aL67) should be removed too?

starknet_py/tests/e2e/docs/code_examples/test_contract.py
```diff
@@ -66,7 +64,7 @@ async def test_declare_v2(account):
         account,
         compiled_contract=compiled_contract["sierra"],
         compiled_contract_casm=compiled_contract["casm"],
-        max_fee=int(1e15),
+        max_fee=int(1e18),
```
franciszekjob, line 67: What's the reason for this change? Btw, we can use `MAX_FEE` constant.

starknet_py/tests/e2e/docs/guide/test_declaring_contracts.py
```diff
@@ -18,7 +22,7 @@ async def test_declaring_contracts(account, map_compiled_contract_and_class_hash
         compiled_contract=compiled_contract,
         compiled_class_hash=class_hash,
         l1_resource_bounds=ResourceBounds(
-            max_amount=5000, max_price_per_unit=int(1e12)
+            max_amount=int(1e5), max_price_per_unit=int(1e13)
```
franciszekjob, line 25: Could you please explain why did we change this?

starknet_py/tests/e2e/docs/guide/test_declaring_contracts.py
```diff
@@ -1,13 +1,17 @@
...
-# TODO (#1419): Fix contract redeclaration
-@pytest.mark.skip(reason="Redeclaration occurred")
+@pytest.mark.skipif(
+    "--contract_dir=v2" not in sys.argv,
+    reason="Contract exists only in v2 directory",
+)
 @pytest.mark.asyncio
-async def test_declaring_contracts(account, map_compiled_contract_and_class_hash):
-    (compiled_contract, class_hash) = map_compiled_contract_and_class_hash
+async def test_declaring_contracts(account, map_compiled_contract_and_class_hash_copy1):
```
franciszekjob, line 13: nit: Also please apply it to other variables 🙏 .
```suggestion
async def test_declaring_contracts(account, map_compiled_contract_and_class_hash_copy_1):
```

starknet_py/tests/e2e/docs/code_examples/test_contract.py
kkawula: The cost of declaration changed since this contract started being skipped, so it was failing due to `Max fee is smaller than the minimal transaction cost` error .
The content of this test code appears in the documentation, so I suppose that using `MAX_FEE` could be confusing

franciszekjob: Ok, I didn't see it's in docs, thanks for explanation.

franciszekjob (approved):
LGTM

PR #1472, opened 2024-09-04
add contract helper for scarp output

Constantine234:
Closes #

## Introduced changes

- 
- 

##

### DIFF
--- a/scarb_output_importer.py
+++ b/scarb_output_importer.py
@@ -1,0 +1,35 @@
+import json
+
+class ScarbOutputImporter:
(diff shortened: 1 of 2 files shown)


kkawula:
Hi @Constantine234 , unfortunately, your solution does not solve the addressed issue and this code isn't aimed to that problem. Pasting code as output from language models isn't usually a proper solution to anything. For this reason, I have to close this pull request.

Constantine234:
@kkawula actually I tried what I know but didn't work then I thought of asking resources on the net anyways thank you for giving me the privilege to try

Issue #1473, opened 2024-09-05
Update `LedgerSigner`


franciszekjob:
### Feature Request

There's new version of [app-starknet](https://github.com/LedgerHQ/app-starknet/tree/main) available (amongst of all it allows clear signing).

Scope:
- bump [version used on our CI](https://github.com/software-mansion/starknet.py/blob/2a8419009c21628c4514f3c3611f3652f5eba124/.github/workflows/checks.yml#L7) to latest one 
- update [`LedgerSigner`](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/net/signer/ledger_signer.py)
- add necessary tests
- bump `ledger-app-dev-tools` to latest version

PR #1474, opened 2024-09-05
Make dependencies for Ledger support optional

droserasprout:
## Introduced changes

`ledgerwallet` and `bip-utils` dependencies made optional. To use `LedgerSigner` the package needs to be installed with `-E ledger`.

These two libraries require lots of cryptography dependencies that significantly increase size of the virtualenv, Docker images, increase installation time and require compilation on many platforms.

`LateException` helper ensures that `ImportError` exception is raised only when Ledger code is actually used. Lazy imports could in addition improve import times, but require much more code changes.

## Breaking changes

Existing users will need to add `ledger` extra to dependencies after the update. 

## Bonus: workaround

If you're dealing with the same issues, found this PR, and it's still open, here's how to override dependencies in pyproject.toml with PDM package manager. Unfortunately, Poetry doesn't support such tricks.

```toml
[tool.pdm.resolution]
excludes = ["bip-utils", "ledgerwallet"]
```


enitrat:
great, we were stuck because of compatibility issues because of this.

franciszekjob:
@droserasprout thanks for the PR! It  looks good, will review it tomorrow :)

franciszekjob:
@droserasprout Could you please fix linting?

droserasprout:
There are multiple `poetry install` calls in the `checks.yml` workflow. I'm not sure which ones require extras, so added everywhere just in case.

franciszekjob:
Ok, no worries. Will review it right after #1481 is merged, because this way "Tests" job in your PR will not be breaking.

franciszekjob:
@droserasprout at first glance PR looks good, and passes tests (without network tests because env variables are not set).
However, we need to wait once https://github.com/software-mansion/starknet.py/issues/1473 is completed as it may changes `LedgerSigner` implementation. Will keep you updated.

franciszekjob (commented):
Looking good 👍 
I think that we need to install Ledger extra using `poetry install -E ledger` only in `Tests` job (ledger is tested only there), in the rest it can be removed. Could you check it please @droserasprout ?

droserasprout:
Done. I've updated a couple of section headers to make SETUP PYTHON part more copypastable.

franciszekjob:
I think it would be better to address it in separate PR. Could you please just keep changes related to optional ledger installation (in `checks.yml` file) and open other PR for cleanup changes?

franciszekjob (approved):
lgtm

PR #1487, opened 2024-09-11
Use previous Ledger app version and re-enable Ledger signer tests

franciszekjob:
Closes #

## Introduced changes

[Ledger Starknet app](https://github.com/LedgerHQ/app-starknet) latest version has changed APDU command. It doesn't support e.g. signing deploy account transaction) which is used in our tests. Due to it, we use previous version.
Once future release of Starknet app will support features used in out tests, updating versions and tests should be addressed in https://github.com/software-mansion/starknet.py/issues/1473.

##

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -191,6 +191,7 @@ jobs:
       - name: Clone LedgerHQ Starknet app repository
         run: |
           git clone https://github.com/LedgerHQ/app-starknet.git
+          cd app-starknet
           git checkout ${{ env.LEDGER_APP_SHA }}
 
       - name: Build the app inside Docker container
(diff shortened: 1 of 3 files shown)


ddoktorski (approved):
Shouldn't the documentation be updated to mention that the latest starknet app version is not supported?

franciszekjob:
> Shouldn't the documentation be updated to mention that the latest starknet app version is not supported?

I think it's better if we include information that our interface doesn't allow clear-signing transactions yet, instead if allows to blind-sign. This is more better information for an end-user imo. Wdyt @ddoktorski ?

franciszekjob:
offline discussion: We will add info about the app's version, as it's visible when installing and it's worth to include it.

PR #1493, opened 2024-09-17
Move Revision class and field to separate file

sashhh31:
Introduced changes:-

-> Added a new filed named schema/revision.py which has the revision and revisionfeild class which was previously defined in 
     schema/common.py .

-> Fixed a bug that was causing circular imports in between models/typed_data.py and schema/common.py .

##

Breaking changes:-

-> Added a new file at schema called revision.py and added imports at models/typed_data.py and schema/common.py likewise

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "starknet.py",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}
--- a/starknet_py/net/schemas/common.py
+++ b/starknet_py/net/schemas/common.py
@@ -2,12 +2,12 @@ import re
 import sys
 from enum import Enum
 from typing import Any, Mapping, Optional, Union
-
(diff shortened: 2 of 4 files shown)


package-lock.json
franciszekjob, line 1: This file is not needed.

starknet_py/net/schemas/common.py
```diff
@@ -348,30 +351,9 @@ def make_dataclass(self, data, **kwargs):
...
-
-        revisions = [revision.value for revision in Revision]
-        if value not in revisions:
-            allowed_revisions_str = "".join(list(map(str, revisions)))
-            raise ValidationError(
-                f"Invalid value provided for Revision: {value}. Allowed values are {allowed_revisions_str}."
-            )
-
-        return Revision(value)
+class MySchema(Schema):
```
franciszekjob, line 354: This schema:
1. is not needed
2. has very general name

franciszekjob (commented):
Please fix type checking.

PR #1497, opened 2024-09-18
Cleanup `checks.yml` GHA workflow

droserasprout:
## Introduced changes

- Consistent usage of "matrix.python_version"
- Unify section headers
- Don't setup Python twice; use Poetry from pre-installed pipx.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -23,25 +23,30 @@ jobs:
   lint-format-typecheck:
     name: Lint - Format - Typecheck
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ "3.12" ]
     steps:
+
+      # ====================== SETUP ====================== #
 
       - uses: actions/checkout@v4
 
-      - name: Install poetry
-        run: |
-          python -m pip install --upgrade pip
-          pip install poetry
-          poetry config installer.modern-installation false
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -23,25 +23,30 @@ jobs:
...
 
-      - name: Install dependencies
-        run: |
-          poetry install -E ledger
+      - name: Install dependencies (+ledger)
+        run: poetry install -E ledger
+
+      # ====================== LINT, FORMAT, TYPECHECK ====================== #
 
       - name: Check poetry.lock
```
franciszekjob, line 51: We can make one-liner here and in  all other occurrences 😄

.github/workflows/checks.yml
```diff
@@ -154,28 +160,19 @@ jobs:
           name: contract-artifacts
           path: starknet_py/tests/e2e/mock/
 
-      - uses: actions/setup-python@v4
-        with:
-          python-version: "3.12"
-          cache: 'pip'
-
```
franciszekjob, line 161: Looks like this removal may affect `pip` caching? Same remark to other places where it's removed.

franciszekjob (changes requested):

franciszekjob:
Closing due to inactivity.

PR #1520, opened 2024-11-21
Add keys (indexed fields aka topics) to the EventType (abi v2)

m-kus:
Issue: Currently ABI v2 does not preserve the event "kind" information which is necessary for decoding event payloads structured as list of keys + list of data fields.

## Introduced changes

This PR adds new `keys` field to the `EventType` which contains the list of event fields marked with `#[key]` in Cairo code.

##

### DIFF
--- a/starknet_py/abi/v2/shape.py
+++ b/starknet_py/abi/v2/shape.py
@@ -57,7 +57,7 @@ class EventBaseDict(TypedDict):
 
 
 class EventStructMemberDict(TypedParameterDict):
-    kind: Union[Literal["key"], Literal["data"]]
+    kind: Literal["data", "key"]
 
 
 class EventStructDict(EventBaseDict):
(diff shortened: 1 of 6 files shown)


starknet_py/abi/v2/shape.py
```diff
@@ -56,7 +57,7 @@ class EventBaseDict(TypedDict):
 
 
 class EventStructMemberDict(TypedParameterDict):
-    kind: Literal["data"]
+    kind: Union[Literal["key"], Literal["data"]]
```
franciszekjob, line 60:
```suggestion
    kind: Literal["data", "key"]
```

franciszekjob (changes requested):

franciszekjob:
Hi @m-kus, thanks for providing the PR :) left one small comment.

franciszekjob (approved):
Thanks for contribution @m-kus :)

PR #1530, opened 2024-12-02
Support SNIP-9

### LINKED ISSUE #1517: Support SNIP-9
### Feature Request

Add support for [SNIP-9](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md)

Reference PRs in starknet.js:
- https://github.com/starknet-io/starknet.js/pull/1202
- https://github.com/starknet-io/starknet.js/pull/1208

baitcode:
Closes #1517

## Introduced changes

[x] Added outside execution model, defined hashing
[x] Extended base account interface
[x] Added all utilities to deal with SNIP-9 nonce, interface support queries and generation OutsideExecution call
[x] Cleaned up as much as I could to keep changes minimal
[x] Added a bit of docs
[x] Added one test for positive scenario
[x] Pinned hash calculation for OutsideExecution 
[x] Added one test for wrong caller scenario

## Small things I had to rework on the way
[x] Moved broadcast transaction generation method closer to usage to get rid of Cyclic dependency
[x] Bugfix type data generation
[x] Allowed for incomplete definition of ParameterDict as `contains` fields is often missing and linter is.

While implementing the feature I realised better what are the problems of current library implementation. I'd love to discuss those and if it proves that my vision is similar to maintainers I'd love to help fixing those.


baitcode:
@franciszekjob Please, take a look.

franciszekjob:
@baitcode 
> I fixed all the required steps. With sad heart I got to admit, that some tests in the project are really bad.
Those to are not reproducible, as they depend on the account that is global and is shared among all other tests (and not immutable)

I see, we can think about refactoring these tests in separate PR.

> And yes, I need help with test_on_network. I can't reproduce those, neither can I understand what is wrong. You can start the review.

Tests on networks are failing because you haven't added secrets to your forked repository
You need to add:
- `SEPOLIA_RPC_URL`
- `SEPOLIA_ACCOUNT_ADDRESS`
- `SEPOLIA_ACCOUNT_PRIVATE_KEY`

Let me know if it works!

baitcode:
@franciszekjob Added repository secrets on my repo. Still not working.

franciszekjob:
@baitcode are you sure that the url of node is correct? Have a look at this error
https://github.com/software-mansion/starknet.py/actions/runs/12266627198/job/34225144979?pr=1530#step:10:6834

baitcode:
@franciszekjob 
Can you please provide exact instruction how to set up branch, and possibly update contribution documentation. I have everything configured. All values are correct.

image

baitcode:
I can see this in Job SetUp 
image
Why is that happening?

baitcode:
@franciszekjob 
May be this is the problem? How do you think?
image

franciszekjob:
@baitcode 
Yup, you're right. I think it would be possible to configure repo the way there will be "Allow and run" option but I need to discuss if internally first. Will update you on that.

For the time being, you can either make a PR in your forked repo and see the result of network tests, or run them locally.

baitcode:
@franciszekjob 
image

baitcode:
@franciszekjob Let's finish this!

starknet_py/net/schemas/common.py
```diff
@@ -80,6 +80,15 @@ def _is_str_and_valid_pattern(self, value: Any) -> bool:
         )
 
 
+class Selector(NumberAsHex):
+    """
+    Field used to serialize and deserialize selector type.
+    """
+
+    MAX_VALUE = 2**32
+    REGEX_PATTERN = r"^0x(0|[a-fA-F1-9]{1}[a-fA-F0-9]{0,7})$"
```
franciszekjob, line 89: Is this regex taken from spec?

starknet_py/net/account/account.py
```diff
@@ -344,6 +392,54 @@ async def sign_invoke_v1(
...
         return _add_signature_to_transaction(execute_tx, signature)
 
+    async def sign_outside_execution_call(
+        self,
+        calls: Calls,
+        execution_time_bounds: ExecutionTimeBounds,
+        *,
+        caller: AddressRepresentation = ANY_CALLER,
+        nonce: Optional[int] = None,
+        version: Optional[SNIP9InterfaceVersion] = None,
```
franciszekjob, line 402: 1. What's the reason that these need to be passed as keyword arguments?
2. nit: Maybe let's rename `version` to `interface_version`, so the name is a bit more specific?

starknet_py/net/client_models.py
```diff
@@ -115,6 +117,24 @@ def init_with_zeros():
         return ResourceBounds(max_amount=0, max_price_per_unit=0)
 
 
+@dataclass
+class ExecutionTimeBounds:
+    """
+    Dataclass representing time bounds within which the given time bounds.
```
franciszekjob, line 123: Description needs to be fixed

docs/guide/account_and_client.rst
```diff
@@ -46,6 +46,15 @@ Account also provides a way of creating signed transaction without sending them.
     :language: python
     :dedent: 4
 
+Creating "Outside transaction" and executing it. `SNIP-9 <https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md>`_
+---------------------------------------------------------------------------------------------------------------------------
+
+Account also provides a way of creating a call and signing to allow for another account (caller) to execute it later on original account behalf. This will also allow caller to execute calls encoded in that transaction for free (signer will pay the fee).
```
franciszekjob, line 52: > signer will pay the fee

It seems to me that executor pays the fee, not the signer.

franciszekjob (changes requested):
@baitcode  Overall looking good 😃 left some comments

starknet_py/hash/outside_execution.py
```diff
@@ -0,0 +1,118 @@
...
+from starknet_py.net.schemas.common import Revision
+from starknet_py.utils import typed_data as td
+
+SNIP9_INTERFACE_ID_TO_SNIP12_REVISION = {
+    SNIP9InterfaceVersion.V1: Revision.V0,
+    SNIP9InterfaceVersion.V2: Revision.V1,
+}
+
+
+def outside_execution_to_typed_data(
```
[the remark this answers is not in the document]
baitcode: That was my initial implementation. And this change leads to cyclic import error. In short - I can't import TypedData in client_models. So I had to decouple.

starknet_py/net/schemas/common.py
baitcode: Pardon. This one left by mistake.

starknet_py/net/account/account.py
baitcode: 1. To allow for customisation and testing
2. :+1:

docs/guide/account_and_client.rst
baitcode: Good catch. Thank you. My bad, didn't understand spec correctly.

starknet_py/hash/outside_execution.py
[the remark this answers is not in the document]
franciszekjob: Ok I see, created https://github.com/software-mansion/starknet.py/issues/1537#issue-2739779904 for this. Could you please add todo in code?

starknet_py/net/client_models.py
```diff
@@ -1106,3 +1126,37 @@ class BlockTransactionTrace:
...
+            "execute_before": self.execute_before,
+            "calls": [
+                {
+                    "to": call.to_addr,
+                    "selector": call.selector,
+                    "calldata": call.calldata,
+                }
+                for call in self.calls
+            ],
+        }
```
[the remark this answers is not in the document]
baitcode: Spent another 4 hours trying to fix that. 

And then I remembered: I avoided using serialiser here initially because there is no serialiser that would produce good result. Abi with TypedDict would do work actually, but it is unclear how to create new one. I struggle to find specs, I don't quite undestand which Abi version to extend, and honestly not sure if that is the correct vector for contribution.

This method prepares data to be user with CairoSerialisers, to create CallData out this model.

Existent SchemaSerialisers turn everything into string as they are focused on RPC communication, thus, they don't fit the task because Ints are being turned into Felt strings, and CairoSerialiser deprecates those explicitly in the code. So that is the conflict. 

CairoSerialisers do not allow object as an input. 

I feel like I am stuck with this one. I would suggest to fix it in separate PR by extending Abi models. Could you provide documentation on this ones?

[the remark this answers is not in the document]
baitcode: Would be great to use TypedDicts for models as much possible.

baitcode:
@franciszekjob please take a look at the last comment.

franciszekjob:
@baitcode ok, pls request review on github and I will take a loook 👍

starknet_py/net/client_models.py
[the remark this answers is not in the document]
franciszekjob: Ok, se we can simply create `OutsideExecutionSchema` and `CallSchema` and use them here. Here is an example of [`TransactionSchema`](https://github.com/software-mansion/starknet.py/blob/45825840ffd013569b82887a9eed89fa26ff41ee/starknet_py/net/schemas/rpc/transactions.py#L121). 

Here's an example code that should work:

```python
class CallSchema(Schema):
    to_addr = Felt(data_key="contract_address", required=True)
    selector = Felt(data_key="entry_point_selector", required=True)


class OutsideExecutionSchema(Schema):
    caller = Felt(data_key="caller", required=True)
    nonce = Felt(data_key="nonce", required=True)
    execute_after = Uint64(data_key="execute_after", required=True)
    execute_before = Uint64(data_key="execute_before", required=True)
    calls = fields.List(fields.Nested(CallSchema()), data_key="calls", required=True)
```

And then use
```python
return cast(Dict, OutsideExecutionSchema().dump(obj=self))
```

starknet_py/net/account/account.py
```diff
@@ -889,3 +996,17 @@ def _parse_calls_cairo_v1(calls: Iterable[Call]) -> List[Dict]:
         calls=ArraySerializer(_call_description_cairo_v1),
     )
 )
+_transaction_serialiser = StructSerializer(
```
franciszekjob, line 999:
```suggestion
_outside_transaction_serializer = StructSerializer(
```

franciszekjob (changes requested):
We're getting closer 😄 please see my comments and let me know if you need any help 🙌

pyproject.toml
```diff
@@ -139,4 +139,3 @@ exclude = [
     "**/__pycache__",
     "starknet_py/tests/e2e/docs",
 ]
-
```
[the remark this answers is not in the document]
baitcode: No. PLease pardon, I dunno how it slipped in.

starknet_py/net/account/account.py
```diff
@@ -290,6 +299,55 @@ async def get_nonce(
...
+        while retry_count > 0:
+            random_stark_address = KeyPair.generate().public_key
+            if await self._check_outside_execution_nonce(random_stark_address):
+                return random_stark_address
+            retry_count -= 1
+        raise RuntimeError("Failed to generate a valid nonce")
+
+    async def _get_outside_execution_version(
+        self,
+    ) -> Union[OutsideExecutionInterfaceVersion, None]:
```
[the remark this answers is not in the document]
baitcode: Oh. Wow. Thank you!

starknet_py/net/client_models.py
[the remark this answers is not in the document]
baitcode: Ouch. I'm so sorry you had to write all this. That is exactly the code I wrote myself. And it does not work. The problem is precisely what I've described: 
```
Felt(data_key="contract_address", required=True)
```
produces string of 60-ish symbols. And this in turn breaks serialisation further, (as only short strings are supported by CairoSerialiser)[https://github.com/software-mansion/starknet.py/blob/45825840ffd013569b82887a9eed89fa26ff41ee/starknet_py/serialization/data_serializers/felt_serializer.py#L31].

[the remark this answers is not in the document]
baitcode: My code works, because it uses int instead of hex int represeantation for Felt, which CairoSerialiser handles perfectly.

baitcode (commented):
Thank you very much for the review. I should've figure out myself that I need to get rid of snip name and remove ambiguity in "outside"/"external" naming. Good job fixing all that out.

starknet_py/net/models/typed_data.py
```diff
@@ -7,7 +7,7 @@
 from starknet_py.net.schemas.common import Revision
 
 
-class ParameterDict(TypedDict):
+class ParameterDict(TypedDict, total=False):
```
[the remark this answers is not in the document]
baitcode: Yeah. It does. Although it seems as a crotch. The problem is with `contains: Optional[str]` despite the fact that it's defined as optional it is still for same reason needed and triggers validation. I even found a discussion about that behaviour. Let me see...
Oh I found a better (one)[https://stackoverflow.com/questions/68814891/typeddict-does-not-allow-optional-keys].
I'll try to replace with `NotNeeded`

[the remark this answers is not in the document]
baitcode: Oh nooo. We need newer python.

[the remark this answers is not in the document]
baitcode: Please check it out.

baitcode:
@franciszekjob 
For some reason [this](https://github.com/software-mansion/starknet.py/pull/1530#discussion_r1885732975) issue appears twice. I replied [here](https://github.com/software-mansion/starknet.py/pull/1530#discussion_r1885773416)

starknet_py/net/models/typed_data.py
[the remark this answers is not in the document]
franciszekjob: Ok I see, to make a field optional with typed dict we need to use [`NotRequired`](https://peps.python.org/pep-0655/), please take a look and change it 🙏 .

starknet_py/net/client_models.py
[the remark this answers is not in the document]
franciszekjob: Hmm 🤔 , but caller is [contract address](https://github.com/starkware-libs/starknet-specs/blob/6d88b7399f56260ece3821c71f9ce53ec55f830b/api/starknet_api_openrpc.json#L1276) which points to [felt](https://github.com/starkware-libs/starknet-specs/blob/6d88b7399f56260ece3821c71f9ce53ec55f830b/api/starknet_api_openrpc.json#L1299). So why is this not valid?

starknet_py/net/account/account.py
```diff
@@ -290,6 +301,55 @@ async def get_nonce(
...
+        )
+        return bool(is_valid)
+
+    async def get_outside_execution_nonce(self, retry_count=10) -> int:
+        while retry_count > 0:
+            random_stark_address = KeyPair.generate().public_key
+            if await self._check_outside_execution_nonce(random_stark_address):
+                return random_stark_address
+            retry_count -= 1
+        raise RuntimeError("Failed to generate a valid nonce")
```
franciszekjob, line 328: Is this logic (trying to get nonce in loop) described somewhere in [spec](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md)? Could you please explain why such approach?

starknet_py/net/client_utils.py
```diff
@@ -79,10 +77,3 @@ def _is_valid_eth_address(address: str) -> bool:
     A function checking if an address matches Ethereum address regex. Note that it doesn't validate any checksums etc.
     """
     return bool(re.fullmatch("^0x[a-fA-F0-9]{40}$", address))
-
-
-def _create_broadcasted_txn(transaction: AccountTransaction) -> dict:
-    return cast(
-        Dict,
-        BroadcastedTransactionSchema().dump(obj=transaction),
-    )
```
franciszekjob, line 88: Why is this moved?

starknet_py/tests/e2e/docs/guide/test_account_sign_outside_transaction.py
```diff
@@ -0,0 +1,53 @@
...
+    from starknet_py.net.client_models import Call, OutsideExecutionTimeBounds
+
+    # Create a call to put value 20 under key 20. That will be executed
+    # as part of outside execution.
+
+    put_call = Call(
+        to_addr=map_contract.address,
+        selector=get_selector_from_name("put"),
+        calldata=[20, 20],
+    )
```
franciszekjob, line 28: nit: Maybe 1 -> 100? Seems like different key and value is a bit better for example purposes 😅 .
```suggestion
    # Create a call to put value 1 under key 100. That will be executed
    # as part of outside execution.

    put_call = Call(
        to_addr=map_contract.address,
        selector=get_selector_from_name("put"),
        calldata=[1, 100],
    )
```

starknet_py/tests/e2e/docs/guide/test_account_sign_outside_transaction.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Create a special signed execution call. This call can now be executed by
+    # the caller specified. In this case, caller is ANY_CALLER, a special constant
+    # that allows any caller to execute the call.
```
franciszekjob, line 32:
```suggestion
    # Create a special signed execution call. This call can now be executed by
    # the specified caller. In this case, caller is ANY_CALLER, a special constant
    # that allows anyone to execute the call.
```

starknet_py/net/account/base_account.py
```diff
@@ -25,7 +28,45 @@
...
+        interface_version: Optional[OutsideExecutionInterfaceID] = None,
+    ) -> Call:
+        """
+        Creates a call for an outcide execution (SNIP-9 specification).
+
+        :param calls: Single call or list of calls to be executed by outside caller.
+        :param execution_time_bounds: Execution time bounds for the call.
+        :param caller: Address of the caller. IMPORTANT! By default it is ANY_CALLER.
+        :param nonce: Nonce for the transaction. Is populated automatically if not provided.
+        :param interface_version: SNIP-9 interface version. Method will check which version account
```
franciszekjob, line 64:
```suggestion
        :param interface_version: Outside execution interface version. Method will check which version account
```

franciszekjob (commented):
@baitcode 
Left some final comments. We should be good to go once they are resolved 👍

starknet_py/net/client_utils.py
baitcode: 1) It resolved cyclic dependency issue.
2) I decided that moving it closer to usage makes more sense than putting it in some generic utils file. This method is only extracted to be used in tests.

starknet_py/net/account/account.py
baitcode: That's a good question. I just (copied)[https://github.com/starknet-io/starknet.js/blob/6e353d3d50226907ce6b5ad53309d55ed51c6874/src/account/default.ts#L655] that approach from js library.

starknet_py/net/client_utils.py
franciszekjob: Ok, sure

starknet_py/net/client_models.py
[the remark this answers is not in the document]
baitcode: When I meant string I meant primitive python type. Serialisers are serialising everything into it's string representation, thus Felt if string of a certain format, and knowledge of the Logical type is encapsulated within the serialiser. 
image

So we have object, that upon passing through serialiser turns into dict[str, str]. But CairoSerialiser works with primitive python types. So Felt in its logic is just int. And when it is not then there is a branch of logic I've sent you link to.

BTW. Thank you so much for the spec. I'll study it and might come with a refactoring proposal.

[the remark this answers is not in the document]
franciszekjob: Ok, got it. Could you please add [this](https://github.com/software-mansion/starknet.py/issues/1537) todo in comment? We may address it in future, as generally it's better to use schema than construct dict by hand.

starknet_py/tests/e2e/account/outside_execution_test.py
```diff
@@ -0,0 +1,136 @@
...
+            put_call,
+        ],
+        execution_time_bounds=OutsideExecutionTimeBounds(
+            execute_after=datetime.datetime.now() - datetime.timedelta(days=10),
+            execute_before=datetime.datetime.now() - datetime.timedelta(days=9),
+        ),
+        caller=ANY_CALLER,
+    )
+
+    tx = await argent_account.execute_v1(calls=[call], max_fee=MAX_FEE)
```
franciszekjob, line 131: Let's use `execute_v3`

starknet_py/tests/e2e/docs/guide/test_account_sign_outside_transaction.py
```diff
@@ -0,0 +1,53 @@
...
+        execution_time_bounds=OutsideExecutionTimeBounds(
+            execute_after=datetime.datetime.now() - datetime.timedelta(hours=1),
+            execute_before=datetime.datetime.now() + datetime.timedelta(hours=1),
+        ),
+        caller=ANY_CALLER,
+    )
+
+    # Execute the call as a normal invoke transaction
+    # can be executed from any account specified in the caller field
+    tx = await account.execute_v1(calls=[call], max_fee=int(1e18))
```
franciszekjob, line 46: Same as above

starknet_py/tests/e2e/account/outside_execution_test.py
```diff
@@ -0,0 +1,136 @@
...
+        selector=get_selector_from_name("put"),
+        calldata=[20, 20],
+    )
+
+    call = await argent_account.sign_outside_execution_call(
+        calls=[
+            put_call,
+            put_call,
+            put_call,
+        ],
```
franciszekjob, line 123: Why 3 times 😅 ?

starknet_py/tests/e2e/account/outside_execution_test.py
```diff
@@ -0,0 +1,136 @@
...
+    tx = await argent_account.execute_v1(calls=[call], max_fee=MAX_FEE)
+
+    with pytest.raises(TransactionRevertedError) as err:
+        await argent_account.client.wait_for_tx(tx.transaction_hash)
+
+    assert "argent/invalid-caller" in err.value.message
+
+
+@pytest.mark.asyncio
+async def test_account_outside_execution_for_impossible_timebounds(
```
franciszekjob, line 100:
```suggestion
async def test_account_outside_execution_for_impossible_time_bounds(
```

starknet_py/tests/e2e/account/outside_execution_test.py
```diff
@@ -0,0 +1,136 @@
...
+            put_call,
+        ],
+        execution_time_bounds=OutsideExecutionTimeBounds(
+            execute_after=datetime.datetime.now() - datetime.timedelta(hours=1),
+            execute_before=datetime.datetime.now() + datetime.timedelta(hours=1),
+        ),
+        caller=ANY_CALLER,
+    )
+
+    tx = await argent_account.execute_v1(calls=[call], max_fee=MAX_FEE)
```
franciszekjob, line 55: v3 as well

starknet_py/tests/e2e/account/outside_execution_test.py
```diff
@@ -0,0 +1,136 @@
...
+            put_call,
+        ],
+        execution_time_bounds=OutsideExecutionTimeBounds(
+            execute_after=datetime.datetime.now() - datetime.timedelta(hours=1),
+            execute_before=datetime.datetime.now() + datetime.timedelta(hours=1),
+        ),
+        caller=account.address,
+    )
+
+    tx = await argent_account.execute_v1(calls=[call], max_fee=MAX_FEE)
```
franciszekjob, line 91: v3

franciszekjob (changes requested):

franciszekjob:
@baitcode just a couple more comments regarding tests, lmk if you need anything 👍

starknet_py/tests/e2e/account/outside_execution_test.py
baitcode: Felt like it's a slightly better test case.

starknet_py/net/account/account.py
baitcode: Yeah, you are absolutely correct. Spec says: 
```
nonce: this is different from the account’s usual nonce, it is used to prevent signature reuse across executions and doesn’t need to be incremental as long as it’s unique.
```

baitcode:
@franciszekjob I need help. For some reason argent account doesn't execute v3 transactions. Reverts with [error](https://github.com/software-mansion/starknet.py/actions/runs/12352239753/job/34468887696?pr=1530#step:17:977):
```
'argent/invalid-tx-version'
```

For now I've reverted execute_v1 method, as it seems that v3 are actually not supported by argent according to that [report](https://diligence.consensys.io/audits/2024/01/argent-account-argent-multisig-starknet-transaction-v3-updates/).

starknet_py/tests/e2e/docs/guide/test_account_sign_outside_transaction.py
```diff
@@ -0,0 +1,53 @@
...
+    await account.client.wait_for_tx(tx.transaction_hash)
+
+    # docs: end
+
+    receipt = await account.client.get_transaction_receipt(tx_hash=tx.transaction_hash)
+
+    assert receipt.finality_status == TransactionFinalityStatus.ACCEPTED_ON_L2
```
franciszekjob, line 53:
```suggestion
import pytest

from starknet_py.net.client_models import TransactionFinalityStatus


@pytest.mark.asyncio
async def test_account_outside_execution_any_caller(
    account,
    argent_account,
    map_contract,
):
    # pylint: disable=import-outside-toplevel,too-many-locals

    # docs: start
    import datetime

    from starknet_py.constants import ANY_CALLER
    from starknet_py.hash.selector import get_selector_from_name
    from starknet_py.net.client_models import Call, OutsideExecutionTimeBounds

    # Create a call to put value 100 at key 1.
    put_call = Call(
        to_addr=map_contract.address,
        selector=get_selector_from_name("put"),
        calldata=[1, 100],
    )

    # Create an outside execution call. This call can now be executed by
    # the specified caller. In this case, anyone will be able to execute it.

    # Note that signing account does not need to have any funds to sign the transaction.
    call = await argent_account.sign_outside_transaction(
        calls=[
            put_call,
        ],
        # The transaction can be executed in specified timeframe.
        execution_time_bounds=OutsideExecutionTimeBounds(
            execute_after=datetime.datetime.now() - datetime.timedelta(hours=1),
            execute_before=datetime.datetime.now() + datetime.timedelta(hours=1),
        ),
        # Use ANY_CALLER, a special constant that allows anyone to execute the call.
        caller=ANY_CALLER,
    )

    # Now, if you're in specified timeframe, you can perform the outside execution by another account.
    tx = await account.execute_v1(calls=[call], max_fee=int(1e18))
    await account.client.wait_for_tx(tx.transaction_hash)

    # docs: end

    receipt = await account.client.get_transaction_receipt(tx_hash=tx.transaction_hash)

    assert receipt.finality_status == TransactionFinalityStatus.ACCEPTED_ON_L2
```

starknet_py/net/account/account.py
```diff
@@ -344,6 +404,56 @@ async def sign_invoke_v1(
         signature = self.signer.sign_transaction(execute_tx)
         return _add_signature_to_transaction(execute_tx, signature)
 
+    async def sign_outside_execution_call(
```
franciszekjob, line 407:
```suggestion
    async def sign_outside_transaction(
```

starknet_py/net/account/base_account.py
```diff
@@ -25,7 +28,45 @@
...
+    @abstractmethod
+    async def supports_interface(
+        self, interface_id: OutsideExecutionInterfaceID
+    ) -> bool:
+        """
+        Check if the account supports the given outside execution interface. Part of ISRC5 standard.
+        """
+
+    @abstractmethod
+    async def sign_outside_execution_call(
```
franciszekjob, line 48: Applies to other occurrences
```suggestion
    async def sign_outside_transaction(
```

docs/guide/account_and_client.rst
```diff
@@ -46,6 +46,15 @@ Account also provides a way of creating signed transaction without sending them.
     :language: python
     :dedent: 4
 
+Creating "Outside transaction" and executing it. `SNIP-9 <https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md>`_
+---------------------------------------------------------------------------------------------------------------------------
+
+Account also provides a way of creating a call and signing to allow for another account (caller) to execute it later on original account behalf. Account does not need to funded with tokens for transaction to execute as caller will pay the execution fee.
```
franciszekjob, line 52:
```suggestion
Outside execution
-----------------

Outside execution allows a protocol to submit a transaction on behalf of another account. This feature is implemented according to `SNIP-9 <https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md>`_.

Account also provides a way of signing transaction which later can be execute by another account. Signer does not need to be funded with tokens as executor will pay the fee.
```

franciszekjob:
@baitcode please see my comments

starknet_py/net/account/account.py
baitcode: @franciszekjob 
I would oppose that, this is misleading. Real transaction will be signed and sent during the execution. This is not a transaction object, it leaves outside transactions.py package. JS library adopted that syntax for naming, but I disagree. What do you think? I'll only change if you insist.

franciszekjob: @baitcode 

> Real transaction will be signed and sent during the execution.

I'm not sure that's correct tbh. Inside `sign_outside_execution_call` we do use `sign_message`. What do you think? I'm open for discussion :D

franciszekjob (commented):
Could you also resolve error while building docs 🙏 ?

baitcode:
Sure. I'm on it atm.

starknet_py/net/account/account.py
baitcode: >> I'm not sure that's correct tbh. 
What confuses you?
>> we do use sign_message
which also points that this method does not create (or prepare) a transaction, as we use `sign_transaction` there.

I find `sign_outside_execution_call` perfect. You clearly understand that you produce call, you understand that it is for outside execution, and you then use it as a call in other account methods.

franciszekjob: Ok, fine, let's leave it like that.

baitcode: we can rename to `prepare_outside_execution_call` hiding the fact that we sign anything. Which might be a tiny bit better or `create_outside_execution_call` which will create a clear distinction from transaction API.

baitcode:
@franciszekjob Is there anything left?

starknet_py/tests/e2e/account/outside_execution_test.py
```diff
@@ -0,0 +1,132 @@
...
+        selector=get_selector_from_name("put"),
+        calldata=[20, 20],
+    )
+
+    call = await argent_account.sign_outside_execution_call(
+        calls=[
+            put_call,
+            put_call,
+            put_call,
+        ],
```
franciszekjob, line 83:
```suggestion
        calls=[
            put_call,
        ],
```

starknet_py/tests/e2e/account/outside_execution_test.py
```diff
@@ -0,0 +1,132 @@
...
```
franciszekjob, line 47:
```suggestion
        calls=[
            put_call,
        ],
```

franciszekjob (commented):
Just these values in tests + could you please add mention the change in `migration_guide.rst` 🙏 ?

docs/migration_guide.rst
```diff
@@ -12,6 +12,10 @@ Migration guide
 
 1. Added :class:`NonZeroType` in order to fix parsing ABI which contains Cairo`s `core::zeroable::NonZero <https://github.com/starkware-libs/cairo/blob/a2b9dddeb3212c8d529538454745b27d7a34a6cd/corelib/src/zeroable.cairo#L78>`_
 
+.. currentmodule:: starknet_py.net
+
+2. Added (SNIP-9)[https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md] support to :class:`account.account.Account`. Now it's possible to create a :class:`client_models.Call` for outside execution using :meth:`account.account.Account.sign_outside_execution_call`.
+
```
franciszekjob, line 18:
```suggestion
2. Added `SNIP-9 <https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md>`_ support to :class:`~starknet_py.net.account.account.Account`. Now it's possible to create a :class:`~starknet_py.net.client_models.Call` for outside execution using :meth:`~starknet_py.net.account.account.Account.sign_outside_execution_call`.
```

franciszekjob (approved):
Please apply suggested change and we're good to go 🚀

docs/migration_guide.rst
baitcode: Oh... Thank you!

baitcode: How do you view those?

baitcode: It fails now. I'm not a guru at writing documentation with sphinx, tbh... will try to fix.

franciszekjob:
I squashed and merged so it's fine 🤔

franciszekjob:
@baitcode you should see the funds on Only Dust 😄

PR #881, opened 2023-03-20
Fix failing checks (poetry issue)

war-in:
Closes #

## Introduced changes

- 
- 

##

- [ ] This PR contains breaking changes

PR #1543, opened 2025-01-08
Set poetry version to `1.8.5`

franciszekjob:
Closes #

## Introduced changes

Before, our CI used latests version of Poetry. This PR changes it to be fixed 1.8.5.
Recently there has been release of [Poetry 2.0.0](https://pypi.org/project/poetry/2.0.0/), which bumped minimal version of Python from 3.8 to 3.9. ATM, starknet.py uses >=3.8, <3.13, hence the need of using Poetry older than 2.0.0

Without this change, CI is failing, here's an [example run](https://github.com/software-mansion/starknet.py/actions/runs/12667535951).

##


.github/workflows/checks.yml
```diff
@@ -32,7 +32,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install poetry
-          poetry config installer.modern-installation false
```
franciszekjob, line 35: note: This poetry installer config was added in https://github.com/software-mansion/starknet.py/pull/881/files, but it looks like we don't need it anymore, thanks to https://github.com/python-poetry/poetry/pull/9392

Below is part of [Poetry 2.0.0 changelog](https://github.com/python-poetry/poetry/releases/tag/2.0.0) 👇 
image

.github/workflows/checks.yml
```diff
@@ -46,7 +45,7 @@ jobs:
 
       - name: Check poetry.lock
         run: |
-          poetry lock --check
+          poetry check --lock
```
franciszekjob, line 48: note: `poetry lock --check` is deprecated, we should use `poetry check --lock`

PR #1550, opened 2025-01-13
Migration guide for v0.25.0

franciszekjob:


docs/api/data_types.rst
franciszekjob, line 1: note: Not all classes representing Cairo data types were included in docs.

ddoktorski (approved):

PR #1553, opened 2025-01-13
Optimize CI

franciszekjob:
Closes #

## Introduced changes

- Use `-n auto` flag when running tests with `pytest`, which will spawn a number of workers processes equal to the number of available CPUs, and distribute the tests randomly across them, [docs reference](https://pytest-xdist.readthedocs.io/en/stable/)
- `test_call_contract` and `test_get_storage_at` previously used a contract shared across other tests. This PR adjusts them, so they use a separate contract, i.e. they are not influenced by other tests

Comparison (Linux)
Before: ~13/14 minutes
Now: ~7/8 minutes

##

### DIFF
--- a/starknet_py/tests/e2e/client/client_test.py
+++ b/starknet_py/tests/e2e/client/client_test.py
@@ -110,14 +110,14 @@ async def test_get_block_by_number(
 
 
 @pytest.mark.asyncio
-async def test_get_storage_at(client, contract_address):
+async def test_get_storage_at(client, contract_address_2):
     storage = await client.get_storage_at(
-        contract_address=contract_address,
+        contract_address=contract_address_2,
         key=get_storage_var_address("balance"),
         block_hash="latest",
     )
 
-    assert storage == 1897
+    assert storage == 1777
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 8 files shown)


starknet_py/tests/e2e/client/client_test.py
franciszekjob, line 1: note: The initial value of `balance` is `1777` when the contract is deployed. Previously, we assumed it was `1897`, as other tests modified the `balance` value. This dependency made the tests interdependent, relying on changes made by other tests. Now they use separate contract, allowing the tests to run independently and ultimately enabling parallel execution.

franciszekjob: I'm not sure if this test was meant to check storage after increasing balance - if so, we can easily add a test for invoke + check storage. However, this is more a test for contract itself, not for our API.

starknet_py/tests/e2e/client/fixtures/prepare_network.py
```diff
@@ -48,6 +48,24 @@ async def deployed_balance_contract(
     return deploy_result.deployed_contract
 
 
+@pytest_asyncio.fixture(scope="package")
+async def deployed_balance_contract_2(
```
franciszekjob, line 52: note: `deployed_balance_contract_2 ` name could be potentially improved.

starknet_py/tests/e2e/client/fixtures/prepare_network.py
```diff
@@ -139,6 +157,15 @@ def contract_address(prepare_network: Tuple[str, PreparedNetworkData]) -> int:
     return prepared_data.contract_address
 
 
+@pytest.fixture()
+def contract_address_2(prepare_network: Tuple[str, PreparedNetworkData]) -> int:
```
ddoktorski, line 161: nit: Maybe add a short comment why this was added and what is the contract?

ddoktorski (approved):

starknet_py/tests/e2e/client/fixtures/prepare_network.py
franciszekjob: Added

Issue #1498, opened 2024-09-19
Support RPC 0.8.0


franciszekjob:
### Feature Request

Add support to JSON-RPC v0.8.0
https://github.com/starkware-libs/starknet-specs/releases/tag/v0.8.0-rc0

TobieTom:
Hola I'm Tobias. I would like to take over this issue. I've already contributed to some projects recently since I just learned Rust and Cairo. I also have experience with js and typescript from web app projects in uni

PR #1563, opened 2025-03-03
Fix and update network tests

franciszekjob:
Towards #1498

## Introduced changes

- Fix network tests
- Temporarily skip network trace api tests

##


.github/workflows/checks.yml
```diff
@@ -314,8 +314,13 @@ jobs:
         run: |
           poetry run poe circular_imports_check
 
+      - uses: asdf-vm/actions/setup@v3
+
+#     TODO(#1498): Remove contracts compilation in the step below once issue with compilation is resolved
       - name: Run tests
         run: |
+          poetry run poe compile_contracts v2
+          poetry run poe compile_contracts v1
```
franciszekjob, line 323: For some reason, compiled contract can't be found while running tests. Here is example CI run which this problem:
https://github.com/software-mansion/starknet.py/actions/runs/13627219872/job/38088663029

This is a temporary adjustment, which will be addresses in incoming PR.

ddoktorski (commented):
Which Sepolia node are we currently using?

ddoktorski (approved):

PR #1573, opened 2025-03-30
Update conf.py

xiaobei0715:
Update the copyright year to 2025.


franciszekjob:
@xiaobei0715 Thanks for the PR. However, the change does not seem necessary and would require updating this file in the future, so we decided to leave it as it is.

PR #1580, opened 2025-04-04
Fix `Contract.deploy_contract_v3`

### LINKED ISSUE #1576: [BUG] Broken deploy flow when not providing ABI
### What happened

https://github.com/software-mansion/starknet.py/blob/118ad51b4c18ed5a2194f871251b6f67f13a2799/starknet_py/contract.py#L761-L776

if ABI is None, it fails with   
Client failed. Message: No contract with address 1647058306414700151140152384281970237772553444278635152170469383502666898324 found.

I think it's going too quick between the tx and the contract fetch. 

Just a slighlty a bit after, the contract is deployed and fetchable. 

### Stack trace

Client failed. Message: No contract with address 14942430588765132354243736532069527212135552510424923206796737841258838809 found.

### Steps to reproduce

deploy a contract v3 without abi

### SDK Version

0.24.3

### Python version

3.10

Linux

- [x] I have searched the existing issues and verified no issue exits for this problem.

franciszekjob:
Closes #1576

## Introduced changes

This PR solves the issue when using `Contract.deploy_contract_v3()` which caused the node to return error that contract can't be found. It happened because right after contract deployment, we tried to fetch its contract class by address.

Now we fetch contract class by class hash.

##


ddoktorski (approved):

franciszekjob:
Checked on testnet ✅ 
image

PR #1593, opened 2025-04-14
Print warning when using incompatible RPC version

### LINKED ISSUE #1592: Print warning on different RPC version of node and SDK
### Feature Request

In the scope of this task, we should add a warning when someone uses node with different version supported by the SDK.

franciszekjob:
Closes #1592

## Introduced changes

This PR adds mechanism, which will print warning when incompatible RPC version is used (only first call)

##


starknet_py/net/http_client.py
```diff
@@ -97,6 +102,28 @@ def handle_rpc_error(result: dict):
...
+                http_method=HttpMethod.POST, address=self.url, payload=payload
+            )
+            spec_version = res["result"]
+
+            if spec_version != EXPECTED_RPC_VERSION:
+                warnings.warn(
+                    f"RPC node with the url {self.url} uses incompatible version {spec_version}. "
+                    f"Expected version: {EXPECTED_RPC_VERSION}",
+                    IncompatibleRPCVersionWarning,
+                    stacklevel=4,
```
franciszekjob, line 123: note: `stacklevel` is set to `4`, so the warning will come from actual file, rather than from http client, e.g. of how it looks below

image

ddoktorski (approved):

PR #1603, opened 2025-05-12
Support Ledger clear signing

### LINKED ISSUE #1473: Update `LedgerSigner`

franciszekjob:
Closes #1473 

## Introduced changes

Add support for clear singing in `LedgerSigner`. Now we're getting displayed tx details before approving/rejecting it.
- Add `LedgerSigningMode`
- Use clear signing by default

https://github.com/user-attachments/assets/92771b9a-3843-4d2f-abef-0b331555ca7e

##


starknet_py/constants.py
```diff
@@ -39,14 +39,6 @@
...
 ROOT_PATH = Path(__file__).parent
 
-# Ledger constants
-STARKNET_CLA = 0x5A
-EIP_2645_PURPOSE = 0x80000A55
-EIP_2645_PATH_LENGTH = 6
-PUBLIC_KEY_RESPONSE_LENGTH = 65
-SIGNATURE_RESPONSE_LENGTH = 65
-VERSION_RESPONSE_LENGTH = 3
-
```
franciszekjob, line 49: note: These variables aren't used across multiple files so let's keep them in the dedicated one.

starknet_py/tests/unit/signer/test_ledger_signer.py
```diff
@@ -158,9 +203,7 @@ async def test_deploy_account_and_transfer(client):
 
     await client.deploy_account(signed_tx)
 
-    recipient_address = (
-        0x1323CACBC02B4AAED9BB6B24D121FB712D8946376040990F2F2FA0DCF17BB5B
-    )
+    recipient_address = 0x123
```
franciszekjob, line 206: note: `0x123` is better as example address.

starknet_py/tests/unit/signer/speculos_automation.json
```diff
@@ -2,10 +2,13 @@
   "version": 1,
   "rules": [
     {
-      "text": "Confirm Hash to sign",
-      "conditions": [],
+      "text": "Blind igning",
```
franciszekjob, line 5: note: It's not a typo - for some reason that's how ledger display the text

starknet_py/net/signer/ledger_signer.py
franciszekjob, line 1: note: When reviewing passed APDUs, please refer to https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +120,365 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+    Device displays the full transaction payload (amounts, addresses, data) so you
+    can review and explicitly approve exactly what you’re signing.
+    """
+
+    BLIND = "blind"
+    """
+    Device omits transaction details and simply asks for your signature, preventing
+    you from verifying the contents and leaving you vulnerable to unknowingly authorizing
+    malicious or unintended actions. ⚠️ Use at your own risk.
+    """
```
franciszekjob, line 148: note: Maybe we should also print a warning when blind signing mode is used?

docs/guide/signing.rst
```diff
@@ -15,7 +15,7 @@ signing algorithm, it is possible to create ``Account`` with custom
 Signing with Ledger
 -------------------
 :ref:`LedgerSigner` allows you to sign transactions using a Ledger device. The device must be unlocked and Starknet app needs to be open.
-Currently used version of Starknet app is ``1.1.1`` and only blind-signing is possible. Clear-signing will be available in the near future.
+Currently used version of `Starknet app <https://github.com/ethereum/ercs/blob/master/ERCS/erc-2645.md>`_ is ``2.3.1``.
```
ddoktorski, line 18: Is it the correct link?

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +120,365 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+        common_tx_fields = tx.get_common_fields(
+            tx_prefix=TransactionHashPrefix.DEPLOY,
+            address=contract_address,
+            chain_id=self.chain_id,
+        )
+        da_mode_hash_bytes = common_tx_fields.get_data_availability_modes().to_bytes(
+            32, byteorder="big"
+        )
+        class_hash_bytes = tx.class_hash.to_bytes(32, byteorder="big")
+        salt_bytes = tx.contract_address_salt.to_bytes(32, byteorder="big")
```
ddoktorski, line 223: nit: Maybe some kind of a helper function/struct to convert V3 transactions to the same struct but with fields as bytes?

starknet_py/net/signer/ledger_signer.py
franciszekjob: Sure, let me add it.

docs/guide/signing.rst
franciszekjob: Should be https://github.com/LedgerHQ/app-starknet, thanks

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +120,365 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
 
     @property
     def public_key(self) -> int:
-        return self.app.get_public_key(derivation_path=self.derivation_path)
+        return self.app.get_public_key()
 
     def sign_transaction(self, transaction: AccountTransaction) -> List[int]:
+        if self.signing_mode == LedgerSigningMode.CLEAR:
+            if isinstance(transaction, DeclareV3):
+                raise ValueError("DeclareV3 signing is not supported bye LedgerSigner")
```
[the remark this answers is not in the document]
franciszekjob: Not really, haven't digged into the reason.

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +120,365 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+
+    if application_name == "LedgerW":
+        application_bytes = (
+            (43).to_bytes(1, byteorder="big")
+            + (206).to_bytes(1, byteorder="big")
+            + (231).to_bytes(1, byteorder="big")
+            + (219).to_bytes(1, byteorder="big")
+        )
+    else:
+        application_bytes = _string_to_4byte_hash(application_name)
```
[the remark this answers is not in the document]
franciszekjob: Seems like these four integers are predefined for LedgerW and in other cases we just generate the representation.
I referred to https://github.com/starknet-io/starknet.js/blob/050616221a58545d79eece7040b4abc205c220e0/src/signer/ledgerSigner221.ts#L644

starknet_py/net/signer/ledger_signer.py
franciszekjob: Hmm, potentially we could add a method for txs classes which would return their values as bytes, but that would require additional struct as the return type (cause returning as tuple of bytes wouldn't be very convenient). Tbh I'm not sure if adding extra code would improve current state, as now it's fairly simple and readable. Wdyt?

.github/workflows/checks.yml
```diff
@@ -2,9 +2,9 @@ name: Checks
 
 env:
   CAIRO_LANG_VERSION: "0.13.1"
-  DEVNET_SHA: fc5a2753a2eedcc27eed7a4fae3ecac08c2ca1b4
-  LEDGER_APP_SHA: db93a5e33d00bd44752fdc8c07aefcffa08d91c3
-  LEDGER_APP_DEV_TOOLS_SHA: a037d42181f4bed9694246256e2c9e2a899e775c302a9c6482c81f87c28e1432
+  DEVNET_SHA: fc5a2753a2eedcc27eed7a4fae3ecac08c2ca1b4 # v0.3.0
```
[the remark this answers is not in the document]
franciszekjob: Nothing forced us to bump it, so it wasn't urgent. Bumped it to v0.4.1.

starknet_py/tests/e2e/fixtures/devnet.py
```diff
@@ -52,13 +56,20 @@ def get_start_devnet_command(devnet_port: int, fork_mode: bool = False) -> List[
             ]
         )
 
+    if predeclare_argent:
```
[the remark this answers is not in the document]
franciszekjob: Seems like nothing prevents use from having it always predeclared. Changed.

starknet_py/net/signer/ledger_signer.py
ddoktorski: Currently we have to use `pylint: disable=too-many-locals` so I think it is worth making some refactoring, but this is a nitpick, so leaving it up to you

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
         return self.app.sign_hash(hash_val=tx_hash)
 
     def sign_message(self, typed_data: TypedData, account_address: int) -> List[int]:
         msg_hash = typed_data.message_hash(account_address)
         return self.app.sign_hash(hash_val=msg_hash)
 
+    # pylint: disable=no-self-use
+    def _decode_signature(self, response: bytes) -> List[int]:
+        r = int.from_bytes(response[33:65], byteorder="big")
+        s = int.from_bytes(response[65:97], byteorder="big")
```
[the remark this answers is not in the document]
franciszekjob: Tx hash and signature length

https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#response-when-tx-is-complete

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+        account_id: int = 0,
+        application_name: str = "LedgerW",
+        signing_mode: LedgerSigningMode = LedgerSigningMode.CLEAR,
+    ):
+        # pylint: disable=line-too-long
         """
-        :param derivation_path_str: Derivation path string of the account.
-        :param chain_id: ChainId of the chain.
+        :param chain_id: Chain ID.
+        :param account_id: ID of Ledger account.
```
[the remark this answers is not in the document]
franciszekjob: Added one sentence with explanation that it's incremental, I think it should be enough.

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+            p1=4,
+            p2=0,
+        )
 
+        # Command 5: Send constructor calldata
+        constructor_bytes = b"".join(
+            val.to_bytes(32, byteorder="big") for val in tx.constructor_calldata
+        )
+        constructor_chunks = []
+        chunk_size = 7 * 32  # 224 bytes
```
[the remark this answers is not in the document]
franciszekjob: Actually, it looks like it should be 192 (6 * 32). See [spec](https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#command-5-constructor-calldata) - there is 6 parts responsible for calldata.

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+    if calldata_size > 0:
+        calldata_size_bytes = calldata_size.to_bytes(32, byteorder="big")
+        calldata_bytes = calldata_size_bytes + b"".join(
+            val.to_bytes(32, byteorder="big") for val in calldata
+        )
+    else:
+        calldata_bytes = int(0).to_bytes(32, byteorder="big")
+
+    call_bytes = to_addr_bytes + selector_bytes + calldata_bytes
+    calldatas = []
```
ddoktorski, line 451:
```suggestion
    calldata_chunks = []
```

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+        response = None
+
+        offset = 1
+        while offset < len(tx.calldata):
+            serialized_call_size = (
+                1 + 1 + 1 + tx.calldata[offset + 2]
+            )  # to_addr + selector + calldata_size + calldata
+            serialized_call = tx.calldata[offset : offset + serialized_call_size]
+
+            calldatas = _call_to_bytes(serialized_call)
```
ddoktorski, line 385:
```suggestion
            calldata_chunks = _call_to_bytes(serialized_call)
```

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+            response = self.app.client.apdu_exchange(
+                ins=3,
+                data=bytes(calldatas[0]),
+                p1=6,
```
ddoktorski, line 389: Why is it 6 not 5?
https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#new-call

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+                p2=0,
+            )
+
+            if len(calldatas) > 1:
+                for part in calldatas[1:]:
+                    response = self.app.client.apdu_exchange(
+                        ins=3, p1=6, p2=1, data=part
```
ddoktorski, line 396: Shouldn't the last call be with `p2=2`?
https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#end-of-calldata-for-the-current-call

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+                data=chunk,
+                p1=5,
+                p2=0,
+            )
 
-def _derivation_path_to_bytes(derivation_path: Bip32Path) -> bytes:
-    """
-    Convert a derivation path to a bytes object.
+        if response is None:
+            raise ValueError("No response received from Ledger device.")
```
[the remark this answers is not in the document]
franciszekjob: No, exchange will always either raise an error or return response, so we mostly want to check if the loop was run.

starknet_py/tests/unit/signer/test_ledger_signer.py
```diff
@@ -3,43 +3,71 @@
...
+                0x123,
+                100,
+                0,
+            ],
+            sender_address=0x123,
+        ),
+        DeployAccountV3(
+            class_hash=0x123,
+            contract_address_salt=0x123,
+            constructor_calldata=[1, 2, 3],
```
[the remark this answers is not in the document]
franciszekjob: Added `test_invoke_v3_long_calldata`

starknet_py/net/signer/ledger_signer.py
franciszekjob: I think it's a typo - look that `l1` is 5 [here](https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#command-4-number-of-calls). Will add an issue in their repo for that.

starknet_py/net/signer/ledger_signer.py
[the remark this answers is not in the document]
ddoktorski: In this case it seems a bit better to check if `constructor_chunks` is not empty before executing the loop - then we are sure that response won't be `None`, but you can treat this as a nitpick.

starknet_py/net/signer/ledger_signer.py
ddoktorski: I think you are right, but it would still be good to get confirmation from them.

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+        calldata_bytes = calldata_size_bytes + b"".join(
+            val.to_bytes(32, byteorder="big") for val in calldata
+        )
+    else:
+        calldata_bytes = int(0).to_bytes(32, byteorder="big")
+
+    call_bytes = to_addr_bytes + selector_bytes + calldata_bytes
+    calldatas = []
+
+    chunk_size = 6 * 32  # 192 bytes
```
ddoktorski, line 453: https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#add-calldata-to-the-current-call
Actually shouldn't it be here 7? It looks that for the constructor calldata in `deploy_account` the chunk size is 6, but for `invoke` it is 7 🤔

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+            ins=3,
+            data=paymaster_bytes,
+            p1=3,
+            p2=0,
+        )
+
+        # Command 4: Send account deployment data
+        account_deployment_bytes = b"".join(
+            val.to_bytes(32, byteorder="big") for val in tx.account_deployment_data
+        )
```
ddoktorski, line 355: Do we need this given that `account_deployment_data` is always empty?

starknet_py/net/signer/ledger_signer.py
ddoktorski: Not updated

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
+                1 + 1 + 1 + tx.calldata[offset + 2]
+            )  # to_addr + selector + calldata_size + calldata
+            serialized_call = tx.calldata[offset : offset + serialized_call_size]
+
+            calldatas = _call_to_bytes(serialized_call)
+            response = self.app.client.apdu_exchange(
+                ins=3,
+                data=bytes(calldatas[0]),
+                p1=6,
+                p2=0,
```
ddoktorski, line 390: Shouldn't `p2` be `0` only for the new call and then `1` for next calls?

starknet_py/net/signer/ledger_signer.py
```diff
@@ -114,64 +121,382 @@ def sign_hash(self, hash_val: int) -> List[int]:
...
-    ]
 
-    if len(path_elements) != EIP_2645_PATH_LENGTH:
-        raise ValueError(f"Derivation path is not {EIP_2645_PATH_LENGTH}-level long")
-    if path_elements[0] != EIP_2645_PURPOSE:
-        raise ValueError("Derivation path is not prefixed with m/2645.")
+        # Command 3: Send paymaster data
+        paymaster_bytes = b"".join(
+            val.to_bytes(32, byteorder="big") for val in tx.paymaster_data
+        )
```
ddoktorski, line 260: Similarly as for `account_deployment_data` do we need this logic for always empty field?

starknet_py/tests/unit/signer/test_ledger_signer.py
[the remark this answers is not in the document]
ddoktorski: It would be good to have the chunk logic tested also for deploy account transaction

starknet_py/net/signer/ledger_signer.py
franciszekjob: It's indeed updated

starknet_py/net/signer/ledger_signer.py
franciszekjob: I would leave it as it's mentioned in APDU spec so we don't have a discrepancy.

starknet_py/net/signer/ledger_signer.py
franciszekjob: Note that this field may not be always empty - starknet 0.15.0 introduces paymaster support so thinking ahead it's good to have it like that.

starknet_py/net/signer/ledger_signer.py
ddoktorski: But in APDU spec it doesn't send any data: https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md#command-4-send-account-deployment-data

starknet_py/net/signer/ledger_signer.py
ddoktorski: Not sure, I guess it may still require some updates on our side. At the moment we officially state in the docs that the supported Starknet app version is `2.3.1`, which doesn't support sending paymaster data.

franciszekjob: Fine, but as I said above - I would leave it like that as it's explicitly adheres to APDU spec.

franciszekjob: Removed

starknet_py/net/signer/ledger_signer.py
franciszekjob: https://github.com/LedgerHQ/app-starknet/issues/54

PR #1620, opened 2025-07-08
Tip support

### LINKED ISSUE #1601: Add tip functionality
### Feature Request

In v0.14.0 of Starknet, transactions are expected to pay a nonzero tip. It would be great to enable building the various txs with the option of putting a tip. E.g. removing `init=False` [here](https://github.com/software-mansion/starknet.py/blob/be84633efe1ded2debd8ff964563b6feba5e1f8f/starknet_py/net/models/transaction.py#L97) and having it as an option in the various Account methods for building/signing/sending txs

cptartur:
Closes #1601

## Introduced changes

- Added support for tip

##

- Dataclasses representing txs now take tip as an required parameter.
- Changed interface of `BaseAccount` so it includes the tip parameter (with default of `0`).

### DIFF
--- a/docs/migration_guide.rst
+++ b/docs/migration_guide.rst
@@ -32,6 +32,36 @@ Version 0.28.0 of **starknet.py** comes with support for RPC 0.9.0-rc.1!
 3. :meth:`Client.wait_for_tx` will no longer raise ``TransactionRejectedError``, see the method docs for details.
 4. :meth:`Client.get_messages_status`: changed ``transaction_hash`` type from ``str`` to ``Hash``.
 
+Tip Support
+-----------
+
+Ability to pass tip for the transaction has been added to following methods.
+These methods will use the default value o ``0`` for tip if not provided.
+
+.. py:currentmodule:: starknet_py.contract
+
+- :meth:`DeclareResult.deploy_v3`
+- :meth:`PreparedFunctionInvokeV3.invoke`
+- :meth:`Contract.declare_v3`
+- :meth:`Contract.deploy_contract_v3`
+
+.. py:currentmodule:: starknet_py.net.account.account
+
+- :meth:`Account.sign_invoke_v3`
+- :meth:`Account.sign_declare_v3`
+- :meth:`Account.sign_deploy_account_v3`
+- :meth:`Account.execute_v3`
+- :meth:`Account.deploy_account_v3`
+
+Additionally, dataclasses representing transactions now require passing a tip.
+No default value is used for tip and it is a required parameter.
+
+.. py:currentmodule:: starknet_py.net.models.transaction
+
+- :class:`InvokeV3`, tip is now required
+- :class:`DeclareV3`, tip is now required
+- :class:`DeployAccountV3`, tip is now required
+
 **********************
 0.27.0 Migration guide
 **********************
--- a/starknet_py/net/models/transaction.py
+++ b/starknet_py/net/models/transaction.py
@@ -94,10 +94,10 @@ class _DeprecatedAccountTransaction(AccountTransaction, ABC):
 @dataclass(frozen=True)
 class _AccountTransactionV3(AccountTransaction, ABC):
     resource_bounds: ResourceBoundsMapping
-    tip: int = field(init=False, default=0)
     nonce_data_availability_mode: DAMode = field(init=False, default=DAMode.L1)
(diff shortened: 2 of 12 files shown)


docs/migration_guide.rst
```diff
@@ -32,6 +32,36 @@ Version 0.28.0 of **starknet.py** comes with support for RPC 0.9.0-rc.1!
 3. :meth:`Client.wait_for_tx` will no longer raise ``TransactionRejectedError``, see the method docs for details.
 4. :meth:`Client.get_messages_status`: changed ``transaction_hash`` type from ``str`` to ``Hash``.
 
+Tip Support
+-----------
+
+Ability to pass tip for the transaction has been added to following methods.
+These methods will use the default value o ``0`` for tip if not provided.
```
franciszekjob:
```suggestion
Ability to pass tip for the transaction has been added to following methods.
If ``tip`` is not provided, a default value of ``0`` will be used
```

starknet_py/tests/e2e/account/account_test.py
```diff
@@ -646,3 +647,68 @@ async def test_account_execute_v3(account, deployed_balance_contract):
         call=get_balance_call
     )
     assert initial_balance + 100 == balance_after_increase
+
+
+@pytest.mark.asyncio
+async def test_invoke_with_tip(account, hello_starknet_class_hash):
```
franciszekjob: nit: We always put version in the name:
```suggestion
async def test_invoke_with_tip(account, hello_starknet_class_hash):
```

starknet_py/tests/e2e/contract_interaction/declare_test.py
```diff
@@ -28,3 +28,22 @@ async def test_throws_when_cairo1_without_compiled_contract_casm_and_class_hash(
             compiled_contract=compiled_contract,
             resource_bounds=MAX_RESOURCE_BOUNDS,
         )
+
+
+@pytest.mark.asyncio
+async def test_declare(
```
franciszekjob: Shouldn't that be `test_declare_v3_with_tip`?:
```suggestion
async def test_declare(
```

starknet_py/net/models/transaction.py
```diff
@@ -94,10 +94,10 @@ class _DeprecatedAccountTransaction(AccountTransaction, ABC):
 @dataclass(frozen=True)
 class _AccountTransactionV3(AccountTransaction, ABC):
     resource_bounds: ResourceBoundsMapping
-    tip: int = field(init=False, default=0)
     nonce_data_availability_mode: DAMode = field(init=False, default=DAMode.L1)
     fee_data_availability_mode: DAMode = field(init=False, default=DAMode.L1)
     paymaster_data: List[int] = field(init=False, default_factory=list)
+    tip: int
```
franciszekjob, line 100: It would make sense to make a check that tip is >= 0, can we add it?

franciszekjob: As follow up, we can add a test with negative tip.

starknet_py/tests/e2e/contract_interaction/declare_test.py
cptartur: Added `v3` to the name, but I couldn't find a general `test_declare_v3` test like that (except for one used for docs) so I though maybe I can treat it as a "default" test.

starknet_py/net/models/transaction.py
cptartur: This should be enforced by marshmallow, it is typed a u64 there.

I've checked, if you pass a negative tip we currently fail when computing the transaction hash to be signed, we get en error from poseidon.py ` ValueError: -123456 out of u_int256_t range` (for too large non-negative value, marshmallow checks will catch it with `marshmallow.exceptions.ValidationError: Invalid value provided for Uint64: 73786976294838206464`).

Not sure if we should have any explicit checks for the `tip` value, we don't check `nonce` provided by the user or `ResourceBounds`.

franciszekjob: Hm ok let's leave it like that then.

franciszekjob (approved):
Might be worth to mention tip functionality somewhere in the guide, maybe in https://starknetpy.readthedocs.io/en/latest/guide/account_and_client.html ?

cptartur:
> Might be worth to mention tip functionality somewhere in the guide, maybe in https://starknetpy.readthedocs.io/en/latest/guide/account_and_client.html ?

Will add that

PR #1625, opened 2025-07-20
Fix fixed size arrays serialization

franciszekjob:
Closes #

## Introduced changes

In https://github.com/software-mansion/starknet.py/pull/1607 we've indeed fixed handling fixed size arrays in ABI parser, but we didn't handle it in [`serializer_for_type`](https://github.com/software-mansion/starknet.py/blob/58a52c55242e407cd009cebe58fcf57c21be3bd1/starknet_py/serialization/factory.py#L75). This caused error e.g. when calling `Contract.deploy_contract_v3` for a contract which included such type.

##


starknet_py/serialization/factory.py
```diff
@@ -101,7 +102,7 @@ def serializer_for_type(cairo_type: CairoType) -> CairoDataSerializer:
             )
         )
 
-    if isinstance(cairo_type, ArrayType):
+    if isinstance(cairo_type, (ArrayType, FixedSizeArrayType)):
         return ArraySerializer(serializer_for_type(cairo_type.inner_type))
```
franciszekjob, line 106: note: I don't think there's a need to create separate serializer; the aspect of "fixed size" only matters when parsing.

starknet_py/tests/e2e/fixtures/abi_structures.py
```diff
@@ -56,6 +58,7 @@
         name_len=FeltType(),
         name=ArrayType(FeltType()),
         pool_id=pool_id_struct,
+        favorite_numbers_triplet=ArrayType(FeltType()),
```
franciszekjob, line 61: note: Even though `favorite_numbers_triplet` is a fixed size array at Cairo level, here there's need to be `ArrayType`. It's because both have same type `felt*`.

franciszekjob:
note: network tests are failing because our node uses 0.9.0-rc.2 (our expected version is rc.1)

starknet_py/tests/e2e/fixtures/abi_structures.py
cptartur: What's the action on user side? Should they just use `ArrayType` if they have fixed size array? 

Maybe it'd be better UX wise if we had `FixedArrayType` then that would just be an alias for `ArrayType`?

cptartur (commented):
Responded to a comment, please re-request

starknet_py/tests/e2e/fixtures/abi_structures.py
franciszekjob: In the code at [that line](https://github.com/software-mansion/starknet.py/pull/1625/files#diff-cecd57a23ab6762ad8aa0f1da84493bc40c10af6f34428e44a13ee54c9c7e4beR105), we forgot to account for `FixedSizeArrayType` when resolving serializers. Because `FixedSizeArrayType` is a distinct ABI type, we need to extend the existing array‑type branch (`if` statement) to handle it explicitly. This change isn’t user‑facing or smth - it simply broadens our internal type check.

cptartur (approved):

PR #1627, opened 2025-07-21
Default tip calculation

cptartur:
Closes #

## Introduced changes

- 
- 

##


starknet_py/net/account/account.py
```diff
@@ -398,9 +402,12 @@ async def sign_declare_v3(
         nonce: Optional[int] = None,
         resource_bounds: Optional[ResourceBoundsMapping] = None,
         auto_estimate: bool = False,
-        tip: int = 0,
+        tip: Optional[int] = None,
     ) -> DeclareV3:
         # pylint: disable=too-many-arguments
+        if tip is None:
+            tip = await get_tips_median(self.client, block_number="latest")
```
franciszekjob: I'm not sure whether we want to use suggested tip by default. Wouldn't it make more sense if the default value of tip is `0` and someone can pass a flag (e.g. `auto_estimate_tip`) which enforces default estimation? Also, manual tip and mentioned flag would be mutually exclusive. It's similar to how we handle resource bounds, but here the default is 0.

cptartur: Yeah, maybe that makes more sense, will change that

starknet_py/contract.py
```diff
@@ -212,6 +213,7 @@ async def deploy_v3(
         :param resource_bounds: Resource limits (L1 and L2) used when executing this transaction.
         :param auto_estimate: Use automatic fee estimation (not recommended, as it may lead to high costs).
         :param tip: The tip amount to be added to the transaction fee.
+        :param auto_estimate_tip: Use automatic fee estimation. Using this option may lead to higher costs.
```
franciszekjob, line 216:
```suggestion
        :param auto_estimate_tip: Use automatic tip estimation. Using this option may lead to higher costs.
```

starknet_py/contract.py
```diff
@@ -392,6 +396,7 @@ async def invoke(
         :param auto_estimate: Use automatic fee estimation (not recommended, as it may lead to high costs).
         :param nonce: Nonce of the transaction.
         :param tip: The tip amount to be added to the transaction fee.
+        :param auto_estimate_tip: Use automatic fee estimation. Using this option may lead to higher costs.
```
franciszekjob: Please replace it in other occurrences as well:
```suggestion
        :param auto_estimate_tip: Use automatic tip estimation. Using this option may lead to higher costs.
```

starknet_py/tests/e2e/mock/contracts_v2/src/lib.cairo
franciszekjob, line 1: Why do we need these contracts, can't we use the already existing one?

starknet_py/net/tip/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    if block_hash is None and block_number is None:
+        block_hash = "pre_confirmed"
+
+    block_with_txs = await client.get_block_with_txs(block_hash, block_number)
+
+    tips = []
+    for tx in block_with_txs.transactions:
+        if isinstance(tx, TransactionV3):
+            tips.append(tx.tip)
```
franciszekjob:
```suggestion
    tips = [
        tx.tip for tx in block_with_txs.transactions if isinstance(tx, TransactionV3)
    ]
```

starknet_py/contract.py
cptartur: Ah right, will fix that

starknet_py/net/tip/__init__.py
```diff
@@ -0,0 +1,36 @@
+import math
+import statistics
+from typing import Optional, Union
+
+from starknet_py.net.client import Client
+from starknet_py.net.client_models import Hash, Tag, TransactionV3
+
+
+async def get_tips_median(
```
franciszekjob: Actually, I think it's better if it's named `estimate_tip` or `get_estimated_tip`. Median is related with internal logic of the calculation, so it shouldn't be directly included in the function name. Moreover, look, that if the logic potentially changes in the future, this will require a breaking change.

```suggestion
async def estimate_tip(
```

starknet_py/tests/e2e/mock/contracts_v2/src/lib.cairo
cptartur: We declare these contracts in tests, if we used the same one, I get the contract already declared error. We could build an utility similar to what we have in `sncast` so it's using a salt in contract code but it seemed easier to just copy it.

starknet_py/net/tip/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    client: Client,
+    block_hash: Optional[Union[Hash, Tag]] = None,
+    block_number: Optional[Union[int, Tag]] = None,
+) -> int:
+    """
+    Get a median of tips for the provided block.
+    If no block is provided, the `pre-confirmed` block is used.
```
franciszekjob:
```suggestion
    Estimates the transaction tip by taking the median of all V3 transaction tips in the specified block.
    If no block is provided, the `pre‑confirmed` block is used.
```

docs/migration_guide.rst
```diff
@@ -1,6 +1,25 @@
 Migration guide
 ===============
 
+***************************
+0.28.0-rc.1 Migration guide
+***************************
+
+Version 0.28.0-rc.1 of **starknet.py** adds support for estimating tips and adds related utilites,
```
franciszekjob:
```suggestion
Version 0.28.0-rc.1 of **starknet.py** adds support for estimating tips and adds related utilites.
```

docs/migration_guide.rst
```diff
@@ -1,6 +1,25 @@
...
+
+``starknet_py.net.tip``
+-----------------------
+
+.. py:currentmodule:: starknet_py.net.tip
+
+1. Added :func:`get_tips_median` for estimating the median of tips from the selected block.
```
franciszekjob:
```suggestion
1. Added :func:`estimate_tip` for automatic transaction tip estimation.
```

docs/migration_guide.rst
```diff
@@ -1,6 +1,25 @@
...
+
+``starknet_py.net.account.account`` and ``starknet_py.contract``
+----------------------------------------------------------------
+
+1. Added ``auto_estimate_tip`` argument to :class:`starknet_py.net.account.account.Account` and :class:`starknet_py.contract.Contract` methods that accept a ``tip`` argument.
```
franciszekjob: Argument is the actual value, params are in the function definition 😅 :
```suggestion
1. Added ``auto_estimate_tip`` param to :class:`starknet_py.net.account.account.Account` and :class:`starknet_py.contract.Contract` methods that accept a ``tip`` argument.
```

docs/migration_guide.rst
```diff
@@ -1,6 +1,25 @@
...
+
+``starknet_py.net.tip``
+-----------------------
+
+.. py:currentmodule:: starknet_py.net.tip
+
+1. Added :func:`get_tips_median` for estimating the median of tips from the selected block.
+
+``starknet_py.net.account.account`` and ``starknet_py.contract``
+----------------------------------------------------------------
```
franciszekjob: No need to add it

docs/migration_guide.rst
franciszekjob: Also, these links to classes shouldn't include whole module path (now they do), pls change it 🙏 
image

starknet_py/tests/e2e/mock/contracts_v2/src/lib.cairo
franciszekjob: If it takes not much time I would opt do build smth like that, so we don't end up with a clutter in test-related data.

docs/migration_guide.rst
cptartur: Changed to "Tip Support"

docs/migration_guide.rst
```diff
@@ -1,6 +1,21 @@
 Migration guide
 ===============
 
+***************************
+0.28.0-rc.1 Migration guide
+***************************
+
+Version 0.28.0-rc.1 of **starknet.py** adds support for estimating tips and adds related utilites.
```
franciszekjob:
```suggestion
Version 0.28.0-rc.1 of **starknet.py** comes with support for automatic transaction tip estimation.
```

franciszekjob (approved):
There are some errors while building docs (duplicated title and no blank space), pls fix them. Apart from that, LGTM.

image

starknet_py/tests/e2e/mock/contracts_v2/src/lib.cairo
cptartur: If it ever comes up again I'd be for building that, but for single usage the effort doesn't seem worth it.

PR #1635, opened 2025-07-30
Update default token in `Account.get_balance`

franciszekjob:
Closes #

## Introduced changes

- 
- 

##


starknet_py/tests/e2e/fixtures/constants.py
```diff
@@ -42,10 +42,6 @@ def _get_env_lambda(env_name):
 )
 DEVNET_PRE_DEPLOYED_ACCOUNT_PRIVATE_KEY = "0xc10662b7b247c7cecf7e8a30726cff12"
 
-STRK_FEE_CONTRACT_ADDRESS = (
-    "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d"
-)
-
```
franciszekjob, line 48: note: Better to hold it in the same place as ETH address

starknet_py/net/account/account.py
```diff
@@ -313,7 +313,7 @@ async def get_balance(
         if token_address is None:
             chain_id = await self._get_chain_id()
             if chain_id in RECOGNIZED_CHAIN_IDS:
-                token_address = FEE_CONTRACT_ADDRESS
+                token_address = ETH_FEE_CONTRACT_ADDRESS
```
ddoktorski: It should be `STRK_FEE_CONTRACT_ADDRESS` now

PR #1636, opened 2025-07-30
Support eth signer

### LINKED ISSUE #1626: Support Eth Signer
### Feature Request

OpenZeppelin are maintaining a Starknet account that supports secp256k1 signatures. This is useful for teams who rely on infrastructure that can only generate ETH signatures.

Request to add a similar functionality to starknet.py

Reference from starknet.js:

https://github.com/starknet-io/starknet.js/blob/7870847048880843031639b3c357fbea2c9d177e/src/signer/ethSigner.ts

ddoktorski:
Closes #1626


starknet_py/tests/e2e/fixtures/accounts.py
```diff
@@ -235,3 +237,49 @@ async def argent_account_v040(
...
+
+@pytest_asyncio.fixture(scope="package")
+async def eth_account(
+    eth_fee_contract: Contract,
+    strk_fee_contract: Contract,
+    client,
+    eth_account_class_hash,
+) -> BaseAccount:
+    # private_key = int.from_bytes(os.urandom(32), byteorder="big")
+    private_key = _get_random_private_key_unsafe()
```
franciszekjob: Looks like it can be removed:
```suggestion
    private_key = _get_random_private_key_unsafe()
```

starknet_py/tests/unit/signer/test_eth_signer.py
```diff
@@ -0,0 +1,159 @@
...
+    call = Call(
+        to_addr=int(STRK_FEE_CONTRACT_ADDRESS, 0),
+        selector=get_selector_from_name("transfer"),
+        calldata=[0x12345, 10000, 0],
+    )
+    res = await eth_account.execute_v3(calls=call, resource_bounds=MAX_RESOURCE_BOUNDS)
+
+    receipt = await eth_account.client.get_transaction_receipt(
+        tx_hash=res.transaction_hash
+    )
```
franciszekjob: nit:
```suggestion
    receipt = await eth_account.client.get_transaction_receipt(res.transaction_hash)
```

starknet_py/tests/unit/signer/test_ledger_signer.py
```diff
@@ -269,6 +269,7 @@ async def test_deploy_account_and_transfer(client):
     assert recipient_balance_before + 100 == recipient_balance_after
 
 
+@pytest.mark.skip(reason="TODO(1637)")
```
franciszekjob:
```suggestion
@pytest.mark.skip(reason="TODO(#1637)")
```

starknet_py/tests/unit/signer/test_eth_signer.py
```diff
@@ -0,0 +1,159 @@
...
+    assert signed_message[3] == expected_s[1]
+    assert signed_message[4] == expected_v
+
+
+@pytest.mark.asyncio
+async def test_validate_signature(
+    client, eth_account_class_hash, eth_fee_contract, strk_fee_contract
+):
+    signer = EthSigner(
+        0x525BC68475C0955FAE83869BEEC0996114D4BB27B28B781ED2A20EF23121B8DE,
```
franciszekjob: nit: Can't this be a dummy value? If not, then let's maybe add a comment on why this has a specific value.

franciszekjob (commented):
Looks good. Please also update docs (either in this or separate PR).

starknet_py/net/signer/eth_signer.py
```diff
@@ -0,0 +1,55 @@
...
+        return _serialize_signature(signature)
+
+
+# eth signature consists of u256 values
+def _serialize_signature(signature: Signature) -> List[int]:
+    serializer = Uint256Serializer()
+    return (
+        serializer.serialize(signature.r)
+        + serializer.serialize(signature.s)
+        + [signature.v]
```
cptartur, line 54: Why we don't serialize `v`?

starknet_py/tests/e2e/fixtures/misc.py
```diff
@@ -61,6 +62,17 @@ def typed_data(request) -> TypedDataDict:
     return typed_data
 
 
+@pytest.fixture()
+def load_typed_data(request) -> TypedData:
```
cptartur: Maybe rename it to `loaded_typed_data` or something since it's a fixture.

starknet_py/tests/unit/signer/test_eth_signer.py
```diff
@@ -0,0 +1,159 @@
...
+def test_pub_key(private_key, expected_public_key):
+    eth_signer = EthSigner(
+        private_key,
+        chain_id=StarknetChainId.MAINNET,
+    )
+    assert eth_signer.public_key == expected_public_key
+
+
+@pytest.mark.parametrize(
+    "load_typed_data, expected_r, expected_s, expected_v",
```
cptartur: Where do we have these values from?

starknet_py/tests/unit/signer/test_eth_signer.py
```diff
@@ -0,0 +1,159 @@
...
+        0x525BC68475C0955FAE83869BEEC0996114D4BB27B28B781ED2A20EF23121B8DE,
+        chain_id=StarknetChainId.SEPOLIA,
+    )
+
+    # Manually serialize u512 into felt array
+    serializer = Uint256Serializer()
+    public_key_bytes = signer.public_key.to_bytes(64, byteorder="big")
+    constructor_calldata = serializer.serialize(
+        int.from_bytes(public_key_bytes[:32], byteorder="big")
+    ) + serializer.serialize(int.from_bytes(public_key_bytes[32:], byteorder="big"))
```
cptartur: We do so in two different places, maybe it can be a function or a fixture? 

And why do we need to do so manually? Is there no serializer for u512?

starknet_py/tests/unit/signer/test_eth_signer.py
```diff
@@ -0,0 +1,159 @@
...
+            0x5,
+            0x89B5C415C2F37F5A6D1292EDC15958B2,
+            0xA2EC923770F021309C58B7FA68A6A4CE,
+            0xC4AE697F7D4999C9C6F6C6F13C97EF65,
+            0x1792B1FB77858EE84EF3244807A73A16,
+            0x0,
+        ],
+    )
+    (res,) = await client.call_contract(call=call)
+    assert res == encode_shortstring("VALID")
```
cptartur: Not sure about the point of this test. We are deploying an account and then checking if a precalculated signature passes validation. Why not calculate the signature ourselves?

starknet_py/net/signer/eth_signer.py
ddoktorski: v equals 0 or 1, hence it is not `u256`

starknet_py/tests/unit/signer/test_eth_signer.py
ddoktorski: Calculated using starknet.js

starknet_py/tests/unit/signer/test_eth_signer.py
ddoktorski: > Is there no serializer for u512?

Currently there is no serializer for `u512`

> We do so in two different places, maybe it can be a function or a fixture?

I haven't added any logic for this as we just need to implement the serializer and use it instead

starknet_py/tests/unit/signer/test_eth_signer.py
ddoktorski: I wanted to have an external input that I could be certain was correct and then validate against it. However, thinking about it now it might be better to simply sign a transaction and then validate it, I’ll update this accordingly

starknet_py/tests/unit/signer/test_eth_signer.py
ddoktorski: In the current form it must be this value, but I will modify this test so it won't be relevant

starknet_py/net/signer/eth_signer.py
franciszekjob: Let's add a small comment for that

starknet_py/tests/unit/signer/test_eth_signer.py
franciszekjob: Potentially we can create an issue to add it and link todo.

starknet_py/tests/e2e/fixtures/accounts.py
```diff
@@ -235,3 +237,48 @@ async def argent_account_v040(
...
+        address=address,
+        eth_fee_contract=eth_fee_contract,
+        strk_fee_contract=strk_fee_contract,
+    )
+
+    account = Account(
+        address=address,
+        client=client,
+        signer=signer,
+        chain=StarknetChainId.SEPOLIA.value,
```
franciszekjob: nit: No need to access `.value`
```suggestion
        chain=StarknetChainId.SEPOLIA,
```

```python
print(StarknetChainId.SEPOLIA)
print(StarknetChainId.SEPOLIA.value)
```

gives

```
393402133025997798000961
393402133025997798000961
```

starknet_py/tests/unit/signer/test_eth_signer.py
```diff
@@ -0,0 +1,118 @@
...
+        to_addr=map_contract.address,
+        selector=get_selector_from_name("put"),
+        calldata=[100, 101],
+    )
+
+    signed_tx = await eth_account.sign_invoke_v3(
+        calls=map_call, resource_bounds=MAX_RESOURCE_BOUNDS
+    )
+    calldata = (
+        [signed_tx.calculate_hash(StarknetChainId.SEPOLIA.value)]
```
franciszekjob: Same as above

starknet_py/net/signer/eth_signer.py
ddoktorski: Updated the comment above the `_serialize_signature` function

docs/api/signer.rst
```diff
@@ -55,3 +55,15 @@ LedgerSigningMode
 .. autoclass:: LedgerSigningMode
     :members:
     :member-order: groupwise
+
+------------
+EthSigner
+------------
```
franciszekjob: nit:
```suggestion
----------
EthSigner
----------
```

franciszekjob (approved):

PR #1643, opened 2025-08-12
Make block number not optional in txs

cptartur:
Closes #

## Introduced changes

- Following changes in the RPC 0.9.0 spec

##

### DIFF
--- a/docs/migration_guide.rst
+++ b/docs/migration_guide.rst
@@ -2,6 +2,20 @@ Migration guide
 ===============
 
 ***************************
+0.28.0-rc.4 Migration guide
+***************************
+
+.. currentmodule:: starknet_py.net.client_models
+
+1. Removed fields ``block_number`` and ``block_hash`` from :class:`TransactionReceipt`
+2. Added a dedicated class :class:`TransactionReceiptWithBlockInfo`, a subclass of :class:`TransactionReceipt`, that has non-optional field ``block_number`` and optional field ``block_hash``.
+
+.. currentmodule:: starknet_py.net.client
+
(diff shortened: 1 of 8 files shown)


docs/migration_guide.rst
```diff
@@ -1,6 +1,22 @@
 Migration guide
 ===============
 
+***************************
+0.28.0-rc.4 Migration guide
+***************************
+
+Version 0.28.0-rc.4 of **starknet.py** comes with partial support for RPC 0.9.0 (without the support for changes in the Websockets methods).
```
franciszekjob: Let's reform it so it says the support for RPC API is fully compatible but websockets aren't.

starknet_py/net/schemas/rpc/transactions.py
```diff
@@ -90,6 +89,15 @@ def make_dataclass(self, data, **kwargs) -> TransactionReceipt:
         return TransactionReceipt(**data)
 
 
+class TransactionReceiptWithBlockInfoSchema(TransactionReceiptSchema):
+    block_hash = Felt(data_key="block_hash", load_default=None)
+    block_number = fields.Integer(data_key="block_number", required=True)
```
franciszekjob, line 94: In [spec](https://github.com/starkware-libs/starknet-specs/blob/f0ae2e919c2564ea9028d7564bccdba6f3badcad/api/starknet_api_openrpc.json#L3002) both fields are required.

cptartur: If you check the tag for 0.9.0, it says only `block_number` is https://github.com/starkware-libs/starknet-specs/blob/c2e93098b9c2ca0423b7f4d15b201f52f22d8c36/api/starknet_api_openrpc.json#L3007

PR #1654, opened 2025-09-20
Move revision enum

### LINKED ISSUE #1367: Move `Revision` from `schemas/common.py`
### Feature Request

Currently `Revision` is stored in `schemas/common.py`. There are circular import errors when storing it in `models/typed_data.py`. `Revision` should be moved to either separate file or `models/typed_data.py` after solving circular import errors.

Abeeujah:
Closes #1367 

There were two causes of Circular Import Error,
- `int_from_bytes`: The root cause is a circular dependency involving `starknet_py.common` and other `starknet_py` modules. The traceback states: `cannot import name 'int_from_bytes' from partially initialized module 'starknet_py.common'`.
- `HEX_PREFIX = "0x"`: This creates a circular dependency between `starknet_py.hash.address` and `starknet_py.hash.utils` when running the module `starknet_py/tests/e2e/fixtures/clients.py`.
`CasmClassSchema`, `Felt` ...

## Introduced changes
Fixing this was moving `RevisionField` out of common to a separate module.

##

### DIFF
--- a/starknet_py/hash/outside_execution.py
+++ b/starknet_py/hash/outside_execution.py
@@ -1,6 +1,6 @@
 from starknet_py.constants import OutsideExecutionInterfaceID
 from starknet_py.net.client_models import OutsideExecution
-from starknet_py.net.schemas.common import Revision
+from starknet_py.net.models.typed_data import Revision
 from starknet_py.utils.typed_data import TypedData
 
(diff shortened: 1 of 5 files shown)


Abeeujah:
Hi @franciszekjob trust you're doing great, can you please review this at your earliest convenience?

PR #1657, opened 2025-10-17
Support blake hash

MKowalski8:
## Introduced changes

- implementation of `blake` function to use with new `RPC 0.10` 

##


starknet_py/hash/blake2s.py
MKowalski8, line 1: It is based on [sequencer implementation](https://github.com/starkware-libs/sequencer/blob/main/crates/blake2s/src/lib.rs)

starknet_py/hash/blake2s.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    unpacked_u32s = []
+    for felt in felts:
+        # Convert felt to 32-byte big-endian representation
+        felt_as_be_bytes = felt.to_bytes(32, byteorder="big")
+
+        if felt < SMALL_THRESHOLD:
+            # Small: 2 limbs only, high-32 then low-32 of the last 8 bytes
+            hi = int.from_bytes(felt_as_be_bytes[24:28], byteorder="big")
+            lo = int.from_bytes(felt_as_be_bytes[28:32], byteorder="big")
```
franciszekjob: nit: No benefit from using abbreviations here
```suggestion
            high = int.from_bytes(felt_as_be_bytes[24:28], byteorder="big")
            low = int.from_bytes(felt_as_be_bytes[28:32], byteorder="big")
```

starknet_py/hash/casm_class_hash.py
```diff
@@ -10,40 +8,57 @@
...
+        version_parts = [int(x) for x in rpc_version.split(".")]
+
+        # RPC 0.10.0 and later use Blake2s
+        if version_parts[0] > 0 or (version_parts[0] == 0 and version_parts[1] >= 10):
+            return HashMethod.BLAKE2S
+    except (ValueError, IndexError):
+        # If we can't parse the version, default to Poseidon
+        pass
+
+    return HashMethod.POSEIDON
```
franciszekjob, line 27: Better to split it into two functions:

```python
def _parse_version(version: str) -> Tuple[int, int, int]:
    parts = version.split(".")
    
    if len(parts) != 3:
        raise ValueError("Version must have three parts separated by dots.")
    
    return int(parts[0]), int(parts[1]), int(parts[2])


def get_casm_hash_method_for_rpc_version(rpc_version: str) -> HashMethod:
    major, minor, patch = _parse_version(rpc_version)

    # RPC 0.10.0 and later use Blake2s
    if (major, minor, patch) >= (0, 10, 0):
        return HashMethod.BLAKE2S

    return HashMethod.POSEIDON
```

starknet_py/tests/unit/hash/blake2s_test.py
```diff
@@ -0,0 +1,54 @@
+# pylint: disable=line-too-long
+# fmt: off
```
franciszekjob: I removed these locally and linter doesn't complain - seems we don't need it?

starknet_py/hash/blake2s.py
franciszekjob: Let's link a permalink

starknet_py/hash/blake2s.py
```diff
@@ -0,0 +1,81 @@
...
+    encode_felt252_to_u32s hint does, then hashes the resulting byte stream
+    with Blake2s-256 and returns the 256-bit digest as a 252-bit field element.
+    """
+    # 1) Unpack each Felt into 2 or 8 u32 limbs
+    u32_words = encode_felts_to_u32s(felts)
+
+    # 2) Serialize the u32 limbs into a little-endian byte stream
+    byte_stream = b"".join(word.to_bytes(4, byteorder="little") for word in u32_words)
+
+    # 3) Compute Blake2s-256 over the bytes and pack the result into a Felt
```
franciszekjob:
```suggestion
    # Unpack each Felt into 2 or 8 u32 limbs
    u32_words = encode_felts_to_u32s(felts)

    # Serialize the u32 limbs into a little-endian byte stream
    byte_stream = b"".join(word.to_bytes(4, byteorder="little") for word in u32_words)

    # Compute Blake2s-256 over the bytes and pack the result into a Felt
```

starknet_py/tests/unit/hash/blake2s_test.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+Test the encode_felt252_data_and_calc_blake_hash function
+with the same results as the Cairo v0.14 implementation.
+Reference: https://github.com/starkware-libs/cairo-lang/releases/tag/v0.14.0
```
franciszekjob: For me it makes more sense to simply add a permalink to blockifier and say that the tested values are taken from there.

starknet_py/hash/compiled_class_hash_objects.py
```diff
@@ -62,14 +65,19 @@ class BytecodeSegmentedNode(BytecodeSegmentStructure):
...
         return (
-            poseidon_hash_many(
-                itertools.chain(  # pyright: ignore
-                    *[
-                        (node.segment_length, node.inner_structure.hash())
-                        for node in self.segments
-                    ]
+            hash_method.hash_many(
+                list(
+                    itertools.chain(  # pyright: ignore
```
franciszekjob: Is this directive needed?

starknet_py/hash/blake2s.py
franciszekjob, line 1: nit: The function docstrings could be adjusted to follow more Pythonic conventions, including specifying input and return types.

starknet_py/hash/casm_class_hash.py
franciszekjob: Also, I removed `except` branch - endpoint with spec version return semver-compatible version so we should always have major, minor and patch parts.

starknet_py/tests/unit/hash/blake2s_test.py
MKowalski8: Yeah, it's oversight taken from `sequencer` repo

starknet_py/tests/unit/hash/blake2s_test.py
MKowalski8: My bad

starknet_py/hash/compiled_class_hash_objects.py
MKowalski8: Do you mean the list conversion?
I added it because previously it was handled inside the `poseidon` hash function, where there was an explicit `values = list(array)` to convert an iterator into a list.
I think it’s cleaner to handle that here instead, since all these hash functions take a `List` as a parameter, it doesn’t always have to be an iterator

franciszekjob: No no, I mean pyright directive.

MKowalski8: I mean, it was there before, so I assumed that it is necessary

MKowalski8: I'll check it better

MKowalski8: Okay, good pick, with `list` it is not needed

starknet_py/hash/blake2s.py
```diff
@@ -0,0 +1,105 @@
+"""
+This module's Blake2s felt encoding and hashing logic is based on StarkWare's
+sequencer implementation:
+https://github.com/starkware-libs/sequencer/blob/main/crates/blake2s/src/lib.rs
```
franciszekjob: Let's use [parmalink](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files) 😄

starknet_py/hash/casm_class_hash.py
```diff
@@ -10,40 +8,62 @@
...
+
+
+def _parse_version(version: str) -> Tuple[int, int, int]:
+    parts = version.split(".")
+
+    if len(parts) != 3:
+        raise ValueError("Version must have three parts separated by dots.")
+
+    return int(parts[0]), int(parts[1]), int(parts[2])
```
franciszekjob: Actually, let's get rid of this logic from our codebase and use https://pypi.org/project/semver/

starknet_py/contract.py
```diff
@@ -721,8 +722,11 @@ async def declare_v3(
         :return: DeclareResult instance.
         """
 
+        rpc_version = await account.client.spec_version()
+        hash_method = get_casm_hash_method_for_rpc_version(rpc_version)
+
         compiled_class_hash = _extract_compiled_class_hash(
-            compiled_contract_casm, compiled_class_hash
+            compiled_contract_casm, compiled_class_hash, hash_method=hash_method
```
franciszekjob, line 729: This generates extra call with every declare. We can reduce this overhead and store RPC version inside `init` of `FullNodeClient`, thanks to such change we will make only 1 request 😅  .

MKowalski8: Okay, I tried to do that, you can check if it is what you expected.

starknet_py/hash/casm_class_hash.py
```diff
@@ -1,6 +1,6 @@
 from typing import List, Optional, Sequence, Tuple
 
-from poseidon_py.poseidon_hash import poseidon_hash_many
+import semver
```
franciszekjob: nit:
```suggestion
from semver import Version
```

starknet_py/hash/casm_class_hash.py
```diff
@@ -10,40 +10,53 @@
...
 CASM_CLASS_VERSION = "COMPILED_CLASS_V1"
 
 
-def compute_casm_class_hash(casm_contract_class: CasmClass) -> int:
+def get_casm_hash_method_for_rpc_version(rpc_version: str) -> HashMethod:
+    # RPC 0.10.0 and later use Blake2s
+
+    version = semver.Version.parse(rpc_version)
+    if version >= semver.Version.parse("0.10.0"):
+        return HashMethod.BLAKE2S
```
franciszekjob, line 25: nit:
```suggestion
def get_casm_hash_method_for_rpc_version(rpc_version: str) -> HashMethod:
    # RPC 0.10.0 and later use Blake2s
    version = semver.Version.parse(rpc_version)
    
    if version >= semver.Version.parse("0.10.0"):
        return HashMethod.BLAKE2S
```

starknet_py/net/full_node_client.py
```diff
@@ -772,11 +773,15 @@ async def spec_version(self) -> str:
 
         :return: String with version of the Starknet JSON-RPC specification.
         """
-        res = await self._client.call(
-            method_name="specVersion",
-            params={},
-        )
-        return res
+        if not self._spec_version:
```
franciszekjob:
```suggestion
        if self._spec_version is None:
```

starknet_py/tests/unit/hash/blake2s_test.py
```diff
@@ -0,0 +1,52 @@
+# fmt: off
```
franciszekjob: It seems to me we don't need it, check it pls.

starknet_py/tests/unit/hash/casm_class_hash_test.py
```diff
@@ -1,8 +1,13 @@
+# pylint: disable=line-too-long
```
franciszekjob: nit: We can move this to specific line, like:

```python
my_long_function_name.......(...) # pylint: disable=line-too-long
```

starknet_py/contract_utils.py
```diff
@@ -2,13 +2,15 @@
...
 from starknet_py.hash.casm_class_hash import compute_casm_class_hash
+from starknet_py.hash.hash_method import HashMethod
 from starknet_py.net.account.base_account import BaseAccount
 from starknet_py.net.client import Client
 
 
 def _extract_compiled_class_hash(
     compiled_contract_casm: Optional[str] = None,
     compiled_class_hash: Optional[int] = None,
+    hash_method: HashMethod = HashMethod.POSEIDON,
```
franciszekjob: I think we should use blake as default, in order to promote newer version of the network.

poetry.lock
franciszekjob, line 1: Not sure whether all of these changes in lock are related 🤔  + version of poetry has changed from 2.0.0 to 2.1.3

Let's install poetry 2.0.0, with

```shell
curl -sSL https://install.python-poetry.org | python3 - --version 2.0.0
```

then check

```shell
poetry --version
```

and finally run

```shell
poetry lock
```

after that we should have minimal changes, needed only for new `semver` package.

franciszekjob (approved):
Please address mentioned changes, apart from that LGTM

poetry.lock
MKowalski8: Okay, I did not know the newer version is a problem. I downgraded the version

MKowalski8: There are less changes, but there still are. Does this seem odd to you now?

franciszekjob: Discussed offline, resolving

franciszekjob:
Let's also mention changes in migration guide

PR #1662, opened 2025-11-12
Support python >=3.10 to <4

### LINKED ISSUE #1539: Add support for Python 3.13
### Feature Request

Python 3.13 is the latest stable Python version, but currently, the lib cannot be installed with Python 3.13.

Requestion to add support

cptartur:
Closes #1539

## Introduced changes

Support python 3.10-3.14

##

- Remove support for python 3.9

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -76,7 +76,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ "3.10", "3.14" ]
+        python-version: [ "3.14" ]
     steps:
 
       # ====================== SETUP ====================== #
@@ -85,7 +85,7 @@ jobs:
       - uses: asdf-vm/actions/setup@v3
       - uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.14"
           cache: 'pip'
 
       - name: Install poetry


Arcticae:
Clue for fixing this:
https://github.com/python-pillow/Pillow/issues/5042
(the issue seems to be building from source)

.github/workflows/checks.yml
```diff
@@ -175,6 +179,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: 'poetry'
 
+      - name: Install dependencies needed for HIDAPI and Pillow
+        run: |
+          sudo apt install python3-dev libusb-1.0-0-dev libudev-dev libjpeg-dev
+
```
cptartur, line 185: What about users, do they need to install these as well? Or it's only for building?

franciszekjob:
> Or do we want to keep 3.14 as max in pyproject toml?

Testing up to 3.14 doesn’t mean we should cap the package at 3.14, let's leave  `requires-python = ">=3.10, <4"`

.github/workflows/checks.yml
franciszekjob: Good catch, is's indeed needed for linux users. Added in https://github.com/software-mansion/starknet.py/pull/1662/commits/5c09bfa0e91982ec14afe531f11d8860be8b72e6

docs/api/signer.rst
```diff
@@ -36,12 +36,20 @@ KeyPair
...
 
     poetry add starknet_py[ledger]
 
+Under a Debian or Ubuntu based system, you will need to install additional packages:
+
+.. code-block:: bash
+
+    sudo apt install python3-dev libusb-1.0-0-dev libudev-dev
+
+They are needed for compiling HIDAPI. Read official `ledgerctl installation guide <https://github.com/LedgerHQ/ledgerctl?tab=readme-ov-file#quick-install)>`_ for more details.
```
cptartur, line 51: I'd consider including these additions in migration guide and changelog.

cptartur (commented):
Can't approve but there you go ✅

PR #1665, opened 2025-11-19
Fix parsing signed ints

### LINKED ISSUE #1664: [BUG] abi parser fails on contract with `i64`
### What happened

V2 type parser fails when parsing abi with `i64`

### Stack trace

```
starknet_py.cairo.v2.type_parser.UnknownCairoTypeError: Type 'core::integer::i64' is not defined.
```

### Steps to reproduce

1. Declare and deploy a contract which include `core::integer::i64` in its code, e.g.
```rust
#[starknet::interface]
pub trait IHelloStarknet<TContractState> {
    fn foo(self: @TContractState, x: i64) -> felt252;
}

#[starknet::contract]
mod HelloStarknet {
    #[storage]
    struct Storage {
    }

    #[abi(embed_v0)]
    impl HelloStarknetImpl of super::IHelloStarknet<ContractState> {
        fn foo(self: @ContractState, x: i64) -> felt252 {
            let res = x * 2;
            return res.into();
        }
    }
}
```

2. Run following script
```python
from starknet_py.proxy.contract_abi_resolver import ContractAbiResolver, ProxyConfig


async def main():
    client = ...
    account = ...
    contract_address = ...
    abi, cairo_version = await ContractAbiResolver(
        address=contract_address,
        client=client,
        proxy_config=ProxyConfig(),
    ).resolve()

    deployer_contract = Contract(
        address=contract_address,
        abi=abi,
        provider=account,
        cairo_version=cairo_version,
    )


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

### SDK Version

0.28.0

### Python version

3.12

Mac

- [ ] I have searched the existing issues and verified no issue exits for this problem.

franciszekjob:
Closes #1664

## Introduced changes

Fix parsing signed integers in ABI

##

### DIFF
--- a/starknet_py/tests/unit/serialization/data_serializers/int_serializer_test.py
+++ b/starknet_py/tests/unit/serialization/data_serializers/int_serializer_test.py
@@ -35,10 +35,6 @@ def _felt(x: int) -> int:
     ],
 )
 def test_valid_values(value, serializer, serialized_value):
-    print("-------")
-    print("value:", value)
-    print("serialized value:", serialized_value)
-    print("-------")
     deserialized = serializer.deserialize(serialized_value)
     assert deserialized == value
 


starknet_py/serialization/data_serializers/int_serializer.py
```diff
@@ -0,0 +1,67 @@
...
+        with context.push_entity("int" + str(self.bits)):
+            self._ensure_valid_int(
+                raw if raw < (1 << (self.bits - 1)) else raw - FIELD_PRIME,
+                context,
+                self.bits,
+            )
+
+            # unsigned → signed
+            if raw >= (1 << (self.bits - 1)):
+                raw = raw - FIELD_PRIME
```
cptartur: I'd swap the order of this, first do conversions of `raw` and then call `_ensure_valid_int` with it.

cptartur: If -x is serialized as `P - x`, shouldn't getting back the value be `FIELD_PRIME - raw`?

starknet_py/serialization/data_serializers/int_serializer.py
```diff
@@ -0,0 +1,67 @@
...
+        context.ensure_valid_type(value, isinstance(value, int), "int")
+        yield from self._serialize_from_int(value, context, self.bits)
+
+    @staticmethod
+    def _serialize_from_int(
+        value: int, context: SerializationContext, bits: int
+    ) -> Generator[int, None, None]:
+        IntSerializer._ensure_valid_int(value, context, bits)
+
+        unsigned = value % FIELD_PRIME
```
cptartur, line 51: Shouldn't we be serializing it as `FIELD_PRIME - x` as docs say?

cptartur: Or is it the same?

franciszekjob: Considering, that `value` is in range of `(FIELD_PRIME;FIELD_PRIME)` - then yes it's the same.
We could change it to smth like:
```python
unsigned = FIELD_PRIME + value if value < 0 else value
```

then it looks more like in the docs.

starknet_py/serialization/data_serializers/int_serializer.py
franciszekjob: Not really, it's just a reverse process to serialization. As we know, if value is negative (when serializing), we convert it to felt as `FIELD_PRIME + value` (because `FIELD_PRIME - (-value)`. So knowing that raw value exceeds the boundary (`1 << (self.bits - 1)`), we have to subtract `FIELD_PRIME`.

However, I can slightly refactor it to smth as follows, so it's more readable:
```python
with context.push_entity("int" + str(self.bits)):
        signed_threshold = 1 << (self.bits - 1)
        deserialized_val = raw if raw < signed_threshold else raw - FIELD_PRIME
        self._ensure_valid_int(
            deserialized_val,
            context,
            self.bits,
        )

        return deserialized_val
```

wdyt?

franciszekjob: Good point, reordered.

starknet_py/serialization/data_serializers/int_serializer.py
cptartur: If that's the same, I think both options are okay so up to you.

starknet_py/tests/unit/serialization/data_serializers/int_serializer_test.py
```diff
@@ -0,0 +1,72 @@
...
+        # boundaries
+        (MIN_I128, u128_serializer, [_felt(MIN_I128)]),
+        (MAX_I128, u128_serializer, [_felt(MAX_I128)]),
+    ],
+)
+def test_valid_values(value, serializer, serialized_value):
+    print("-------")
+    print("value:", value)
+    print("serialized value:", serialized_value)
+    print("-------")
```
cptartur:
```suggestion
```

PR #1668, opened 2025-11-20
Fix checking circular imports

franciszekjob:
Closes #

## Introduced changes

Fix checking circular imports and add test case in which we actually have them

##

### DIFF
--- a/circular.py
+++ b/circular.py
@@ -1,8 +1,11 @@
 import importlib.util
 import os
+import shutil
 import sys
 
 import pytest
+
+PACKAGE_NAME = "starknet_py"
 
 
 def _import_from_path(module_name, file_path):
@@ -12,12 +15,12 @@ def _import_from_path(module_name, file_path):
     spec.loader.exec_module(module)
 
 
-def assert_no_circular_imports(package_name: str):
-    for path, _, files in os.walk(package_name):
+def assert_no_circular_imports():
+    for path, _, files in os.walk(PACKAGE_NAME):
         py_files = [f for f in files if f.endswith(".py")]
         for file in py_files:
             file_path = os.path.join(path, file)
-            relative_path = os.path.relpath(file_path, package_name)
+            relative_path = os.path.relpath(file_path, PACKAGE_NAME)
             module_path_no_ext = relative_path.removesuffix(".py")
 
             # Handle __init__.py files specially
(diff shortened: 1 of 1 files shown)


circular.py
```diff
@@ -10,9 +15,79 @@ def _import_from_path(module_name, file_path):
...
+
+            relative_path = os.path.relpath(file_path, PACKAGE_NAME)
+            module_path_no_ext = relative_path.removesuffix(".py")
+            # Handle __init__.py files specially
+            if module_path_no_ext.endswith("__init__"):
+                module_path_no_init = module_path_no_ext.removesuffix(
+                    "__init__"
+                ).rstrip(os.sep)
+                if not module_path_no_init:  # Top-level __init__.py
+                    module_name = PACKAGE_NAME
```
cptartur: Nit, but I don't like these non explicit checks like this. Isn't explicitly checking for len (or what's intended here) more readable?

circular.py
```diff
@@ -10,9 +15,79 @@ def _import_from_path(module_name, file_path):
...
+                else:
+                    dotted_module_path = module_path_no_init.replace(os.sep, ".")
```
cptartur, line 36: Out of curiosity, what possible case there is that the `os.sep` is different than dots?

cptartur: We can keep it anyways.

circular.py
```diff
@@ -10,9 +15,79 @@ def _import_from_path(module_name, file_path):
...
+        )
+        with pytest.raises(ImportError, match=error_regex):
+            assert_no_circular_imports()
+    finally:
+        # Clean up temporary files
+        if os.path.exists(module_path):
+            shutil.rmtree(module_path)
+        sys.modules.pop(f"{PACKAGE_NAME}.{module_name}.file_a", None)
+        sys.modules.pop(f"{PACKAGE_NAME}.{module_name}.file_b", None)
+        sys.modules.pop(f"{PACKAGE_NAME}.{module_name}", None)
```
cptartur, line 74: Wouldn't it be easier to maintain if we created the two packages you are using for two tests below and kept them somewhere in the repo? And run tests on them? Instead of maintaining this logic.

circular.py
```diff
@@ -10,9 +15,79 @@ def _import_from_path(module_name, file_path):
     spec.loader.exec_module(module)
 
 
-def test_circular_imports():
-    for path, _, files in os.walk("starknet_py"):
+def assert_no_circular_imports():
```
cptartur, line 18: So to make sure I understand. This iterates over every file in the package and tries importing from it, checking if there are no errors (so no circular imports)?

franciszekjob: Yup

circular.py
franciszekjob: Purpose of this test is to detect circular imports within the package itself - placing the files somewhere else would no longer check the real import graph. Also, I think it's better to generate these intentionally-invalid “polluting” modules dynamically so they remain fully isolated, temporary, and do not clutter the repository.

circular.py
franciszekjob: > Out of curiosity, what possible case there is that the os.sep is different than dots?

os.sep won't be dots here. Look at `module_path_no_init` - values are as follows:

image

as you can see we have os separators in the path, so that's why we replace them with `.` so the import 

Buut, it turns out we can theoretically get rid of this replacement so module names with mix of `.` and os separators will work - e.g. `starknet_py.abi/v2/parser`) will be valid. However, I would still leave it so the import paths are pythonic.

circular.py
franciszekjob: Using the truthiness check is the standard Python idiom. However, I see it may be a bit confusing here so I updated the comment to smth like:
```
# Top-level __init__.py gives empty module path
```

should be clearer now.

circular.py
cptartur: I mean, you can have a directory where you store these mock packages instead of creating them like that. I didn't mean for them to be put in the starknet-py package directly.
It's easier to maintain these than code that generates them in memory.

cptartur (approved):
Pick whatever approach works better for you with creating test dirs. I prefer to have mocks in the directory but I won't die on that hill.

circular.py
franciszekjob: We can't have these files in other package than starknet_py because we cannot have multiple top-level packages in a flat layout, e.g. https://app.readthedocs.org/projects/starknetpy/builds/30467365/

PR #1680, opened 2025-11-27
fix NumberAsHex to support negative hex values

feltroidprime:


franciszekjob:
Closing as this is a bug in compiler, the SDK implementation follows spec.

Issue #1681, opened 2025-12-05
Run network tests on multiple RPC versions


franciszekjob:
In the scope of this task we should extend our network tests to be run on a few RPC versions (i.e. supported RPC versions range). This should prevent from potential regressions (e.g. changes the network, which affect RPC compatibility).

PR #1689, opened 2026-01-07
Extend network tests

franciszekjob:
**Stack**
- #1697
- #1692
- #1689 ⬅️ 
- #1701
- #1695
- #1694

Towards #1681

## Introduced changes

Extend network tests with account tests

##


starknet_py/tests/e2e/account/account_test.py
```diff
@@ -81,24 +81,6 @@ async def test_estimated_fee_greater_than_zero(account, erc20_contract):
...
-        resource_bounds=MAX_RESOURCE_BOUNDS,
-    )
-
-    estimated_fee = await account.client.estimate_fee(tx=declare_tx)
-
-    assert isinstance(estimated_fee.overall_fee, int)
-    assert estimated_fee.overall_fee > 0
-    assert estimated_fee.calculate_overall_fee() == estimated_fee.overall_fee
-
-
```
franciszekjob, line 101: note: Duplicate of `test_account_estimate_fee_for_declare_transaction`

.github/workflows/checks.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
       - name: Install dependencies needed for HIDAPI and Pillow
         run: |
-          sudo apt install python3-dev libusb-1.0-0-dev libudev-dev libjpeg-dev
+          sudo apt-get update && sudo apt install --fix-missing python3-dev libusb-1.0-0-dev libudev-dev libjpeg-dev
```
franciszekjob: note: I experienced that this step fails indeterministically, hence this change.

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    contract = await Contract.from_address(
+        provider=account_sepolia_testnet, address=STRK_FEE_CONTRACT_ADDRESS
+    )
+
+    # Prefund account with 5 * deployment fee
+    invocation = await contract.functions["transfer"].invoke_v3(
+        address, deploy_account_fee.overall_fee * 5, auto_estimate=True
+    )
```
franciszekjob: note: 5 is arbitrary, but wanted to have a safe margin for the deployment fee.

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -22,6 +22,38 @@ setup_scarb() {
     fi
 }
 
+apply_contract_salt() {
+    SALT="$1"
+
+    echo "Updating salted contracts with salt: ${SALT}"
+
+    shopt -s nullglob
```
franciszekjob: note: nullglob is enabled to avoid issues if no files match

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -33,6 +65,11 @@ compile_contracts_with_scarb() {
     echo "Checking Cairo contracts formatting"
     scarb fmt --check
 
+    SALT=$(uuidgen | tr -d '-')
+    trap 'cd "$CONTRACTS_DIRECTORY" && revert_contract_salt "$SALT"' EXIT
```
franciszekjob, line 69: note: trap is better than imperatively calling `revert_contract_salt`, as it will be executed both when compilation succeeds or fails.

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob, line 1: note: Sleeps are added to lower risk of tests failure due to old nonce, tests like to fails sometimes because of it. If you see a better way I'm happy to change it.

starknet_py/tests/e2e/mock/contracts_salted/.tool-versions
```diff
@@ -0,0 +1 @@
+scarb 2.10.0
```
ddoktorski: Why this Scarb version?

starknet_py/tests/e2e/mock/contracts_salted/Scarb.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "contracts_salted"
+version = "0.1.0"
+
+[dependencies]
+starknet = "2.10.0"
+
+[[target.starknet-contract]]
+casm = true
+sierra = true
+casm-add-pythonic-hints = true
```
ddoktorski: Why is it added here?

starknet_py/tests/e2e/mock/compile_contracts.sh
ddoktorski: Please add a comment in the code what it does

starknet_py/tests/e2e/mock/compile_contracts.sh
ddoktorski: Please add a comment in the code what it does

starknet_py/tests/e2e/tests_on_networks/account_test.py
ddoktorski, line 1: Are we sure we want to run this test suite on every CI run?

starknet_py/tests/e2e/mock/contracts_salted/.tool-versions
franciszekjob: Bumped to 2.15.1, previously take the same as in other package.

starknet_py/tests/e2e/mock/contracts_salted/Scarb.toml
franciszekjob: Removed

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: We can surely run them only in the compatibility suite, but I’m wondering whether keeping these 4 tests in the regular CI does more good than harm - especially since we don't have any basic network-backed coverage for these methods (on regular run). WDYT?

ddoktorski: Running interactions with the network hundreds of times with each change doesn’t seem ideal, maybe only run them on merge?

franciszekjob: Ok, makes sens. Changed in https://github.com/software-mansion/starknet.py/pull/1689/changes/56be486c310c1587f368e2320b39e60db24caf5e

starknet_py/tests/e2e/mock/contracts_salted/.tool-versions
ddoktorski, line 1: Can we have only one `.tool-versions` file in the `tests` directory?

.github/workflows/checks.yml
ddoktorski: I have no context of the root problem, but could this fix create new issues if some dependencies aren't installed, making new errors harder to understand? Wouldn't it be better to add a retry mechanism for the entire command in case it fails?

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+from starknet_py.net.udc_deployer.deployer import Deployer, _get_random_salt
+from starknet_py.tests.e2e.fixtures.misc import load_contract
+from starknet_py.tests.e2e.utils import _new_address
+
+
+@pytest.mark.network_account_test
+@pytest.mark.asyncio
+async def test_execute_v3(account_sepolia_testnet):
+    sleep(10)
+    # https://sepolia.starkscan.co/contract/0x0589A8B8BF819B7820CB699EA1F6C409BC012C9B9160106DDC3DACD6A89653CF
```
ddoktorski: starkscan is gone
```suggestion
```

pyproject.toml
```diff
@@ -79,6 +79,7 @@ test_ci_v1 = "coverage run -a -m pytest -n auto --contract_dir=v1 starknet_py --
 test_ci_v2 = "coverage run -a -m pytest -n auto --contract_dir=v2 starknet_py --ignore=starknet_py/tests/e2e/docs --ignore=starknet_py/tests/e2e/tests_on_networks"
 
 test_ci_on_networks = "coverage run -a -m pytest --contract_dir=v2 starknet_py/tests/e2e/tests_on_networks"
+test_ci_on_networks_extended = "coverage run -a -m pytest --contract_dir=v2 starknet_py/tests/e2e/tests_on_networks"
```
ddoktorski, line 82: It's identical as `test_ci_on_networks`

starknet_py/tests/e2e/mock/compile_contracts.sh
```diff
@@ -20,6 +20,38 @@ setup_scarb() {
...
+apply_contract_salt() {
+    SALT="$1"
+
+    echo "Updating salted contracts with salt: ${SALT}"
+
+    shopt -s nullglob # Make unmatched globs expand to nothing instead of a literal pattern
+
+    for FILE in ./src/*.cairo; do
+        sed -i.bak "s/__salt_placeholder__/${SALT}/g" "$FILE"
+        rm "$FILE".bak 2> /dev/null
```
ddoktorski: nit:
```suggestion
        rm "$FILE.bak" 2> /dev/null
```

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+from starknet_py.net.signer.key_pair import KeyPair
+from starknet_py.net.udc_deployer.deployer import Deployer, _get_random_salt
+from starknet_py.tests.e2e.fixtures.misc import load_contract
+from starknet_py.tests.e2e.utils import _new_address
+
+
+@pytest.mark.network_account_test
+@pytest.mark.asyncio
+async def test_execute_v3(account_sepolia_testnet):
+    sleep(10)
```
ddoktorski: comment explaining this might be helpful for the future

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+        compiled_class_hash,
+        auto_estimate=True,
+    )
+
+    declare_response = await account_sepolia_testnet.client.declare(
+        transaction=signed_tx
+    )
+    await account_sepolia_testnet.client.wait_for_tx(declare_response.transaction_hash)
+
+    assert declare_response.class_hash != 0
```
ddoktorski: Can we assert exact value?

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+        class_hash=ARGENT_V040_CLASS_HASH,
+        salt=salt,
+        key_pair=key_pair,
+        client=client_sepolia_testnet,
+        constructor_calldata=constructor_calldata,
+        auto_estimate=True,
+    )
+    await deploy_result.wait_for_acceptance()
+
+    assert isinstance(account_sepolia_testnet, BaseAccount)
```
ddoktorski: It asserts the fixture account, not the deployed one

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+    assert deploy_result.account.address == address
+
+    transaction = await client_sepolia_testnet.get_transaction(
+        tx_hash=deploy_result.hash
+    )
+    assert isinstance(transaction, DeployAccountTransactionV3)
+    assert transaction.constructor_calldata == constructor_calldata
+
+
+@pytest.mark.network_account_test
```
ddoktorski: What is the purpose of the new mark?

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+        contract_address_salt=salt,
+        constructor_calldata=constructor_calldata,
+        auto_estimate=True,
+    )
+    deploy_account_fee = await new_account.estimate_fee(tx=tx)
+    deploy_account_fee = (
+        deploy_account_fee[0]
+        if isinstance(deploy_account_fee, list)
+        else deploy_account_fee
+    )
```
ddoktorski, line 74: Why do we check the type here? I guess it should always be a single value, not the list

starknet_py/tests/e2e/tests_on_networks/account_test.py
ddoktorski: Given that we run it on testnet maybe make it x2 to avoid losing tokens?

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+@pytest.mark.network_account_test
+@pytest.mark.asyncio
+async def test_deploy_v3(account_sepolia_testnet):
+    sleep(10)
+    calldata = []
+    salt = _get_random_salt()
+    deployer = Deployer()
+
+    # https://sepolia.starkscan.co/class/0x0227f52a4d2138816edf8231980d5f9e6e0c8a3deab45b601a1fcee3d4427b02
```
ddoktorski:
```suggestion
```

starknet_py/tests/e2e/tests_on_networks/account_test.py
```diff
@@ -0,0 +1,159 @@
...
+        salt=salt,
+    )
+
+    tx_response = await account_sepolia_testnet.execute_v3(
+        calls=contract_deployment.call, auto_estimate=True
+    )
+    await account_sepolia_testnet.client.wait_for_tx(tx_response.transaction_hash)
+
+    assert isinstance(contract_deployment.address, int)
+    assert contract_deployment.address != 0
```
ddoktorski: Can we improve these asserts?

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: Please take a look at `estimate_fee` method signature, it indeed returns union of single and list value:

https://github.com/software-mansion/starknet.py/blob/26b6519d2abc2b39b3928538354c3452ed246d47/starknet_py/net/account/account.py#L220-L226

starknet_py/tests/e2e/mock/contracts_salted/.tool-versions
franciszekjob: In `contracts_v1` we use `scarb 0.4.1`. I can check if bumping scarb there will cause some problems and whether we can use single file, but let's not add it to this PR.

pyproject.toml
franciszekjob: Changed, now `test_ci_on_networks` has an additional `--ignore=starknet_py/tests/e2e/tests_on_networks/account_test.py`

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: https://github.com/software-mansion/starknet.py/pull/1689/changes/e4abe4392095b6691c92b24f9b17663e4396ddf9

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: It was introduced to mark tests that should be run in extended network suite. But actually, it can be simplified by adding `--ignore=starknet_py/tests/e2e/tests_on_networks/account_test.py` to `test_ci_on_networks`. We won't need marker  then(removed in https://github.com/software-mansion/starknet.py/pull/1689/changes/dce37d4367abf3254e41c95dcdecf0c7c36bf9c9).

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: I ran it few times with x2 and it always failed (e.g. run [here](https://github.com/software-mansion/starknet.py/actions/runs/22860423559/job/66337653709)). Seems that x3 (and more) works, let's stick to 4 to be "safer".

franciszekjob:
I temporarily changed the workflow so it runs account tests as well, it passes - link to run:
https://github.com/software-mansion/starknet.py/actions/runs/22868205186/job/66340763503?pr=1689

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: Removed.

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: Definitely, added in https://github.com/software-mansion/starknet.py/pull/1689/changes/f80fcb192147af8ca237feb54a753eecff665104

.github/workflows/checks.yml
franciszekjob: Turns out these could be removed - at the time when I was doing changes, there was some kind of issue with runner, and these changes fixed it. It's gone now.

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: `declare_response.class_hash` is sierra class hash and `compiled_class_hash` is casm class hash, so we can't compare these values. If you take a look at other declare tests (.e.g [here](https://github.com/software-mansion/starknet.py/blob/2b8d3fdf53e96183bca17b9f8404497d33840c22/starknet_py/tests/e2e/declare/declare_test.py#L9) or [here](https://github.com/software-mansion/starknet.py/blob/2b8d3fdf53e96183bca17b9f8404497d33840c22/starknet_py/tests/e2e/docs/code_examples/test_contract.py#L62), we don't assert specific value, so I think it's fine as it is.

starknet_py/tests/e2e/mock/contracts_salted/.tool-versions
ddoktorski: Actually I think `contracts_salted` should be put in `contracts_v2` directory either in different package or behind a feature flag, then we have one `.tool-versions` for all "new" contracts

starknet_py/tests/e2e/tests_on_networks/account_test.py
ddoktorski: Can it return a list in this test?

starknet_py/tests/e2e/tests_on_networks/account_test.py
ddoktorski: Why not to compute sierra class hash separately and assert it here?

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: No, but I added it for typecheck, it fails without this change

```
  /Users/franciszekjob/Projects/starknet/starknet.py/starknet_py/tests/e2e/tests_on_networks/account_test.py:75:37 - error: Cannot access attribute "overall_fee" for class "List[EstimatedFee]"
    Attribute "overall_fee" is unknown (reportAttributeAccessIssue)
2 errors, 0 warnings, 0 information

starknet_py/tests/e2e/tests_on_networks/account_test.py
franciszekjob: Changed, take a look https://github.com/software-mansion/starknet.py/pull/1689/changes/6b4616121ac75f611cd8c5d4aa4247dcce91b609

starknet_py/tests/e2e/mock/contracts_salted/.tool-versions
franciszekjob: Ok, makes sense, I removed them as separate package.

ddoktorski (approved):

PR #1695, opened 2026-01-11
Refactor contracts utils

franciszekjob:
**Stack**
- #1697
- #1692
- #1689
- #1701
- #1695 ⬅
- #1694

Towards #1681

## Introduced changes

- 
- 

##

### DIFF
--- a/.github/actions/compile-contracts/action.yml
+++ b/.github/actions/compile-contracts/action.yml
@@ -33,4 +33,5 @@ runs:
 
     - name: Compile contracts ${{ inputs.package }}
       if: inputs.cache != 'true' || steps.cache-contracts.outputs.cache-hit != 'true'
+      shell: bash
       run: poetry run poe compile_contracts ${{ inputs.package }}
--- a/.github/actions/compile-contracts/action.yml
+++ b/.github/actions/compile-contracts/action.yml
@@ -33,6 +33,4 @@ runs:
 
     - name: Compile contracts ${{ inputs.package }}
       if: inputs.cache != 'true' || steps.cache-contracts.outputs.cache-hit != 'true'
-      shell: bash
-      run: |
-        poetry run poe compile_contracts ${{ inputs.package }}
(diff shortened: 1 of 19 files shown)


.github/actions/compile-contracts/action.yml
```diff
@@ -0,0 +1,37 @@
...
+    description: 'Name of contracts package'
+    required: true
+  contracts_base_path:
+    description: 'Base path to contracts directory'
+    required: false
+    default: 'starknet_py/tests/e2e/mock'
+  cache:
+    description: 'Whether to cache contracts'
+    required: false
+    default: true
```
ddoktorski, line 15: Are there any cases that we don't want to cache it?

starknet_py/tests/e2e/fixtures/misc.py
```diff
@@ -145,30 +138,20 @@ def block_with_tips_mock():
...
-    V2 = "V2"
-
-
 class UnknownArtifacts(BaseException):
     pass
 
 
-def load_contract(contract_name: str, version: Optional[ContractVersion] = None):
-    if version is None:
+def load_contract(contract_name: str, package: Optional[str] = None):
```
ddoktorski, line 145: nit: Can't we still use an enum and get the str name by calling `.value` below?

.github/actions/compile-contracts/action.yml
franciszekjob: Yes, in the next PR there are salted contracts which we don't want to cache.

starknet_py/tests/e2e/fixtures/misc.py
franciszekjob: Previously we had only two packages and some kind of version aspect - now when we have more packages, imo it doesn't make sense to pass it as enum. Simple package param as string seems most straight forward.

ddoktorski (approved):

PR #1706, opened 2026-03-02
Fix circular import tests

franciszekjob:
Closes #

## Introduced changes

- 
- 

##

### DIFF
--- a/starknet_py/utils/typed_data.py
+++ b/starknet_py/utils/typed_data.py
@@ -52,14 +52,17 @@ class Domain:
     revision: Optional[Revision] = None
 
     def __post_init__(self):
-        raw_value = (
-            self.revision.value if isinstance(self.revision, Enum) else self.revision
+        self.resolved_revision = (
+            Revision(self.revision) if self.revision else Revision.V0
         )
-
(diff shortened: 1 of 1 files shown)


franciszekjob:
note: e.g. run with failure, before the change
https://github.com/software-mansion/starknet.py/actions/runs/22576484289/job/65397237487?pr=1706

starknet_py/utils/typed_data.py
```diff
@@ -52,9 +52,14 @@ class Domain:
...
-        self.resolved_revision = (
-            Revision(self.revision) if self.revision else Revision.V0
+        raw_value = (
+            self.revision.value if isinstance(self.revision, Enum) else self.revision
         )
+
+        if raw_value is None:
+            self.resolved_revision = Revision.V0
+        else:
+            self.resolved_revision = Revision(raw_value)
```
ddoktorski:
```suggestion
    val = self.revision.value if isinstance(self.revision, Enum) else self.revision
    self.resolved_revision = Revision.V0 if val is None else Revision(val)
```

ddoktorski (approved):

PR #1707, opened 2026-03-04
Fix stark curve in `KeyPair.generate()`

### LINKED ISSUE #1704: Fix `KeyPair.generate()`
### Feature Request

ATM [`KeyPair.generate()`](https://github.com/software-mansion/starknet.py/blob/9c45ed62a56f447883925910a5b9e597aefad260/starknet_py/net/signer/key_pair.py#L28) doesn't use a correct value when generating private key, i.e. it uses field prime instead of elliptic curve order.

Reference PR in `starknet-rust`: https://github.com/software-mansion/starknet-rust/pull/98

ddoktorski:
Closes #1704

### DIFF
--- a/docs/api/signer.rst
+++ b/docs/api/signer.rst
@@ -32,6 +32,9 @@ KeyPair
     :undoc-members:
     :member-order: groupwise
 
+.. warning::
+    **Not Audited:** The ``KeyPair.generate()`` function has not been audited for cryptographic security. Use at your own risk.
+
 ------------
 LedgerSigner
 ------------
--- a/docs/guide/generating_key_pair.rst
+++ b/docs/guide/generating_key_pair.rst
@@ -19,6 +19,9 @@ suitable for managing secrets such as account authentication, tokens, and simila
     :start-after: docs-generate: start
     :end-before: docs-generate: end
 
+.. warning::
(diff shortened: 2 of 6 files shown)


franciszekjob (commented):
Let's mention the fix in migration guide too

docs/api/signer.rst
```diff
@@ -32,6 +32,9 @@ KeyPair
     :undoc-members:
     :member-order: groupwise
 
+.. warning::
+    **Not Audited:** The ``KeyPair.generate()`` function has not been audited for cryptographic security. Use at your own risk.
+
```
franciszekjob, line 37: I think it also makes sense to print such warning when directly using the function (similarly as we do for blind signing with Ledger). wdyt?

https://github.com/software-mansion/starknet.py/blob/5a92f63a98cc18a85918ff13f96dd39a6a50732e/starknet_py/net/signer/ledger_signer.py#L487-L494

ddoktorski: Tbh I think it would be too much

franciszekjob: I don't have a strong opinion on this, but why (considering we show it already in docs)?

ddoktorski: Usually warnings are for actionable items, so I'm worried this might just create noise for users

franciszekjob (approved):

Issue #1698, opened 2026-01-21
Fix Windows CI


franciszekjob:
In the scope of this task we should create a  linux Docker image, inside of it we should install starknet-devnet, run it and expose on port 5050. Also, update `checks.yml` part responsible for  starknet-devnet installation on Windows, so it uses the one from container.

PR #1718, opened 2026-04-30
Add explicit `package` in `load_contract` calls

ddoktorski:
Towards #1698

- Add explicit `package=` in `load_contract` calls where possible, especially for contracts that only exists in v2 directory to avoid relying on default behavior

### DIFF
--- a/starknet_py/tests/e2e/account/account_test.py
+++ b/starknet_py/tests/e2e/account/account_test.py
@@ -735,7 +735,7 @@ async def test_deploy_account_v3_auto_estimate_tip(
 
 @pytest.mark.asyncio
 async def test_declare_v3_with_tip(account):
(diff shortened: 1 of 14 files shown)


ddoktorski:

- #1718 ◀
    - #1719
        - #1720
            - #1721

franciszekjob (commented):
Are we always explicitly passing package in the function call? If so, do we want to keep `package` param optional at all?

ddoktorski:
> Are we always explicitly passing package in the function call? If so, do we want to keep package param optional at all?

Not always, e.g. https://github.com/software-mansion/starknet.py/blob/ddoktorski/fix-load-config/starknet_py/tests/e2e/contract_interaction/declare_test.py#L39

franciszekjob (approved):

PR #1719, opened 2026-04-30
Stop using `sys.args` logic in test

ddoktorski:
Towards #1698

- `sys.args` logic is unereliable when tests are run on multiple workers using `pytest-xdist`, each worker executes a separate command which doesn't include intiial `--contract_dir` value. To fix this we set global `_contract_dir` variable that is executed before running tests

### DIFF
--- a/starknet_py/tests/e2e/docs/quickstart/test_using_account.py
+++ b/starknet_py/tests/e2e/docs/quickstart/test_using_account.py
@@ -1,15 +1,15 @@
 import os
-import sys
 
 import pytest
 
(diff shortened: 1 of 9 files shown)


ddoktorski:

- #1718
    - #1719 ◀
        - #1720
            - #1721

starknet_py/tests/e2e/docs/quickstart/test_using_account.py
ddoktorski, line 16: Note: This is an example of a test that hasn't been checked on CI for a long time because of `sys.args` bug

franciszekjob (approved):

PR #1720, opened 2026-04-30
Use new contract in `test_declare_v3_tx`

ddoktorski:
Towards #1698

- This test fails when running on a single thread as the same contract is already declared in other test. We should have more sophisticated solution for example with salting instead of adding new contracts that are basically the same, but for now this is the easiest fix


ddoktorski:

- #1718
    - #1719
        - #1720 ◀
            - #1721
                - #1722

franciszekjob (approved):

PR #1721, opened 2026-04-30
Fix Windows CI tests

### LINKED ISSUE #1698: Fix Windows CI

ddoktorski:
Closes #1698

- Set up devnet on Windows jobs using WSL
- Bump devnet to stable `0.8.0` version
- Bump `poethepoet`

### DIFF
--- a/.github/actions/compile-contracts/action.yml
+++ b/.github/actions/compile-contracts/action.yml
@@ -27,9 +27,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: ${{ inputs.contracts_base_path }}/${{ inputs.package }}/target
-        key: ${{ runner.os }}-contracts-${{ inputs.package }}-${{ hashFiles(format('starknet_py/tests/e2e/mock/{0}/**', inputs.package)) }}
-        restore-keys: |
-          ${{ runner.os }}-contracts-${{ inputs.package }}-
+        key: contracts-${{ inputs.package }}-${{ hashFiles(format('starknet_py/tests/e2e/mock/{0}/**', inputs.package)) }}
 
     - name: Compile contracts ${{ inputs.package }}
       if: inputs.cache != 'true' || steps.cache-contracts.outputs.cache-hit != 'true'
--- a/.github/workflows/windows-tests.yml
+++ b/.github/workflows/windows-tests.yml
@@ -56,13 +56,13 @@ jobs:
         shell: wsl-bash {0}
         run: |
           nohup ./starknet-devnet --port 5051 --accounts 1 --seed 1 --state-archive-capacity full --initial-balance 1000000000000000000000000000000000000000000000000000000000000000000 --predeclare-argent --fork-network ${{ env.SEPOLIA_RPC_URL }} > devnet_fork.log 2>&1 &
-          for i in $(seq 60); do curl -sf http://127.0.0.1:5051/is_alive && break; sleep 1; done || (echo "Devnet (fork) failed to start"; exit 1)
+          for i in $(seq 60); do curl -sf http://127.0.0.1:5051/is_alive && exit 0; sleep 1; done; echo "Devnet (fork) failed to start"; exit 1
 
       - name: Start Devnet via WSL
         shell: wsl-bash {0}
         run: |
           nohup ./starknet-devnet --port 5050 --accounts 1 --seed 1 --state-archive-capacity full --initial-balance 1000000000000000000000000000000000000000000000000000000000000000000 --predeclare-argent > devnet.log 2>&1 &
-          for i in $(seq 60); do curl -sf http://127.0.0.1:5050/is_alive && break; sleep 1; done || (echo "Devnet failed to start"; exit 1)
+          for i in $(seq 60); do curl -sf http://127.0.0.1:5050/is_alive && exit 0; sleep 1; done; echo "Devnet failed to start"; exit 1
 
       - name: Verify devnet is running
         shell: pwsh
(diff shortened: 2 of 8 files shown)


ddoktorski:

- #1718
    - #1719
        - #1720
            - #1721 ◀
                - #1722

ddoktorski:
Successful run: https://github.com/software-mansion/starknet.py/actions/runs/25166099967

.github/actions/compile-contracts/action.yml
```diff
@@ -27,9 +27,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: ${{ inputs.contracts_base_path }}/${{ inputs.package }}/target
-        key: ${{ runner.os }}-contracts-${{ inputs.package }}-${{ hashFiles(format('starknet_py/tests/e2e/mock/{0}/**', inputs.package)) }}
-        restore-keys: |
```
ddoktorski, line 31: Note: This is not strictly related to main changes, but there is no benefit in restoring partial cache, so it was removed

.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,120 @@
...
+jobs:
+  run:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.10", "3.14" ]
+    env:
+      DEVNET_URL: http://127.0.0.1:5050
+      DEVNET_FORK_URL: http://127.0.0.1:5051
```
franciszekjob, line 25: nit: Does it make sense to pass these envs if we're using separate jobs here?

.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,120 @@
...
+          cache: 'pip'
+
+      # ====================== SETUP DEVNET ====================== #
+
+      - name: Setup WSL
+        uses: Vampire/setup-wsl@d1da7f2c0322a5ee4f24975344f67fc0f5baf364 # v7.0.0
+        with:
+          additional-packages: curl ca-certificates uuid-runtime jq git
+
+      - name: Download and Extract Devnet (PowerShell)
```
franciszekjob:
```suggestion
      - name: Download and Extract Devnet
```

.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,120 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.10", "3.14" ]
+    env:
+      DEVNET_URL: http://127.0.0.1:5050
+      DEVNET_FORK_URL: http://127.0.0.1:5051
+      SEPOLIA_RPC_URL: ${{ secrets.SEPOLIA_RPC_URL }}
+    steps:
+      - uses: actions/checkout@v4
```
franciszekjob: nit: I know you've copied it so that's fine but we can use v6:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,120 @@
...
+        run: |
+          curl.exe -L -o devnet.tar.gz https://github.com/0xSpaceShard/starknet-devnet-rs/releases/download/v${{ env.DEVNET_VERSION }}/starknet-devnet-x86_64-unknown-linux-musl.tar.gz
+          tar.exe -xzf devnet.tar.gz
+
+      - name: Start Devnet with fork mode via WSL
+        if: ${{ inputs.fork_mode }}
+        shell: wsl-bash {0}
+        run: |
+          nohup ./starknet-devnet --port 5051 --accounts 1 --seed 1 --state-archive-capacity full --initial-balance 1000000000000000000000000000000000000000000000000000000000000000000 --predeclare-argent --fork-network ${{ env.SEPOLIA_RPC_URL }} > devnet_fork.log 2>&1 &
+          for i in $(seq 60); do curl -sf http://127.0.0.1:5051/is_alive && break; sleep 1; done || (echo "Devnet (fork) failed to start"; exit 1)
```
franciszekjob: Since `for` loop ends with `sleep 1` in each iteration (which always returns an exit code of `0`), the whole line will always be considered successful. This means the condition after `||` will never trigger, and the pipeline will continue even if devnet fails to start.

.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,120 @@
...
+    inputs:
+      fork_mode:
+        type: boolean
+        default: false
+      test_task:
+        type: string
+        required: true
+      circular_imports_check:
+        type: boolean
+        default: false
```
franciszekjob: Is this input actually necessary (I see there's `if` which uses it, but why do we want to conditionally run circular imports check)?

.github/workflows/windows-tests.yml
ddoktorski: Should be fixed now

.github/workflows/windows-tests.yml
ddoktorski: Removed, now this check is run in both calls (standard checks and docs)

.github/workflows/windows-tests.yml
ddoktorski: These envs are used by fixtures in `devnet.py` so they're necessary

franciszekjob (approved):

.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,116 @@
+name: Windows Tests
+
+env:
+  DEVNET_VERSION: "0.8.0"
```
franciszekjob, line 4: nit: (same below)
```suggestion
  DEVNET_VERSION: "0.8.1"
```

franciszekjob (approved):

.github/workflows/windows-tests.yml
ddoktorski: https://github.com/software-mansion/starknet.py/pull/1722

PR #1723, opened 2026-05-18
Fix extended network tests in `checks.yml`

ddoktorski:
- `starknet_py/tests/e2e/tests_on_networks/account_test.py::test_declare_v3` fails on push, as it requires salted contract to be generated. Otherwise it fails with `contract already declared`


ddoktorski:
Note: Failing tests not related to this change. Fail on development branch: https://github.com/software-mansion/starknet.py/actions/runs/26031046426

